PR #14, opened 2015-10-23
IPv6 compatibility, multiple address compatibility

philpennock:
The current code and tests fail in the presence of IPv6 and fail if there are multiple local IP addresses.

For main logic, still use the external service to find local IP, but also check local IP addresses for any regular IPs.  For tests, ask for only our IPv4 addresses, which is enough to not have the `xip.io` usage.

### DIFF
--- a/acme/simple_http_challenge.go
+++ b/acme/simple_http_challenge.go
@@ -24,6 +24,8 @@ type simpleHTTPChallenge struct {
 func (s *simpleHTTPChallenge) CanSolve(domain string) bool {
 
 	// determine public ip
+	// CAVEAT: assumes one public IP, breaks badly if have both IPv4 and IPv6,
+	// let alone multiple IPv6 addresses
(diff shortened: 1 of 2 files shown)


xenolf:
Hello and thanks for the PR! :)

I was actually thinking about removing the pre-solve checks from challenges. The issue I see with those is that we can never reliably determine if a certain server responds to a domain name.

What's your take on this?

philpennock:
I haven't read the acme-spec to understand the security issues in play, sorry.  I was just unbreaking for my use-case (before things failed later because I'm not in the closed beta, but hey, I got to play a little).

mholt:
@philpennock I think it's less of a security issue and more of a question of preference. We're thinking that `CanSolve()` is redundant and not always correct due to system configurations, so we might just rip those methods out entirely. Would you be OK with that?

xenolf:
I removed challenge pre-checks in the latest revision.

PR #16, opened 2015-10-26
Add a webroot mode which copy the token over ssh to a web server

yaslama:


mholt:
Before this is merged in, consider that this package may eventually make it into the Go standard library -- and if it does, all contributors to lego will need to agree to the Go CLA. @yaslama would that be OK with you?

mholt:
@yaslama Thanks for the PR. Does this implement something in the spec (I don't recognize this off-hand)? Do you have a link to some documentation about what this is?

yaslama:
Hi.
First, no problem for me to agree to the Go CLA. Concerning the spec, something like that is implemented in https://github.com/letsencrypt/letsencrypt/blob/3885a40c421636805a77a0f1c573d1727169eddb/letsencrypt/plugins/webroot.py
See also https://github.com/letsencrypt/letsencrypt/search?q=webroot&type=Code&utf8=%E2%9C%93
I just added a way to send the token file over ssh instead of storing it in the local fs.
It's not very easy to add tests because we need to be sure to have access to a web server whose root directory can be accessed by ssh/sftp.
If we can add a concept of authenticator plugin to each type of challenge, we could create 2 authenticator for the simpleHTTPChallenge, one running a local http server and another copying the token to the webroot of a running http server.
BTW, @mholt , thanks for your work on caddy.

mholt:
I like the 'plugin' idea -- maybe just move this change outside the core Simple HTTP solver and change it so that the authenticator can be swapped out. Kind of like how [bufio.Scanner](https://golang.org/pkg/bufio/#Scanner) defaults to `bufio.ScanLines` - but it can be customized. @xenolf what do you think?

xenolf:
Hey!
Yeah I also like the plugin idea.
I will think about an interface and implement that. @yaslama would refactor your PR on top of the changes I'm going to make? I don't know exactly when I will get around doing that, but I will let you know.

yaslama:
No problem for me. As soon as the plugin interface is committed, I'll refactor my PR. Thanks a lot!

PR #21, opened 2015-11-06
Make acme.Logger optional; otherwise use standard log.Logger

mholt:
My logic behind this change is basically this: a user of this library can set `acme.Logger` to something custom if they want to; otherwise it uses the default log.Logger instance. (Currently, acme package _never_ uses the standard log.Logger, which is causing problems in Caddy which uses that logger.)

Also, my next PR will clean up the log statements a bit, making them more consistent and also removing the superfluous `\n` at the end (the log package adds them for us, which is nice -- keeps our strings cleaner).

### DIFF
--- a/acme/client.go
+++ b/acme/client.go
@@ -10,22 +10,23 @@ import (
 	"io/ioutil"
 	"log"
 	"net/http"
-	"os"
 	"regexp"
 	"strconv"
 	"strings"
 	"time"
 )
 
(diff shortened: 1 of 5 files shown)


mholt:
Uhmmm not sure why that Windows build failed. Maybe just a fluke?

mholt:
Most recent commit just rewords a few of the log messages and removes a couple unused functions. Also a couple nil pointer checks.

PR #26, opened 2015-11-13
Add JSON tags to RegistrationResource as used in official client

zakjan:
It enables (un)marshaling regr.json files used in official client.


xenolf:
LGTM :)

Issue #27, opened 2015-11-16
Webroot authenticator


zakjan:
Do you plan to implement webroot authenticator? ACME challenge file can be written to a directory, while another webserver (eg. nginx) is configured to read from it.

Not only it allows to run the client LE application as non-root user, it is also possible to run multiple cert requests at once, because it doesn't block any incoming port.

xenolf:
Well yeah. 
We had a pull request to add this (#16) but unfortunately it was closed by the author.

For now my focus is at implementing the remaining challenges to get full ACME coverage and SAN certificates.

zakjan:
OK. Maybe I'll try to implement it in my spare time.

I noticed you merged http-01, which is great timing because simple-http will be shut down this week. https://community.letsencrypt.org/t/turning-off-old-insecure-challenge-types/3825

xenolf:
That would be great :)

> I noticed you merged http-01, which is great timing because simple-http will be shut down this week.

Yes this was intentional as [caddy](https://github.com/mholt/caddy) depends on this library.

zakjan:
Closing in favor of #32

PR #28, opened 2015-11-16
Add support for webroot over ssh for http-01 challenge

yaslama:
This PR replace #16 (see also #27)

### DIFF
--- a/acme/client.go
+++ b/acme/client.go
@@ -57,7 +57,7 @@ type Client struct {
 // usr - A filled in user struct
 // keyBits - Size of the key in bits
 // optPort - The alternative port to listen on for challenges.
-func NewClient(caURL string, usr User, keyBits int, optPort string) (*Client, error) {
+func NewClient(caURL string, usr User, keyBits int, optPort string, webRoot string) (*Client, error) {
 	privKey := usr.GetPrivateKey()
 	if privKey == nil {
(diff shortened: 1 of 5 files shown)


mholt:
As far as I understand it, webroot basically places a file on the file system for the web server to serve up during the challenge, then the file can (should?) be deleted. I may be misunderstanding how it works, but I imagine we should make sure the token file gets cleaned up afterward, right?

Also, I'm not sure that building it directly into the client workflow is quite the right thing to do here. It seems like webroot is just one specific way of doing http-01 (not other challenges); instead of having a "web app" serve the response for the ACME server, you mimic that behavior for a static file server by plopping down a file at the right path.

Perhaps the actual _way_ the challenge is solved should be swappable... or something.

I think the "webroot" way of solving this challenge is good, but maybe there's a better way to design it for this library.

zakjan:
I'd rather first implement "local webroot" instead of even more specific "webroot over ssh"

yaslama:
Ok. I'll try to implement a way to "swap" the way to solve the challenge first.

yaslama:
See #32 where I "extracted" the webserver method from http_challenge. If it's ok, I'll add the webroot methods for local filesystem and remote filesystem.

xenolf:
This should probably get moved to an implementation using providers if you are up for it :)

PR #29, opened 2015-11-17
fix missing failures return value

tpng:


xenolf:
Thank you for your PR :) 
Unfortunately I already included this fix in #30 - Closing this in favor of #30.

PR #32, opened 2015-11-18
In HTTP-01, modularize the way of presenting the token.

yaslama:
First, with a simple local webserver.


zakjan:
LGTM :)

acme/http_challenge.go
```diff
@@ -4,56 +4,43 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"net"
 	"net/http"
-	"strings"
 	"time"
 )
 
+type httpChallengeMethod interface {
```
mholt: Shouldn't this be exported so users can write their own (if they want to)?

xenolf: I think we do want the user to be at least able to write their own.

xenolf:
This is looking pretty good. But could you fix the tests? :)

acme/http_challenge.go
yaslama: The commit https://github.com/xenolf/lego/commit/61a0912a11df96a7956b60a5a1b66c63f29f402d fixes that

acme/http_challenge_webserver.go
```diff
@@ -0,0 +1,85 @@
...
+	"strings"
+)
+
+type HttpChallengeWebserver struct {
+	optPort string
+	start   chan net.Listener
+	end     chan error
+}
+
+func NewHttpChallengeWebserver(optPort string) HttpChallengeMethod {
```
mholt: For some reason, the word "Webserver" is really bothering me - maybe because it's actually two words. Can we change this to `NewHTTPChallengeWebServer` instead with a capital S and a capital HTTP?

yaslama: In this case, httpChallenge can be renamed to HTTPChallenge if we change it to be exported (like I proposed in my remark to https://github.com/xenolf/lego/issues/36) ?

xenolf:
I'd like to discuss the following change to the `HttpChallengeMethod` interface 

```go
type HttpChallengeMethod interface {
	PresentToken(domain, token, keyAuth string) error
	CleanUp() error
}
```

This would get rid of the callback to `PresentToken` and would establish a clear seperation between "challenges" and "solvers".
See [this gist](https://gist.github.com/xenolf/53ea12aec783176515d8) for a quick implementation.

mholt:
^ I support this suggestion. This PR is definitely on the right track, and I think @xenolf's suggestion is a fine way to improve it.

xenolf:
Any updates on this? :)

janeczku:
This is a great suggestion. Some notes on the interface:

`type HttpChallengeSolver interface`

We shouldn't limit the scope of this interface to HTTP challenges. We also need this stuff for the other types of challenges (TLS-SNI, DNS, ...). So this should be global interface that we can use for any type of challenge. Since all challenge types work by provisioning some kind of resource based on domain, token and keyAuth, they can share a single interface.

About the interface naming: `HttpChallengeMethod`or `HttpChallengeSolver` imply that there are several ways to finalize the http-01 challenge. But there is only one method:
> Provision the validation resource to an HTTP server

Of course there maybe different **Providers** or **Provisioners** that implement that method, hence `ChallengeProvider` or simply `Provider` or `Provisioner` would be a better name.

**Proposal**

*client.go*

```GO
type Provisioner interface {
	ProvisionResource(domain, token, keyAuth string) error
	CleanupResource(domain, token string) error
}
```

*http_challenge.go*

```GO
type httpChallenge struct {
	jws         *jws
	provisioner Provisioner
}
```

*tls_sni_challenge.go*

```GO
type tlsSNIChallenge struct {
	jws         *jws
	provisioner Provisioner
}
```

you get the idea...

jehiah:
@xenolf @yaslama i think you can close this now based on #76

PR #35, opened 2015-11-20
Don't try appending /directory to caURL

mholt:
The spec doesn't require /directory, so I think it's best left for the user to specify it. Better to change it now than later. Also improved some comments/docs and fixed a test.


xenolf:
LGTM

Issue #39, opened 2015-12-04
Listen IP/interface of lego unclear


sharpner:
Hi there!

I upgraded to v0.1 and now I am getting this error: 
```
2015/12/04 10:10:14 Could not create client:decode directory: invalid character '<' looking for beginning of value
```

I am using the following parameters:
```
lego --domains=mydomain.de -B 4096 --email="email@mydomain.de" -s "https://acme-v01.api.letsencrypt.org/" --path=/var/container/nginx/newcert/ --port 9999 run
```

Did something change? I tried using the existing keys and creating new ones, the error is still the same.

Thanks!

sharpner:
found it, you now need to use ```https://acme-v01.api.letsencrypt.org/directory```

sharpner:
So somethings fishy with version v0.1.0. 
I couldn't get it to work no matter what: 

```
/root/lego --domains=mydomain.de -B 4096 --email="n.wagensonner@manyminds.de"  --path=/var/container/nginx/cert/ --port 9999 run
2015/12/04 10:31:49 [INFO] acme: Obtaining bundled SAN certificate for mydomain.de
2015/12/04 10:31:50 [INFO] acme: Trying to solve TLS-SNI-01
2015/12/04 10:31:52 [codinator.de] Could not obtain certificates
	The server could not validate our request.
```

So I downgraded lego to #6a803c626577b215325830567f8131b985851cf6 and ran:
```
/root/lego.old --domains=mydomain.de -B 4096 --email="n.wagensonner@manyminds.de" -s "https://acme-v01.api.letsencrypt.org/directory" --path=/var/container/nginx/cert/ --port 9999 run
2015/12/04 10:32:11 [INFO] acme: Obtaining bundled SAN certificate for mydomain.de
2015/12/04 10:32:11 [INFO] acme: Trying to solve HTTP-01
2015/12/04 10:32:11 [INFO] Served key authentication
2015/12/04 10:32:13 The server validated our request
2015/12/04 10:32:13 [INFO] acme: Validations succeeded; requesting certificates
2015/12/04 10:32:15 [INFO] acme: Requesting issuer cert from https://acme-v01.api.letsencrypt.org/acme/issuer-cert
2015/12/04 10:32:15 [codinator.de] Server responded with a certificate.
```

with success. Is it my mistake?

sharpner:
okay I found the issue, lego now only listens to the public interface/ip of the given domain(/s).

xenolf:
[...]

Issue #36, opened 2015-11-20
Suggestion about NewClient


mholt:
To make NewClient easier to use and more customizable, I propose the following usage:

```go
challengeTypes := acme.ChallengeHTTP | acme.ChallengeTLSSNI | acme.ChallengeDNS
client, err := acme.NewClient(dirURL, user, keyBits, challengeTypes)
// err check

// the client could have exported fields to allow it to be customized, for example:
client.HTTPChallengeServer = myServer // or whatever
client.HTTPChallengePort = optPort // instead of passing it into NewClient, which feels weird for just one challenge type
// any other customizations here, like challenge priorities:
client.ChallengePriority = []acme.ChallengeType{acme.ChallengeHTTP, acme.ChallengeTLSSNI}
// ^ although I'm not sure if the bitmask above should be mutually exclusive with priorities...
```

Basically, change the last argument of NewClient to be a bitmask (maybe? depends on how challenge priorities is implemented) and allow the *Client to be customized after its construction.

yaslama:
Why not something like ?
```Go
challenges := []acme.Challenges{acme.NewHTTPChallenge(acme.NewHTTPChallengeWebServer(optPort)), acme.NewTlsSNIChallenge()}
client, err := acme.NewClient(dirURL, user, keyBits, challenges)
```

mholt:
The only thing about calling `New...Challenge` once for a client is that a client may have multiple challenges to perform, thus needing to create new ones for each challenge -- I think (@xenolf would know better).

xenolf:
I think that the challenges should definitely be internal to the library and not passed in through NewClient. The "solvers" (or however we want to call them) may be passed in as the user should have the option to add custom ones to the flow.
If a user wants to customize a built-in "solver" he may construct the solver himself, otherwise the client will add them with default values to the internal slice. 

All of the above is just a braindump and I am looking for feedback :)

mholt:
> I think that the challenges should definitely be internal to the library and not passed in through NewClient.

If you mean that challenges shouldn't be constructed & passed into NewClient, I agree. I still like the idea of the user being able to control which challenges are allowed, though. (Then again, maybe this can be one of the things that is set later, like the priority.)

> The "solvers" (or however we want to call them) may be passed in as the user should have the option to add custom ones to the flow. If a user wants to customize a built-in "solver" he may construct the solver himself, otherwise the client will add them with default values to the internal slice.

:+1:

dpiddy:
[...]

PR #40, opened 2015-12-04
Allow users to specify which IP address to listen on

vanbroup:


sharpner:
nice timing would fix #39

xenolf:
Hello and thanks for your pull request. :)
I am open to the idea of letting the user explicitly specify a listening IP. 
Could you think of a way of doing this without changing NewClient? We were discussing changes to NewClient in #36 and wanted to move the optPort out of NewClient as well.

xenolf:
This feature was implemented in a different way and landed in master.

PR #41, opened 2015-12-05
Simplify the challenge code by splitting off a validation function

tommie:
A suggested simplification/refactoring of the challenge code.

This is part of a greater goal of implementing a file-backed HTTP-01 challenge.

The individual commits are more pleasant than the full pull-request, but they are closely related.

### DIFF
--- a/acme/client.go
+++ b/acme/client.go
@@ -16,8 +16,13 @@ import (
 	"time"
 )
 
-// Logger is an optional custom logger.
-var Logger *log.Logger
+var (
+	// DefaultSolvers is the set of solvers to use if none is given to NewClient.
+	DefaultSolvers = []string{"http-01", "tls-sni-01"}
+
+	// Logger is an optional custom logger.
+	Logger *log.Logger
+)
 
 // logf writes a log entry. It uses Logger if not
 // nil, otherwise it uses the default log.Logger.
@@ -56,9 +61,12 @@ type Client struct {
 // the ACME directory located at caDirURL for the rest of its actions. It will
 // generate private keys for certificates of size keyBits. And, if the challenge
 // type requires it, the client will open a port at optPort to solve the challenge.
-// If optPort is blank, the port required by the spec will be used, but you must
-// forward the required port to optPort for the challenge to succeed.
-func NewClient(caDirURL string, user User, keyBits int, optPort string) (*Client, error) {
+//
+// If optSolvers is nil, the value of DefaultSolvers is used. If given explicitly,
+// it is a set of solver names to enable. The "http-01" and "tls-sni-01" solvers
+// take an optional TCP port to listen on after a colon, e.g. "http-01:80". If
+// the port is not specified, the port required by the spec will be used.
+func NewClient(caDirURL string, user User, keyBits int, optSolvers []string) (*Client, error) {
 	privKey := user.GetPrivateKey()
 	if privKey == nil {
 		return nil, errors.New("private key was nil")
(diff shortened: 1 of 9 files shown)


mholt:
I agree with stripping out the async code; since the challenges require polling anyway, there's at least a couple seconds of blocking there which I don't think the goroutines make faster. Also, :100: for deleting code :smile:

acme/client.go
```diff
@@ -16,8 +16,13 @@ import (
 	"time"
 )
 
-// Logger is an optional custom logger.
-var Logger *log.Logger
+var (
+	// DefaultSolvers is the set of solvers to use if none is given to NewClient.
+	DefaultSolvers = []string{"http-01", "tls-sni-01"}
```
mholt: I like where this is going. Maybe these could be consts instead?

acme/client.go
```diff
@@ -56,9 +61,12 @@ type Client struct {
 // the ACME directory located at caDirURL for the rest of its actions. It will
 // generate private keys for certificates of size keyBits. And, if the challenge
 // type requires it, the client will open a port at optPort to solve the challenge.
-// If optPort is blank, the port required by the spec will be used, but you must
-// forward the required port to optPort for the challenge to succeed.
-func NewClient(caDirURL string, user User, keyBits int, optPort string) (*Client, error) {
+//
+// If optSolvers is nil, the value of DefaultSolvers is used. If given explicitly,
+// it is a set of solver names to enable. The "http-01" and "tls-sni-01" solvers
+// take an optional TCP port to listen on after a colon, e.g. "http-01:80". If
```
mholt: The string with the port is clever, but I'm not sure it makes much sense, since the first part isn't a host. I think it feels more idiomatic to specify the optional port separately in some sort of struct maybe.

acme/client.go
tommie: Unfortunately slice literals are not const expressions, so not possible.

mholt: My bad. Not the slice, I meant the string values inside it.

acme/client.go
```diff
@@ -165,74 +172,16 @@ func (c *Client) Register() (*RegistrationResource, error) {
...
-}
-
-// ObtainSANCertificate tries to obtain a single certificate using all domains passed into it.
+// ObtainCertificate tries to obtain a single certificate using all domains passed into it.
 // The first domain in domains is used for the CommonName field of the certificate, all other
 // domains are added using the Subject Alternate Names extension.
 // If bundle is true, the []byte contains both the issuer certificate and
 // your issued certificate as a bundle.
-func (c *Client) ObtainSANCertificate(domains []string, bundle bool) (CertificateResource, map[string]error) {
+func (c *Client) ObtainCertificate(domains []string, bundle bool) (CertificateResource, map[string]error) {
```
mholt: This is good. I guess a regular certificate can be obtained by passing in just one hostname anyway.

acme/client.go
tommie: Agreed. The problem I'm trying to get around is that not all solvers will have the concept of a port. E.g. the file challenge for http-01 I've made has a root path, but no port. (It's a directory where Apache serves the file from.)

So the options need to be per-solver, or per category of solvers. This was the simplest solution. Another would be to make solver an exported type and allow feeding that into NewClient. But that would also require another API change (not having to supply the jws object on construction). I felt that was overkill for my purposes: allowing multiple implementations of the http-01 solver.

acme/client.go
tommie: Ah, yes.

acme/client.go
tommie: Yepp, the code paths merged in requestCertificate anyway.

acme/client.go
xenolf: The issue I see with it is that a user has to specify all the challenges here just to set an alternate port.

acme/http_challenge.go
```diff
@@ -105,17 +34,17 @@ func (s *httpChallenge) startHTTPServer(domain string, token string, keyAuth str
...
-	s.start <- listener
+	defer listener.Close()
 
-	path := "/.well-known/acme-challenge/" + token
+	path := "/.well-known/acme-challenge/" + chlng.Token
 
 	// The handler validates the HOST header and request type.
 	// For validation it then writes the token the server returned with the challenge
-	http.HandleFunc(path, func(w http.ResponseWriter, r *http.Request) {
+	mux := http.NewServeMux()
```
xenolf: Does a mux really solve the same problem here? I used channels not for more "speed" but to be able to explicitly close the listener and wait for it to close. Otherwise there would be racing conditions when solving more then one challenge.

mholt:
Ah, I forgot about the real reason the channels were used: to solve more than one challenge. Is there a simpler way to do that so that multiple challenges still work?

acme/http_challenge.go
tommie: Closing the listener (on line 40 above) will release the listener port and cause http.Serve to return (in the goroutine line 58). listener.Close() will not return until the port is freed, so there should be no race on I/O resources. The http.Serve call will eventually return, but since it has no global state now that it has a private ServeMux, this shouldn't cause an issue.

tommie:
There is no difference in challenges concurrency. Note that the solver calls were completely blocking (only returning when the ACME server had validated) anyway.

xenolf:
This was merged in 595f684e27a15f3d63b1c90122a887f9ff24305b with some changes added on top.
Thank you!

Issue #44, opened 2015-12-07
Exit code is always 0


alcaeus:
When certificate generation fails, the client still exits with code 0:
```
user@host:~$ ./lego --domains=<redacted> -B 4096 -m <redacted> --path=/etc/ssl/letsencrypt --port=9999 run
2015/12/07 15:02:40 [INFO] acme: Obtaining bundled SAN certificate for <redacted>
2015/12/07 15:02:40 [INFO] acme: Trying to solve HTTP-01
2015/12/07 15:02:41 [<redacted>] Could not obtain certificates
	The server could not validate our request.
user@host:~$ echo $?
0
```

This makes checking for errors in the certificate generation process unnecessarily difficult.

xenolf:
You are right, this is not how it should be. Thank you for bringing this to our attention.

xenolf:
Resolved in 1b07617

PR #45, opened 2015-12-07
Make sure the CLI returns proper non-zero exit codes on errors.

xenolf:
Fix for #44


mholt:
+1 to failing fast in these cases rather than just printing a message.

PR #66, opened 2015-12-30
Implement custom User-Agent string

mholt:
Also a couple miscellaneous vet fixes.

Tried to be complaint with http://tools.ietf.org/html/rfc7231#section-5.5.3


mholt:
YAY!!! :tada: Thanks :smile:

PR #68, opened 2016-01-01
Base64 encode dns-01 record

janeczku:
https://github.com/letsencrypt/boulder/pull/1295 changed encoding of DNS record to base64url without padding.

### DIFF
--- a/acme/dns_challenge.go
+++ b/acme/dns_challenge.go
@@ -2,11 +2,12 @@ package acme
 
 import (
 	"crypto/sha256"
-	"encoding/hex"
+	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"net/http"
+	"strings"
 	"time"
 )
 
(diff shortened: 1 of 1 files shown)


xenolf:
Thank your for this :)

acme/dns_challenge.go
```diff
@@ -34,9 +35,9 @@ func (s *dnsChallenge) Solve(chlng challenge, domain string) error {
 	}
 
 	keyAuthShaBytes := sha256.Sum256([]byte(keyAuth))
-	// FIXME: Currently boulder does not conform to the spec as in it uses hex encoding instead
-	// of the base64 encoding mentioned by the spec. Fix this if either the spec or boulder changes!
-	keyAuthSha := hex.EncodeToString(keyAuthShaBytes[:sha256.Size])
+	// base64URL encoding without padding
+	keyAuthSha := base64.URLEncoding.EncodeToString(keyAuthShaBytes[:sha256.Size])
+	keyAuthSha = strings.TrimRight(keyAuthSha, "=")
```
shurcooL: What version of Go does `lego` use/require?

If it uses 1.5+ only, then the above can be simplified by using the new `base64.RawURLEncoding`:

https://godoc.org/encoding/base64#RawURLEncoding

Using it will not emit padding "=" on the right, so `keyAuthSha = strings.TrimRight(keyAuthSha, "=")` would become unnecessary.

xenolf: I'd like to stay compatible with go 1.4 for the time being.

shurcooL: Got it. Then my comment does not apply.

Issue #64, opened 2015-12-28
Failover by trying other challenges


mholt:
It would be good - perhaps optionally? - if failure to verify a domain name could try again using a different challenge (whichever challenges are not excluded by the user) to maximize the chances of getting a certificate.

I know we've talked about this before and is on the roadmap already, but just documenting this here so we don't forget.

PR #69, opened 2016-01-05
Try all available solvers instead of a random one

abh:
This exposes some other bug where the client doesn't accept that it
got a successful response, so it doesn't actually work but hopefully
it'll expose how it's currently broken. :-) I couldn't get further tonight
before having to sleep.

This might be related to #64; though I don't understand how it's
intended to work now as it just appears to try a random challenge.

### DIFF
--- a/acme/client.go
+++ b/acme/client.go
@@ -408,9 +408,11 @@ func (c *Client) solveChallenges(challenges []authorizationResource) map[string]
 			for i, solver := range solvers {
 				// TODO: do not immediately fail if one domain fails to validate.
 				err := solver.Solve(authz.Body.Challenges[i], authz.Domain)
-				if err != nil {
-					failures[authz.Domain] = err
+				if err == nil {
+					// we're done
+					return nil
 				}
+				failures[authz.Domain] = err
 			}
 		} else {
 			failures[authz.Domain] = fmt.Errorf("[%s] acme: Could not determine solvers", authz.Domain)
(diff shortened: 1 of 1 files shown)


acme/client.go
```diff
@@ -432,13 +434,9 @@ func (c *Client) chooseSolvers(auth authorization, domain string) map[int]solver
...
 		}
-
-		// If we can solve the whole combination, return the solvers
-		if len(solvers) == len(combination) {
-			return solvers
-		}
 	}
-	return nil
+
+	return solvers
```
mholt: I'm not sure I understand the changes here. What exactly is the current behavior and what is the new behavior?

acme/client.go
```diff
@@ -408,9 +408,11 @@ func (c *Client) solveChallenges(challenges []authorizationResource) map[string]
 			for i, solver := range solvers {
 				// TODO: do not immediately fail if one domain fails to validate.
 				err := solver.Solve(authz.Body.Challenges[i], authz.Domain)
-				if err != nil {
-					failures[authz.Domain] = err
+				if err == nil {
+					// we're done
+					return nil
```
mholt: What about for the other challenges (for other domains, see for loop on L405)?

abh: Hmm, that's possible. I think/thought that the loop for going through each domain was higher up in the code, but maybe not. I was only testing it with one domain.

With the current behavior any failure will make it fail, as far as I could tell.

acme/client.go
abh: The current behavior is to try one solver (more or less at random, depending on the map order for each execution). The change doesn't actually work, but the intention was to make it try both solvers and succeed as soon as one works.

acme/client.go
xenolf: This function finds all appropriate solvers for all given domains and executes them. Each entry in the challenges array stands for one domain. Returning early after the first successful solve will definitely break multiple domains.

acme/client.go
xenolf: The check that was deleted on line 437 is essential tho. If this check fails we can't solve the whole combination and should not try to do so. A combination can contain multiple challenges for one domain.

It is true that we should probably try more then one solver if it fails.

xenolf:
Closing for inactivity.

PR #70, opened 2016-01-05
Listen on *:port when the port is specified

abh:
The expectation should be that if you're listening to a non-default port
you are forwarding the connection in some way and listening on the
IP from DNS won't be useful.

Also print the {ip}:port that's being listened to.

This should help on #39; the issue I ran into was that lego could
listen on public-ip:port (when I specified the port), but I was
forwarding the connection to private-ip:port.

### DIFF
--- a/acme/http_challenge.go
+++ b/acme/http_challenge.go
@@ -95,12 +95,13 @@ Loop:
 func (s *httpChallenge) startHTTPServer(domain string, token string, keyAuth string) {
 
 	// Allow for CLI port override
-	port := ":80"
+	port := domain + ":80"
 	if s.optPort != "" {
 		port = ":" + s.optPort
 	}
 
-	listener, err := net.Listen("tcp", domain+port)
+	logf("[INFO] Listening to %s", port)
+	listener, err := net.Listen("tcp", port)
 	if err != nil {
 		// if the domain:port bind failed, fall back to :port bind and try that instead.
 		listener, err = net.Listen("tcp", port)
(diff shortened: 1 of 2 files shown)


acme/http_challenge.go
```diff
@@ -95,12 +95,13 @@ Loop:
 func (s *httpChallenge) startHTTPServer(domain string, token string, keyAuth string) {
 
 	// Allow for CLI port override
-	port := ":80"
+	port := domain + ":80"
```
mholt: This should probably use `net.JoinHostPort`

acme/http_challenge.go
```diff
@@ -95,12 +95,13 @@ Loop:
...
 	if s.optPort != "" {
 		port = ":" + s.optPort
 	}
```
mholt: This PR seems to ignore or negate the fallback (see below, L107)... I don't think this will work.

acme/http_challenge.go
abh: It wasn't using that before (on L103). I can rename it to 'listen', that makes sense.

acme/http_challenge.go
abh: The change is to make it so if you specify a port number it defaults to the fallback behavior; it doesn't ignore it just make it happen up front. It could maybe be optimized to not try the fallback behavior (too), but I'm not sure there's any harm in trying and failing the same listen twice.

The old behavior (listen on the DNS resolved IP) doesn't make sense if you are specifying a port since you need something else to forward the port and who knows what you are trying to do.

xenolf:
The code in question has been changed in the latest release. (commit de29381f7a0bf0288e6164715dff695ed014960a)

The default is to listen on all interfaces while allowing for `:port` or `interface:port` overrides.
Keep in mind that the listen address has no longer anything to do with the DNS name being verified.

PR #73, opened 2016-01-11
Merging master into add-dns-challenge

captncraig:
Also adding client.UseDNSChallenge in order to allow library users to access the dns functionality.


acme/client.go
```diff
@@ -611,3 +577,54 @@ func parseLinks(links []string) map[string]string {
...
+		hdr, err = getJSON(uri, &challengeResponse)
+		if err != nil {
+			return err
+		}
+	}
+}
+
+// UseDNSChallenge configures the client to use the provided DNSProvider
+// as it's only challenge
+func (c *Client) UseDNSChallenge(d DNSProvider) {
```
captncraig: this is a change I made outside of just a merge.

acme/dns_challenge.go
```diff
@@ -44,6 +44,10 @@ func (s *dnsChallenge) Solve(chlng challenge, domain string) error {
 		return err
 	}
 
+	// sleep a little bit to let the record propigate.
```
captncraig: This is the only other change I made besides merging. Temporary, but it makes things work until querying is implemented.

acme/client.go
mholt: I'd rather just stick to the ExcludeChallenges method, listing every challenge except DNS. (There's only 4 challenges total.)

captncraig: Exclude doesn't give the capability I need. I have to give it the dns provider which has a ton of internal configuration.

xenolf:
DNS is now in master :)

PR #75, opened 2016-01-13
Add explicit fallthrough

ernestoalejo:
An empty case in Go does not fallthrough to the next one.

### DIFF
--- a/acme/client.go
+++ b/acme/client.go
@@ -468,10 +468,8 @@ func (c *Client) requestCertificate(authz []authorizationResource, bundle bool, 
 		PrivateKey: privateKeyPem}
 
 	for {
-
 		switch resp.StatusCode {
-		case 202:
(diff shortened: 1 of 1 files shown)


shurcooL:
If the intent is to have 202 and 201 handled identically (currently, nothing happens for 202 since there's no `fallthrough`), then an alternative way, instead of adding `fallthrough`, is to write:

```Go
switch resp.StatusCode {
case 201, 202:
    ...
```

I think that might be nicer/cleaner.

ernestoalejo:
+1, I thought to use `201, 202` and then I completely forgot when writing the patch. I'll update it.

xenolf:
Thanks for catching this :)
I will merge it after you updated the PR.

ernestoalejo:
@xenolf Changed

PR #76, opened 2016-01-15
http-01: refactor provider for presenting token

jehiah:
This implements the last suggestion from #32 by @janeczku to separate solving the `http-01` challenge, from making the `keyAuth` available at the appropriate path.

This change creates a new interface

```go
type HTTPChallengeProvider interface {
	PresentToken(domain, path, keyAuth string) error
	CleanUp(domain, path string)
}
```

And a `httpChallengeServer` (extracted from `httpChallenge`) which implements that interface and is used by `Client`. 

I am looking to use `lego` as a library to generate certs and will need to have my own logic for making the token available, so the suggestion in #32 was quite useful.

@xenolf If you have any thoughts on this change (or suggestions on impelemtnation) i'm all ears.

### DIFF
--- a/acme/challenges.go
+++ b/acme/challenges.go
@@ -1,0 +1,15 @@
+package acme
+
+type Challenge string
+
+const (
+	// HTTP01 is the "http-01" ACME challenge https://github.com/ietf-wg-acme/acme/blob/master/draft-ietf-acme-acme.md#http
+	// Note: HTTP01ChallengePath returns the URL path to fulfill this challenge
+	HTTP01 = Challenge("http-01")
+	// TLSSNI01 is the "tls-sni-01" ACME challenge https://github.com/ietf-wg-acme/acme/blob/master/draft-ietf-acme-acme.md#tls-with-server-name-indication-tls-sni
+	// Note: TLSSNI01ChallengeCert returns a certificate to fulfill this challenge
+	TLSSNI01 = Challenge("tls-sni-01")
+	// DNS01 is the "dns-01" ACME challenge https://github.com/ietf-wg-acme/acme/blob/master/draft-ietf-acme-acme.md#dns
+	// Note: DNS01Record returns a DNS record which will fulfill this challenge
+	DNS01 = Challenge("dns-01")
+)
--- a/acme/client.go
+++ b/acme/client.go
@@ -54,7 +54,7 @@ type Client struct {
 	jws        *jws
 	keyBits    int
 	issuerCert []byte
-	solvers    map[string]solver
+	solvers    map[Challenge]solver
 }
 
 // NewClient creates a new ACME client on behalf of the user. The client will depend on
@@ -93,11 +93,26 @@ func NewClient(caDirURL string, user User, keyBits int) (*Client, error) {
 	// REVIEW: best possibility?
 	// Add all available solvers with the right index as per ACME
 	// spec to this map. Otherwise they won`t be found.
-	solvers := make(map[string]solver)
-	solvers["http-01"] = &httpChallenge{jws: jws, validate: validate}
-	solvers["tls-sni-01"] = &tlsSNIChallenge{jws: jws, validate: validate}
+	solvers := make(map[Challenge]solver)
+	solvers[HTTP01] = &httpChallenge{jws: jws, validate: validate}
+	solvers[TLSSNI01] = &tlsSNIChallenge{jws: jws, validate: validate}
 
 	return &Client{directory: dir, user: user, jws: jws, keyBits: keyBits, solvers: solvers}, nil
+}
+
+// SetChallengeProvider specifies a custom provider that will make the solution available
+func (c *Client) SetChallengeProvider(challenge Challenge, p ChallengeProvider) error {
+	switch challenge {
+	case HTTP01:
+		c.solvers[challenge] = &httpChallenge{jws: c.jws, validate: validate, provider: p}
+	case TLSSNI01:
+		c.solvers[challenge] = &tlsSNIChallenge{jws: c.jws, validate: validate, provider: p}
+	case DNS01:
+		c.solvers[challenge] = &dnsChallenge{jws: c.jws, provider: p}
+	default:
+		return fmt.Errorf("Unknown challenge %v", challenge)
+	}
+	return nil
 }
 
 // SetHTTPAddress specifies a custom interface:port to be used for HTTP based challenges.
@@ -109,9 +124,8 @@ func (c *Client) SetHTTPAddress(iface string) error {
 		return err
 	}
 
-	if chlng, ok := c.solvers["http-01"]; ok {
-		chlng.(*httpChallenge).iface = host
-		chlng.(*httpChallenge).port = port
+	if chlng, ok := c.solvers[HTTP01]; ok {
+		chlng.(*httpChallenge).provider = &httpChallengeServer{iface: host, port: port}
 	}
 
 	return nil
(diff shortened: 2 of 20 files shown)


xenolf:
Hey there :)
Thank you for your contribution!

The change in itself looks sound. What I'm thinking about at the moment is though if it would make sense to have a general interface to all challenges along this implementation.
We will need something like this for DNS-01 and TLS-SNI-01 as well. Some discussion about this is going on in #36.

jehiah:
@xenolf can you take another look. This incorporates feedback from #36 and exports a generic interface which (i think) will work for all challenges.

```go
type ChallengeProvider interface {
 	Present(domain, token, keyAuth string) error
 	CleanUp(domain string)
 }
```

Then there is a un-exported https and tls provider (ie: the http/tls servers that are the default providers for this package), and an exported function to set a custom providers, and exported helpers build the path for the `http-01` challenge, and the cert for the `tls-sni-01` challenge.

I'm of the opinion that this package should not export `ProvidersOpts` (as mentioned in #36) as that is a concept that concrete  implementations need to care about; this package just exports the interface any provider needs to implement to be callable by the appropriate challenge setup/teardown before the solver is run. 

I've tried to keep things named consistently, but if you have better naming suggestions on anything, just comment. I'm also happy to apply this refactoring to the dns provider when that lands in master

jehiah:
@xenolf I see you've started landing support for `dns-01` challenge in master. I've rebased this and applied the same `ChallengeProvider` interface to the DNS code contained within. I'm not perfectly confident my changes are perfect as i don't have a test setup for the three DNS providers that are currently implemented, but i look forward to your feedback either way.

acme/client.go
```diff
@@ -100,6 +100,24 @@ func NewClient(caDirURL string, user User, keyBits int) (*Client, error) {
 	return &Client{directory: dir, user: user, jws: jws, keyBits: keyBits, solvers: solvers}, nil
 }
 
+// SetChallengeProvider specifies a custom provider that will make the solution available
+func (c *Client) SetChallengeProvider(challenge string, p ChallengeProvider) {
+	switch challenge {
+	case "http-01":
```
xenolf: I think when going down this route we should export constants for these identifiers. Like `acme.HTTP01`.

acme/client.go
```diff
@@ -100,6 +100,24 @@ func NewClient(caDirURL string, user User, keyBits int) (*Client, error) {
...
+		if chlng, ok := c.solvers[challenge]; ok {
+			chlng.(*httpChallenge).provider = p
+		}
+	case "tls-sni-01":
+		if chlng, ok := c.solvers[challenge]; ok {
+			chlng.(*tlsSNIChallenge).provider = p
+		}
+	case "dns-01":
+		if chlng, ok := c.solvers[challenge]; ok {
```
janeczku: Since solvers is not exported and dns-01 is not used by the default client this won't work though, would it?

acme/provider.go
```diff
@@ -0,0 +1,7 @@
+package acme
+
+// ChallengeProvider makes the result of a challenge available to be solved
+type ChallengeProvider interface {
+	Present(domain, token, keyAuth string) error
+	CleanUp(domain string) error
```
jehiah: @xenolf a question for you is how you feel about the `CleanUp` signature here

Having only domain seems intuitively sufficient, but it turned out to be slightly awkward for the DNS provider as they need to know exact records to cleanup, and that meant they had to cache some data locally. If you would rather i switch this signature to be `CleanUp(domain, token, keyAuth string) errror` it would clean up that case.

xenolf: Yes, the second one would be the better signature imo. I feel like it shouldn't be required for providers to cache any of the information the Present hook provides.

@jehiah I'm sorry that it's taking me that long to review this, but I'm getting there, I promise :)

jehiah: cool. i'll update shortly.

no worries i maintain enough open source projects that i know the routine from both sides =)

acme/client.go
jehiah: good idea; update forthcoming

acme/client.go
xenolf: That's right. DNS-01 would never work in that case. I think if the check does not work out a new Challenge should get inserted with the choosen solver.

acme/provider.go
jehiah: resolved

acme/client.go
jehiah: resolved

acme/client.go
jehiah: @xenolf I wasn't sure how you intended `dnsChallenge` to get exposed, but I think this works. If you do `SetChallengeProvider(acme.DNS01, ....)` it will now wrap the provider in `dnsChallenge` as needed.

resolved.

note: i don't think this same situation applies to http01 & tls-sni-01 as well since they start out as present in `Client`. If my assumption is wrong i can make the same change there.

xenolf: It does not apply for the default `NewClient` you are right. But I want to add a `New...` function which does not add any default solvers but lets the user configure it completely to his liking. For this case it would be great to have the functionality already there.

jehiah: Actually now that provider state is separate from the challenge solver; it's simpler to just create a new challenge object for the new provider in each of these cases. 

resolved again =)

acme/challenges.go
```diff
@@ -0,0 +1,12 @@
+package acme
+
+type Challenge string
+
+var (
+	// HTTP01 is the "http-01" ACME challenge https://github.com/ietf-wg-acme/acme/blob/master/draft-ietf-acme-acme.md#http
+	HTTP01 = Challenge("http-01")
```
xenolf: I think those should be consts.

acme/dns_challenge.go
```diff
@@ -19,37 +20,49 @@ var preCheckDNS preCheckDNSFunc = checkDNS
 
 var preCheckDNSFallbackCount = 5
 
-// DNSProvider represents a service for managing DNS records.
-type DNSProvider interface {
-	CreateTXTRecord(fqdn, value string, ttl int) error
-	RemoveTXTRecord(fqdn, value string, ttl int) error
+// DNSRecord returns the DNS record which will fulfill a DNS challenge
+func DNSRecord(domain, keyAuth string) (fqdn string, value string, ttl int) {
```
xenolf: This is expected to get called from Solver implementations, right? In that case we might need to document that fact on the interface.

jehiah: I'm not sure i follow. Can you be more specific which interface should have comments/docstrings mentioning this function? Or are you just saying this should be more specific that it's for the `dns-01` challenge?

(There is a similar helper functions for generating a cert for `tls-sni-01`)

xenolf: Oh yes, right... the TLS-SNI-01 function for the cert needs docs as well.

What I meant was it should be clear that client implementations need to call these functions in order to get the needed information for the challenge. I think people may not necessarily read the implementations of the challenges while trying to write their own.

I guess the `ChallengeProvider` might be the place to document this.

jehiah: resolved.

it seemed better to add the comment on each individual `Challenge` type instead of the generic `ChallengeProvider` interface

jehiah:
@xenolf let me know if you have any other feedback here, or if you'd like me to squash what's here for merge.

xenolf:
Nah I think I'm happy now ;)
Squash it and I'll merge it in.

jehiah:
squashed

Issue #79, opened 2016-01-22
ACME challenge route53 does not build on 32bit


denquixote:
den@abdim:lego$ git log -1
    commit 5992793edd44b2adbf095c1bc70245bb2a440b2c
    Author: xenolf <xenolf@users.noreply.github.com>
    Date:   Fri Jan 22 02:25:27 2016 +0100
    
        Refactor DNS precheck
    
    den@abdim:lego$ go version
    go version go1.5.3 linux/386
    
    den@abdim:lego$ go build
    # github.com/xenolf/lego/acme
    acme/dns_challenge_route53.go:66: constant 9223372036854775807 overflows int

PR #80, opened 2016-01-22
Making sure we get all zones from Route53

janeczku:
Kind of a followup to #79 (you beat me to it):
Apparently Route53 never returns more than 100 zones even if we request a higher count. This PR makes sure that we get all available zones when there are more then 100.


xenolf:
> Kind of a followup to #79 (you beat me to it):

I'm sorry, I didn't know if you were watching or not :)

PR #82, opened 2016-01-24
Fix the example in the README

tsileo:
The change fixes the example in the README.

`ObtainCertificate` returns a map of "failures" and not a standard error as second return value.


mholt:
Oops, thanks!

PR #83, opened 2016-01-25
Mark DNS challenge as completed in the README

weppos:
It looks like it was merged in 602aeba6c1abc035f4f17a4e6863f616aa574c24. Am I correct?

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ Current features:
   - [x] HTTP (http-01)
   - [x] TLS with Server Name Indication (tls-sni-01)
   - [ ] Proof of Possession of a Prior Key (proofOfPossession-01)
-  - [ ] DNS (dns-01) - Implemented in branch, blocked by upstream.
+  - [x] DNS (dns-01)
 - [x] Certificate bundling
 - [x] Library support for OCSP
 


mholt:
Merged into master, yes. However, upstream still has a bug that prevents it from working even with the staging endpoint. We expect that will be fixed tomorrow.

weppos:
@mholt Do you have a specific reference I can read to learn more? I'm investigating to add a new DNS provider to use the library, is there anything specific I have to keep in mind?

Thanks for your work!

mholt:
@weppos Yeah, gimme a little bit to write a quick guide in the wiki and I'll link you to it when it's ready. It's pretty straightforward but there are a few things to know.

weppos:
@mholt no rush. I can dig into the existing cloudflare and route53 implementations and start from them. I was actually curious to better understand what is the bug you mentioned that prevents the lib to work with staging. 

Just in case I'll came across it. I run the test locally, and the all passed.

xenolf:
The bug in question is https://github.com/letsencrypt/boulder/issues/1391. 
It is already fixed upstream but was not yet deployed to either staging nor live.

The tests pass because we are not testing against a live boulder instance :)

weppos:
Oh, right. I already hit that issue a few days ago, given I work for one of the 3 DNS providers mentioned in the LE forum where the issue was reproducible. The good news is that the issue is not blocking me to develop the adapter for LEGO.

Thanks for your swift responses! I really appreciate your support.

mholt:
Here's a tutorial: https://github.com/xenolf/lego/wiki/Writing-a-Challenge-Solver - let me know if it's confusing or unclear or just make changes yourself :stuck_out_tongue:

PR #89, opened 2016-01-30
dns_challenge preCheckDNS: let system resolver decide IPv4 ./. IPv6.

### LINKED ISSUE #88: IPv6-only support
Pending [Let's Encrypts's IPv6 support](https://github.com/letsencrypt/boulder/milestones/IPv6%20Support) (IPv6-only hosts), I guess bootstrapping off a [magic IPv4 DNS forwarder](https://github.com/xenolf/lego/blob/d70e2869d257e5301ec9940ad8e95ea8afaad2c5/acme/dns_challenge.go#L76) is going to be a problem...

    /*  .../lego/acme/dns_challenge.go */
    func checkDNS(domain, fqdn string) bool {
    	// check if the expected DNS entry was created. If not wait for some time and try again.
    	m := new(dns.Msg)
    	m.SetQuestion(domain+".", dns.TypeSOA)
    	c := new(dns.Client)
    	in, _, err := c.Exchange(m, "8.8.8.8:53")  // <= ouch
    	if err != nil {
    		return false
    	}
    	// ...

There are systems out there that let only certain DNS forwarders through their firewalls (to mitigate cache poison attacks etc), so maybe it would be an idea to provide a configurable forwarder?

denquixote:
We can ask the OS resolver for the IP of Google's public anycast DNS.
No need to "bootstrap" with literal IP address. The OS resolver knows
best about IPv4 ./. IPv6.

Mostly fixes #88.

### DIFF
--- a/acme/dns_challenge.go
+++ b/acme/dns_challenge.go
@@ -73,7 +73,7 @@ func checkDNS(domain, fqdn string) bool {
 	m := new(dns.Msg)
 	m.SetQuestion(domain+".", dns.TypeSOA)
 	c := new(dns.Client)
-	in, _, err := c.Exchange(m, "8.8.8.8:53")
+	in, _, err := c.Exchange(m, "google-public-dns-a.google.com:53")
 	if err != nil {
 		return false
 	}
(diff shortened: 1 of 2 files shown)


mholt:
I didn't even know about that domain, I always just used the IP address :stuck_out_tongue_winking_eye:

Thanks for the PR

PR #90, opened 2016-01-30
Fix bugs with subdomains

weppos:
While running a few tests, I discovered two bugs: one related to the use of subdomains, and the other related to a very subtle edge case with multiple records.

Each bug is described in the corresponding commit message, along with the successful test message.


xenolf:
The dreaded `:=` scoping :D that got quite a few times as well.

Thanks for fixing those issues!

PR #96, opened 2016-02-03
Refactor DNS check

### LINKED ISSUE #95: NXDOMAIN errors in DNS challenge
Had to hardcode a 5 sec sleep after creating the TXT record because of frequent errors:

> acme: Error 0 - urn:acme:error:connection - DNS problem: NXDOMAIN looking up TXT for _acme-challenge.xxxx.xxxx.com

We need to tweak DNS check here a bit: https://github.com/xenolf/lego/blob/master/acme/dns_challenge.go#L70

We should not rely on the SOA record content. This gives us only the master server of the zone - which might not even be part of the NS record that recursive resolvers query to find the nameserers. Boulder uses unbound for that purpose (https://github.com/letsencrypt/boulder/issues/1112#issuecomment-171111436).
So the record might well be available at the master server when we query it, but this does not mean that the record has (already) been propagated to the actual nameservers that Boulders recursive resolver will query. We should therefore lookup the NS record and use the nameservers listed therein. Maybe query them all.

janeczku:
Fixes #95 

* Implements a (semi-)~~recursive~~ iterative lookup mechanism to find all authoritative
nameservers for the zone enclosing the domain.
This is necessary because a direct query for NS records at a recursive DNS server (e.g. Google Public DNS) or at the zone's primary nameserver will often yield an incomplete list of authoritative nameservers or even those of a another zone (e.g.
when the domain is a CNAME pointing to another zone).
* All authoritative nameservers are then queried for the TXT record to make
sure that it has been completely propagated.
* The waitFor utility function now takes an interval parameter and returns
the last seen error on timeout.

### DIFF
--- a/acme/dns_challenge.go
+++ b/acme/dns_challenge.go
@@ -6,17 +6,18 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"net"
 	"strings"
 	"time"
 
 	"github.com/miekg/dns"
 )
 
-type preCheckDNSFunc func(domain, fqdn string) bool
+type preCheckDNSFunc func(domain, fqdn, value string) error
 
-var preCheckDNS preCheckDNSFunc = checkDNS
+var preCheckDNS preCheckDNSFunc = checkDnsPropagation
 
-var preCheckDNSFallbackCount = 5
+var recursionMaxDepth = 10
 
 // DNS01Record returns a DNS record which will fulfill the `dns-01` challenge
 func DNS01Record(domain, keyAuth string) (fqdn string, value string, ttl int) {
@@ -60,50 +61,141 @@ func (s *dnsChallenge) Solve(chlng challenge, domain string) error {
 		}
 	}()
 
-	fqdn, _, _ := DNS01Record(domain, keyAuth)
+	fqdn, value, _ := DNS01Record(domain, keyAuth)
 
-	preCheckDNS(domain, fqdn)
+	logf("[INFO][%s] Checking DNS record propagation...", domain)
+
+	if err = preCheckDNS(domain, fqdn, value); err != nil {
+		return err
+	}
 
 	return s.validate(s.jws, domain, chlng.URI, challenge{Resource: "challenge", Type: chlng.Type, Token: chlng.Token, KeyAuthorization: keyAuth})
 }
 
-func checkDNS(domain, fqdn string) bool {
-	// check if the expected DNS entry was created. If not wait for some time and try again.
-	m := new(dns.Msg)
-	m.SetQuestion(domain+".", dns.TypeSOA)
-	c := new(dns.Client)
-	in, _, err := c.Exchange(m, "google-public-dns-a.google.com:53")
+// checkDnsPropagation checks if the expected DNS entry has been propagated to
+// all authoritative nameservers. If not it waits and retries for some time.
+func checkDnsPropagation(domain, fqdn, value string) error {
+	authoritativeNss, err := lookupNameservers(toFqdn(domain))
 	if err != nil {
-		return false
+		return err
 	}
 
-	var authorativeNS string
-	for _, answ := range in.Answer {
-		soa := answ.(*dns.SOA)
-		authorativeNS = soa.Ns
+	if err = waitFor(30, 2, func() (bool, error) {
+		return checkAuthoritativeNss(fqdn, value, authoritativeNss)
+	}); err != nil {
+		return err
 	}
 
-	fallbackCnt := 0
-	for fallbackCnt < preCheckDNSFallbackCount {
-		m.SetQuestion(fqdn, dns.TypeTXT)
-		in, _, err = c.Exchange(m, authorativeNS+":53")
+	return nil
+}
+
+// checkAuthoritativeNss checks whether a TXT record with fqdn and value exists on every given nameserver.
+func checkAuthoritativeNss(fqdn, value string, nameservers []string) (bool, error) {
+	for _, ns := range nameservers {
+		r, err := dnsQuery(fqdn, dns.TypeTXT, net.JoinHostPort(ns, "53"))
 		if err != nil {
-			return false
+			return false, err
 		}
 
-		if len(in.Answer) > 0 {
-			return true
+		if r.Rcode != dns.RcodeSuccess {
+			return false, fmt.Errorf("%s returned RCode %s", ns, dns.RcodeToString[r.Rcode])
 		}
 
-		fallbackCnt++
-		if fallbackCnt >= preCheckDNSFallbackCount {
-			return false
+		var found bool
+		for _, rr := range r.Answer {
+			if txt, ok := rr.(*dns.TXT); ok {
+				if strings.Join(txt.Txt, "") == value {
+					found = true
+					break
+				}
+			}
 		}
-
-		time.Sleep(time.Second * time.Duration(fallbackCnt))
+		if !found {
+			return false, fmt.Errorf("%s did not return the expected TXT record", ns)
+		}
 	}
 
-	return false
+	return true, nil
+}
+
+// dnsQuery directly queries the given authoritative nameserver.
+func dnsQuery(fqdn string, rtype uint16, nameserver string) (in *dns.Msg, err error) {
+	m := new(dns.Msg)
+	m.SetQuestion(fqdn, rtype)
+	m.SetEdns0(4096, false)
+	m.RecursionDesired = false
+
+	in, err = dns.Exchange(m, nameserver)
+	if err == dns.ErrTruncated {
+		tcp := &dns.Client{Net: "tcp"}
+		in, _, err = tcp.Exchange(m, nameserver)
+	}
+
+	return
+}
+
+// lookupNameservers returns the authoritative nameservers for the given domain name.
+func lookupNameservers(fqdn string) ([]string, error) {
+	var referralNameservers []string
+
+	// We start recursion at the gTLD origin
+	// so we don't have to manage root hints
+	labels := dns.SplitDomainName(fqdn)
+	tld := labels[len(labels)-1]
+	nss, err := net.LookupNS(tld)
+	if err != nil {
+		return nil, fmt.Errorf("Could not resolve TLD %s %v", tld, err)
+	}
+
+	for _, ns := range nss {
+		referralNameservers = append(referralNameservers, ns.Host)
+	}
+
+	// Follow the referrals until we hit the authoritative NS
+	for depth := 0; depth < recursionMaxDepth; depth++ {
+		var r *dns.Msg
+		var err error
+
+		for _, ns := range referralNameservers {
+			r, err = dnsQuery(fqdn, dns.TypeNS, net.JoinHostPort(ns, "53"))
+			if err != nil {
+				continue
+			}
+
+			if r.Rcode == dns.RcodeSuccess {
+				break
+			}
+
+			if r.Rcode == dns.RcodeNameError  {
+				return nil, fmt.Errorf("Could not resolve NXDOMAIN %s", fqdn)
+			}
+		}
+
+		if r == nil {
+			break
+		}
+
+		if r.Authoritative {
+			// We got an authoritative reply, which means that the
+			// last referral holds the authoritative nameservers.
+			return referralNameservers, nil
+		}
+
+		referralNameservers = nil
+
+		for _, rr := range r.Ns {
+			if ns, ok := rr.(*dns.NS); ok {
+				referralNameservers = append(referralNameservers, strings.ToLower(ns.Ns))
+			}
+		}
+
+		// No referrals to follow
+		if len(referralNameservers) == 0 {
+			break
+		}
+	}
+
+	return nil, fmt.Errorf("Could not determine nameservers for %s", fqdn)
 }
 
 // toFqdn converts the name into a fqdn appending a trailing dot.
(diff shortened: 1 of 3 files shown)


acme/dns_challenge.go
```diff
@@ -62,48 +64,64 @@ func (s *dnsChallenge) Solve(chlng challenge, domain string) error {
...
+			}
 		}
 
-		fallbackCnt++
-		if fallbackCnt >= preCheckDNSFallbackCount {
-			return false
+		if len(in.Ns) > 0 {
+			for _, r := range in.Ns {
+				rr, ok := r.(*dns.NS)
+				if ok && recursionsCnt < preCheckDNSMaxRecursions {
```
xenolf: Guess we should query all NS at some point, right? Instead of artificially limit the number to five.

janeczku:
@xenolf Right now we only check the first NS in the record. The MaxRecursion actually limits the recursion depth and protects us from an infinite loop due to misconfigured DNS zones.
I agree we should query all NS in the records!

janeczku:
We also better add a test case before merging this :smile:

xenolf:
Are you going to add that to your PR or should I add it later? :)

Thanks for you hard work on this issue! :clap:

janeczku:
I will update this with the check for all authoritative NS. If you could add a test when thats done? :smirk:

xenolf:
:laughing: Yeah I will. Them bloody tests :stuck_out_tongue_winking_eye:

janeczku:
@xenolf So the changes for querying all authoritative nameservers are in. This is a bit of a rewrite, so please review. Tested with CloudFlare DNS.

janeczku:
Added some unit tests FYI. :bowtie:

acme/dns_challenge.go
```diff
@@ -124,19 +202,20 @@ func unFqdn(name string) string {
 	return name
 }
 
-// waitFor polls the given function 'f', once per second, up to 'timeout' seconds.
-func waitFor(timeout int, f func() (bool, error)) error {
+// waitFor polls the given function 'f', once every 'interval' seconds, up to 'timeout' seconds.
+func waitFor(timeout, interval int, f func() (bool, error)) error {
```
xenolf: I'm just thinking out loud here... but maybe we should add a util.go for stuff like this?

janeczku: That was on my mind too! :smile: But maybe we can do the code re-organization later in a separate patch?

xenolf: Yeah, I agree... out of scope of this PR :)

janeczku: Do you have any other stuff on your mind that ought to move to the util.go?

acme/dns_challenge.go
```diff
@@ -62,48 +63,125 @@ func (s *dnsChallenge) Solve(chlng challenge, domain string) error {
...
+		}
+
+		if r.Rcode != dns.RcodeSuccess {
+			return false, fmt.Errorf("NS %s returned RCode %s", ns, dns.RcodeToString[r.Rcode])
+		}
+
+		var found bool
+		for _, rr := range r.Answer {
+			if _, ok := rr.(*dns.TXT); ok {
+				found = true
```
xenolf: Maybe also check the value returned by the TXT record here?

acme/dns_challenge.go
xenolf: Yes, limitReader for example. I'm also thinking about crypto.go as there are multiple functions in there which are mere utility functions.

acme/dns_challenge.go
janeczku: Yes

acme/dns_challenge.go
```diff
@@ -62,48 +63,125 @@ func (s *dnsChallenge) Solve(chlng challenge, domain string) error {
...
+		
+		for _, ns := range referralNameservers {
+			r, err = dnsQuery(fqdn, dns.TypeNS, net.JoinHostPort(ns, "53"))
+			if err != nil {
+				continue
+			}
+			if r.Rcode == dns.RcodeSuccess {
+				break
+			}
+			if r.Rcode == dns.RcodeNameError {
```
willglynn: Consider:

    $ dig any internal-name.willglynn.com @8.8.8.8

    ; <<>> DiG 9.8.3-P1 <<>> any internal-name.willglynn.com @8.8.8.8
    ;; global options: +cmd
    ;; Got answer:
    ;; ->>HEADER<<- opcode: QUERY, status: NXDOMAIN, id: 39092
    ;; flags: qr rd ra; QUERY: 1, ANSWER: 0, AUTHORITY: 1, ADDITIONAL: 0

    ;; QUESTION SECTION:
    ;internal-name.willglynn.com.	IN	ANY

    ;; AUTHORITY SECTION:
    willglynn.com.		1799	IN	SOA	ns1.willglynn.com. awsdns-hostmaster.amazon.com. 1 7200 900 1209600 300

    ;; Query time: 52 msec
    ;; SERVER: 8.8.8.8#53(8.8.8.8)
    ;; WHEN: Sat Feb  6 23:16:44 2016
    ;; MSG SIZE  rcvd: 110

In this case, querying `internal-name.willglynn.com` fails with `NXDOMAIN`, causing `lookupNameservers()` to exit at this location each time it's called. This ultimately causes `lego` to spin forever, long after the desired `_acme-challenge.internal-name.willglynn.com IN TXT` record became globally resolvable from `ns1`..`ns4.willglynn.com`.

A logging statement confirms that `referralNameservers` contains `ns1`..`ns4.willglynn.com` at the moment this branch exits the function, which is the correct set of authoritative servers for that FQDN.

acme/dns_challenge.go
```diff
@@ -60,50 +61,129 @@ func (s *dnsChallenge) Solve(chlng challenge, domain string) error {
...
 
 	return s.validate(s.jws, domain, chlng.URI, challenge{Resource: "challenge", Type: chlng.Type, Token: chlng.Token, KeyAuthorization: keyAuth})
 }
 
-func checkDNS(domain, fqdn string) bool {
-	// check if the expected DNS entry was created. If not wait for some time and try again.
+// checkDnsPropagation queries all nameservers that are authoritative for 'domain'
+// to check whether DNS propagation for the expected DNS record is complete.
+func checkDnsPropagation(domain, fqdn, value string) (bool, error) {
+	nameservers, err := lookupNameservers(toFqdn(domain))
```
willglynn: I think this should be `lookupNameservers(toFqdn(fqdn))` instead.

Scenario: a request for `domain = "foo.bar.com"`, which would have `fqdn = "_acme-challenge.foo.bar.com"`.

Assuming `bar.com IN NS ns1.bar.com`, `lookupNameservers("foo.bar.com.")` would return `["ns1.bar.com"]`, and the function would query `ns1.bar.com` for `_acme-challenge.foo.bar.com`. This is the common case.

However, `ns1.bar.com` could have a record `_acme-challenge.foo.bar.com IN NS ns.confusion.party` – delegating *that specific name* to a different nameserver, in which case this function should query `ns.confusion.party` instead.

`lookupNameservers(toFqdn(fqdn))` would be able to find this `NS` record and follow the delegation, while `lookupNameservers(toFqdn(domain))` would not.

acme/dns_challenge.go
janeczku: Hi Willy! Thank your for reviewing the code! :smiley: 

> This ultimately causes lego to spin forever

No it won't. Lego will gracefully exit after 60 seconds, which should be plenty of time for the user  to think about why he is trying to issue a certificate for a nonexistent domain name.

While it might be technically legal to do that under the ACME spec, i don't see why we should specifically enable this behavior. In fact none of the other challenge methods support this scenario.

This being said, i do agree that the very purpose of the `checkDnsPropagation` func is to verify the validation domain rather than the certificate domain. Hence, we should pass the fqdn in the `lookupNameservers` func, which incidentally should solve the usage scenario you described.

willglynn: In my use case, `internal-name.willglynn.com` isn't nonexistent, it's just not visible to public resolvers due to [split horizon DNS](https://en.wikipedia.org/wiki/Split-horizon_DNS). The `_acme-challenge` record obviously needs to be visible, but there's no constraint on `internal-name` itself, hence `lookupNameservers()` queries returning `NXDOMAIN`.

As for spinning forever, I see the timeout-related code, but… `lego` does *not* time out after a minute in this branch. I'm not sure where it's getting stuck (nor is there any obvious way to get verbose output), but after several minutes of waiting, `tshark` shows my machine repeatedly querying my local recursive resolver for `com. IN NS`:

    2016/02/07 01:25:32 [INFO][internal-name.willglynn.com] acme: Obtaining bundled SAN certificate
    2016/02/07 01:25:32 [INFO][internal-name.willglynn.com] acme: Could not find solver for: http-01
    2016/02/07 01:25:32 [INFO][internal-name.willglynn.com] acme: Could not find solver for: tls-sni-01
    2016/02/07 01:25:32 [INFO][internal-name.willglynn.com] acme: Trying to solve DNS-01
    … wait a long time …
    … switch terminals …
    $ tshark -t ad -i en0 'port 53'
    Capturing on 'Wi-Fi'
      1 2016-02-07 01:34:42.914351 172.16.3.100 -> 172.16.3.1   DNS 63 Standard query 0xc0f5 NS com
      2 2016-02-07 01:34:42.920370   172.16.3.1 -> 172.16.3.100 DNS 565 Standard query response 0xc0f5 NS m.gtld-servers.netNS c.gtld-servers.netNS d.gtld-servers.netNS b.gtld-servers.netNS e.gtld-servers.netNS h.gtld-servers.netNS g.gtld-servers.netNS l.gtld-servers.netNS f.gtld-servers.netNS a.gtld-servers.netNS k.gtld-servers.netNS i.gtld-servers.netNS j.gtld-servers.net
      3 2016-02-07 01:34:42.920926 172.16.3.100 -> 172.16.3.1   DNS 63 Standard query 0x4ef1 NS com
      4 2016-02-07 01:34:42.927413   172.16.3.1 -> 172.16.3.100 DNS 565 Standard query response 0x4ef1 NS d.gtld-servers.netNS b.gtld-servers.netNS e.gtld-servers.netNS h.gtld-servers.netNS g.gtld-servers.netNS l.gtld-servers.netNS f.gtld-servers.netNS a.gtld-servers.netNS k.gtld-servers.netNS i.gtld-servers.netNS j.gtld-servers.netNS m.gtld-servers.netNS c.gtld-servers.net

janeczku: Need to investigate this. Thanks for digging deep! :+1:

acme/dns_challenge.go
janeczku: >  ns1.bar.com could have a record _acme-challenge.foo.bar.com IN NS ns.confusion.party – delegating that specific name to a different nameserver, in which case this function should query ns.confusion.party instead.

No, i don't think it should. Because `ns.confusion.party` is **not** the authority for the certificate domain ` foo.bar.com`.

The purpose of the ACME challenge is to prove ownership of the domain for which a certificate is requested. In your scenario you don't show authority for the certificate domain `foo.bar.com` because you did not create a record in the `bar.com` zone. `ns1.bar.com` might have delegated `_acme-challenge.foo.bar.com` to you, but that does not mean that you have authority over `bar.com`.

janeczku:
@willglynn If you will, please test your scenario against the updated PR. Thanks!

acme/dns_challenge.go
willglynn: I am dissatisfied with this explanation for three reasons.

First, by looking up nameservers for `domain` instead of `fqdn`, `lego` will be checking for the `_acme-challenge IN TXT` record differently than normal recursive resolvers, which would follow such a delegation. This includes `boulder`, which [implements DNS01](https://github.com/letsencrypt/boulder/blob/e23cf3991ec298e9330a87394b82cceed862ffb5/va/validation-authority.go#L432-L434) by [asking a normal resolver to perform a normal recursive lookup](https://github.com/letsencrypt/boulder/blob/0de0ed550b9869792619363586af04cd2873e848/bdns/dns.go#L250-L251).

Second, we agree that the delegation `_acme-challenge.foo.bar.com IN NS ns.confusion.party` and subsequent creation of a `TXT` record at `ns.confusion.party` would not indicate that I have authority over `bar.com`, but I never said it did. Instead, I argue that this record specifically means that `bar.com` has *delegated the authority* to issue ACME certificates for `foo.bar.com` to `ns.confusion.party`. If I can fulfill a DNS-01 challenge by adding a record at `ns.confusion.party`, then I've fulfillled the DNS-01 challenge, full stop.

There are situations where it is undesirable or impractical from a policy/procedure standpoint to modify `bar.com` within short periods of time. (Think of those big companies with those big meetings for change control that needs oodles of review.) By delegating `_acme-challenge` ahead of time to a different nameserver with different controls, it's possible to satisfy `boulder` with a cron job intended to renew certificates, and it's possible to satisfy The Powers That Be™ by eliminating the need to dynamically modify `bar.com` in any way. I believe `lego` should not object.

Finally, the ACME spec does not require that `_acme-challenge.foo.bar.com` be served by the same nameservers as `foo.bar.com`, only that [a query for TXT records under the validation domain name – `_acme-challenge.*` – returns a matching value](https://tools.ietf.org/html/draft-ietf-acme-acme-01#section-7.5).

willglynn:
I should be able to check in on this tomorrow. Who knows; I might even be able to contribute tests :smile:

janeczku:
Awesome :ok_hand: :yum:

willglynn:
I don't have immediate feedback, but I have been collecting scenarios for testing.

(CNAMEs make for all sorts of fun.)

janeczku:
The iterative DNS lookup used in this PR proved overkill, since the objective (getting a definite list of authoritative NS for a domain) can actually very well be accomplished by plain old and cheap recursive queries to a public DNS resolver. :grimacing:
Landing this PR would thus introduce unnecessary complexity.
Closed in favor of #116 

@willglynn The API has not changed so any tests you might have prepared can be applied to the new PR.

Issue #94, opened 2016-02-01
DNS challenge with Route 53 fails with "InvalidTXTRDATA encountered"


jimmycuadra:
Continued from my comments in #2:

```
$ lego --email="user@example.com" --domains="example.com" --dns="route53" --exclude="http-01" --exclude="tls-sni-01" run
2016/01/29 23:06:03 [INFO][example.com] acme: Obtaining bundled SAN certificate
2016/01/29 23:06:03 [INFO][example.com] acme: Could not find solver for: http-01
2016/01/29 23:06:03 [INFO][example.com] acme: Could not find solver for: tls-sni-01
2016/01/29 23:06:03 [INFO] acme: Trying to solve DNS-01
2016/01/29 23:06:04 [example.com] Could not obtain certificates
    Error presenting token Request failed, got status code: 400. Response: <?xml version="1.0"?>
<ErrorResponse xmlns="https://route53.amazonaws.com/doc/2013-04-01/"><Error><Type>Sender</Type><Code>InvalidChangeBatch</Code><Message>Invalid Resource Record: FATAL problem: InvalidTXTRDATA encountered at hcpTWGxkPEGUMxJMPsXtsPE0TSflliWqAKH1FJ8crvg</Message></Error><RequestId>REDACTED</RequestId></ErrorResponse>
```

This is with lego built from source at commit 031c5b5 (HEAD as of right now).

janeczku:
@jimmycuadra The fix is not in the HEAD. Please built from this branch and report back: https://github.com/xenolf/lego/tree/route53-literal-quotes

jimmycuadra:
Ah, my mistake! Using that branch:

```
lego --email="user@example.com" --domains="example.com" --dns="route53" --exclude="http-01" --exclude="tls-sni-01" run
2016/02/02 19:03:28 [INFO][example.com] acme: Obtaining bundled SAN certificate
2016/02/02 19:03:28 [INFO][example.com] acme: Could not find solver for: tls-sni-01
2016/02/02 19:03:28 [INFO][example.com] acme: Could not find solver for: http-01
2016/02/02 19:03:28 [INFO][example.com] acme: Trying to solve DNS-01
2016/02/02 19:03:31 Error cleaning up example.com Request failed, got status code: 400. Response: <?xml version="1.0"?>
[...]

PR #98, opened 2016-02-05
Add full CLI example and IAM policy for Route 53 to the README.

jimmycuadra:
Following up from #94 with additional docs!


xenolf:
Thank you for this! :)

PR #101, opened 2016-02-06
Allow to customize the TSIG algorithm and allow to omit the DNS port

pkern:
The TSIG algorithm is currently hardcoded as HMAC-MD5. Any other key type cannot be used. Let's add an environment variable that allows to pass through the pseudo domain name used for the HMAC type.

Also append the default DNS port to the target nameserver to be updated, in case none is specified. This is caused by the Go API wanting to know the port for the connection, but we know a sensible destination default.

### DIFF
--- a/acme/dns_challenge_rfc2136.go
+++ b/acme/dns_challenge_rfc2136.go
@@ -2,30 +2,36 @@ package acme
 
 import (
 	"fmt"
-	"github.com/miekg/dns"
 	"time"
+
+	"github.com/miekg/dns"
 )
 
 // DNSProviderRFC2136 is an implementation of the ChallengeProvider interface that
 // uses dynamic DNS updates (RFC 2136) to create TXT records on a nameserver.
 type DNSProviderRFC2136 struct {
-	nameserver string
-	zone       string
-	tsigKey    string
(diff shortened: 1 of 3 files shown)


acme/dns_challenge_rfc2136.go
```diff
@@ -2,31 +2,42 @@ package acme
...
+		nameserver += ":53"
+	}
 	d := &DNSProviderRFC2136{
 		nameserver: nameserver,
 		zone:       zone,
 		records:    make(map[string]string),
 	}
-	if len(tsigKey) > 0 && len(tsigSecret) > 0 {
+	if len(tsigAlgorithm) > 0 && len(tsigKey) > 0 && len(tsigSecret) > 0 {
+		if tsigAlgorithm == "" {
```
xenolf: This case is always false as we do not enter this branch if `len(tsigAlgorithm)` is zero.

xenolf:
Thank you for your contribution :)

I added an inline comment for you to review.

acme/dns_challenge_rfc2136.go
pkern: Fixed. Thanks. :)

PR #102, opened 2016-02-07
Add support for additional AWS authentication sources

willglynn:
AWS client tools commonly support passing credentials via `AWS_ACCESS_KEY_ID` + `AWS_SECRET_ACCESS_KEY`, but supporting only this is insufficient. For example, access key IDs provided by STS require passing in `AWS_SECURITY_TOKEN` as a third value, and EC2 instances are often provided dynamic credentials at runtime via the EC2 metadata service.

This changeset makes `lego` attempt to find credentials in the same way that the `aws` CLI tool attempts to find credentials. The result is even less auth code than before because `goamz` provides all this with `aws.GetAuth()`.

### DIFF
--- a/acme/dns_challenge_route53.go
+++ b/acme/dns_challenge_route53.go
@@ -15,29 +15,29 @@ type DNSProviderRoute53 struct {
 }
 
 // NewDNSProviderRoute53 returns a DNSProviderRoute53 instance with a configured route53 client.
-// Authentication is either done using the passed credentials or - when empty -
-// using the environment variables AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY.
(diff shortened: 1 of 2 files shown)


willglynn:
Note also that this proposed behavior matches the official AWS SDK's [default credential chain](https://github.com/aws/aws-sdk-go/blob/master/aws/defaults/defaults.go#L78-L96), customarily achieved by [omitting the `Credentials` field of `aws.Config`](http://docs.aws.amazon.com/sdk-for-go/api/aws/Config.html#Credentials-field).

janeczku:
LGTM :+1:

mholt:
I like that this PR is a net gain in LOC because of comments, but is actually a net reduction.

xenolf:
Great PR, thank you!

PR #104, opened 2016-02-07
GetOCSPForCert should fail if there are no OCSP servers in the cert.

chrismoos:
This issue came up on another project, see: https://github.com/mholt/caddy/pull/579


mholt:
Great - thank you!

Issue #100, opened 2016-02-06
Imports & Dependencies


jehiah:
I sweat dependency trees, and really enjoy light libraries. One of the initial appeals to me of `lego` over [`hlandau/acme`](https://github.com/hlandau/acme) was how light this project was in terms of dependencies to accomplish what i needed. :heart: 

Unfortunately, with the current code structure, the addition of support for multiple DNS providers to solve the `dns-01`  challenge has increased the dependency tree :see_no_evil:.

To import `github.com/hlandau/acme` the current dependency tree you need to support is 

```
$ go list -f '{{join .Deps "\n"}}' |  xargs go list -f '{{if not .Standard}}{{.ImportPath}}{{end}}'
github.com/crackcomm/cloudflare
github.com/miekg/dns
github.com/mitchellh/goamz/aws
github.com/mitchellh/goamz/route53
github.com/square/go-jose
github.com/square/go-jose/cipher
github.com/vaughan0/go-ini
github.com/weppos/dnsimple-go/dnsimple
golang.org/x/crypto/ocsp
golang.org/x/crypto/sha3
golang.org/x/net/context
```

Of these, I'd love to skip 6 of them (`vaughan0/go-ini` is a sub dependency of `mitchellh/goamz/route53`), so I'd like to propose two options for refactoring the code structure to make these dependencies separate from the main `acme` package.

a) Move all `providers` into a single `xenolf/lego/providers` package
b) Move per-challenge pacakges (ie: `xenolf/lego/providers/dns01` `xenolf/lego/providers/http01` etc)
c) move to per-provider packages (ie: `xenolf/lego/providers/dns/cloudflare`, `xenolf/lego/providers/dnsimple`, ...)
d) some combination of the above? 

(Note: i'm proposing structure, not specific naming). Since i personally don't care about the DNS providers, 2 and 3 are equal for me. I imagine others might feel differently, and the ideal case might change when more options w/ dependencies are added (like say #28).

The key is that the `acme` package doesn't import these packages (because that makes them all a dependency), the cli code would import them and set them as appropriate on the `Client`. Depending on the approach, the default http and tls-sni providers may or may not be moved.

thoughts @xenolf ?

xenolf:
Yes... I also do not feel too good with all these dependencies directly in the acme package.

My thought was that I first wanted to get the feature stable and then do a clean up to structure it in a way I feel comfortable with.

As we progress on DNS-01 I think it is a good time to start thinking about this.

[...]

PR #105, opened 2016-02-07
Add Dyn DNS provider

aebruno:
This commit adds support for Dyn, specifically Dyn Managed DNS. Makes
use of the Dyn REST API [1] and requires 3 environment variables for
credentails:

    DYN_CUSTOMER_NAME
    DYN_USER_NAME
    DYN_PASSWORD

The Dyn DNS API requires a session to be established first using the
credentials above, After creating a session an api token is used for
subsequent requests.

The unit test performs a "live test" and currently requires a valid Dyn
account and domain.

[1] https://help.dyn.com/rest/

### DIFF
--- a/providers/dns/dyn/dyn.go
+++ b/providers/dns/dyn/dyn.go
@@ -14,7 +14,7 @@ import (
 
 var dynBaseURL = "https://api.dynect.net/REST"
 
-type DynResponse struct {
+type dynResponse struct {
 	// One of 'success', 'failure', or 'incomplete'
 	Status string `json:"status"`
 
@@ -22,7 +22,7 @@ type DynResponse struct {
 	Data json.RawMessage `json:"data"`
 
 	// The ID of the job that was created in response to a request.
-	JobId int `json:"job_id"`
+	JobID int `json:"job_id"`
 
 	// A list of zero or more messages
 	Messages json.RawMessage `json:"msgs"`
@@ -48,7 +48,7 @@ func NewDNSProvider(customerName, userName, password string) (*DNSProvider, erro
 	}, nil
 }
 
-func (d *DNSProvider) sendRequest(method, resource string, payload interface{}) (*DynResponse, error) {
+func (d *DNSProvider) sendRequest(method, resource string, payload interface{}) (*dynResponse, error) {
 	url := fmt.Sprintf("%s/%s", dynBaseURL, resource)
 
 	body, err := json.Marshal(payload)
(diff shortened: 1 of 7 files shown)


xenolf:
Thank you for this!

I'm sorry that it takes me a while to go over this. We are currently discussing how we want to handle external providers going forward in #100 and #111.

aebruno:
Thanks for the great library. I'd be happy to move the dyn provider to a another package or however you end up handling the providers.

xenolf:
Hello there and sorry for the wait :worried: 
#144 was merged just now and that's the new layout.

aebruno:
Hey, I rebased and moved the Dyn DNS provider out of the acme package per #144. Tried to follow the other providers the best I could. Let me know how things look and if you'd like anything changed.

xenolf:
Thanks for your work @aebruno, this is looking quite good. Just one small thing: Could you please use a HTTP client with a timeout like `http.Client{Timeout: 10 * time.Second}` instead of the go default one for your HTTP requests?

aebruno:
All set. Added timeout to HTTP client. Let me know if you spot anything else.

aebruno:
Not sure why the CI tests failed on looking up the authoritative zone?

providers/dns/dyn/dyn.go
```diff
@@ -0,0 +1,249 @@
...
+	"net/http"
+	"strconv"
+	"time"
+
+	"github.com/xenolf/lego/acme"
+)
+
+var dynBaseURL = "https://api.dynect.net/REST"
+
+type DynResponse struct {
```
xenolf: This doesn't need to be exported, right?

providers/dns/dyn/dyn.go
```diff
@@ -0,0 +1,249 @@
...
+	// One of 'success', 'failure', or 'incomplete'
+	Status string `json:"status"`
+
+	// The structure containing the actual results of the request
+	Data json.RawMessage `json:"data"`
+
+	// The ID of the job that was created in response to a request.
+	JobId int `json:"job_id"`
```
xenolf: This should be called `JobID`.

xenolf:
@aebruno the DNS tests seem to be a bit flaky on travis for some reason. I restarted the build and it went through OK.

I just added a few minor nitpicks from running `golint` on your code, other then that this is good to merge :)

aebruno:
@xenolf good catch. I made the dynRequest private to the package and cleaned up the JobID name. Thanks! Anything else just let me know.

PR #108, opened 2016-02-07
http-01 challenge: disable TCP connection reuse

### LINKED ISSUE #107: Reverse proxying to lego requires modifying reverse proxy transport
Hi. I'm not sure if this is a bug or not, but if not it might be worth documenting somewhere if it isn't.

My setup is that I wanted to run lego on a non-standard port (using the --http flag) and have http-01 acme challenges forwarded to lego from my two web servers.

My first attempt at doing so was to add 302 http redirects but for some reason the Let's Encrypt validation servers just wouldn't play ball (even though everyone on the net says http redirects works fine). This was nothing to do with lego, but I just thought I'd mention it.

Having failed at that I decided to use httputil.ReverseProxy from the Go standard library to send on the challenges. This worked great with my first web server which has a single domain name on the certificate. However my second web server, which has a SAN Multi-Domain certificate, failed. The first domain on the certificate validated fine with Let's Encrypt's servers, but the second did not whith lego returning an acme "404" error.

After a bit of head scratching I turned off re-use of TCP connections between different HTTP requests in the transport of the reverse proxy by setting DisableKeepAlives in the transport to true. Then everything worked perfectly! :)

Anyway, I just thought I'd let you guys know in case it was of interest.

xi2:
If TCP connection reuse (KeepAlives) are left on then, in a sequence
of challenges arising from a multiple-domain certficate, challenges
after the 1st can fail due to reusing the now defunct tcp connection
used in the first challenge. This has been observed when using the Go
standard library reverse proxy to forward the challenges to Lego.

Fixes #107

### DIFF
--- a/acme/http_challenge_server.go
+++ b/acme/http_challenge_server.go
@@ -58,6 +58,12 @@ func (s *httpChallengeServer) serve(domain, token, keyAuth string) {
 		}
 	})
 
-	http.Serve(s.listener, mux)
+	httpServer := &http.Server{
+		Handler: mux,
+	}
+	// Once httpServer is shut down we don't want any lingering
+	// connections, so disable KeepAlives.
+	httpServer.SetKeepAlivesEnabled(false)
+	httpServer.Serve(s.listener)
 	s.done <- true
 }


mholt:
LGTM.

Issue #110, opened 2016-02-08
`lego dns-test` feature request


willglynn:
I'd like to be able to run `lego --dns=foo --domains=totally.legit.biz dns-test`, which would:

1. Use the `foo` DNS provider to create a `_acme-challenge.totally.legit.biz` record with a unique value for testing. 
2. Watch and wait for `_acme-challenge.totally.legit.biz` to report that value just like a normal DNS challenge, preferably with extra log output so I can see what it's doing.
3. Clean up after itself.

This would allow users to test and troubleshoot their configuration without either a) pestering live ACME services or b) launching their own local ACME service for testing.

This would be safe to do even if someone else is concurrently answering challenges for the same name because identically-named TXT records can coexist and [DNS-01 requires only one of them to match](https://tools.ietf.org/html/draft-ietf-acme-acme-01#section-7.5):

     To validate a DNS challenge, the server performs the following steps:

     1.  Compute the SHA-256 digest of the key authorization

     2.  Query for TXT records under the validation domain name

     3.  Verify that the contents of one of the TXT records matches the
         digest value

mholt:
Why not just use the staging endpoint? `https://acme-staging.api.letsencrypt.org/directory` This way you don't contact the live staging endpoint or have to launch your own ACME server.

willglynn:
Huh. That… didn't occur to me. Good idea!

PR #113, opened 2016-02-08
Mention the Let's Encrypt staging endpoint in the README

willglynn:
I completely forgot that the Let's Encrypt staging endpoint exists (#110). Might be worth a mention in the README.


mholt:
Great! Thanks.

PR #116, opened 2016-02-09
Refactor DNS check

janeczku:
Replaces #96 

* Gets a list of all authoritative nameservers by looking up the NS RRs for the root domain (zone apex)
* Verifies that the expected TXT record exists on all nameservers before sending off the challenge to ACME server

### DIFF
--- a/acme/dns_challenge.go
+++ b/acme/dns_challenge.go
@@ -13,11 +13,11 @@ import (
 	"github.com/miekg/dns"
 )
 
-type preCheckDNSFunc func(domain, fqdn, value string) error
+type preCheckDNSFunc func(fqdn, value string) (bool, error)
 
 var preCheckDNS preCheckDNSFunc = checkDnsPropagation
 
-var recursionMaxDepth = 10
+var recursiveNameserver = "google-public-dns-a.google.com"
 
 // DNS01Record returns a DNS record which will fulfill the `dns-01` challenge
 func DNS01Record(domain, keyAuth string) (fqdn string, value string, ttl int) {
@@ -65,28 +65,43 @@ func (s *dnsChallenge) Solve(chlng challenge, domain string) error {
 
 	logf("[INFO][%s] Checking DNS record propagation...", domain)
 
-	if err = preCheckDNS(domain, fqdn, value); err != nil {
+	err = waitFor(30, 2, func() (bool, error) {
+		return preCheckDNS(fqdn, value)
+	})
+	if err != nil {
 		return err
 	}
 
 	return s.validate(s.jws, domain, chlng.URI, challenge{Resource: "challenge", Type: chlng.Type, Token: chlng.Token, KeyAuthorization: keyAuth})
 }
 
-// checkDnsPropagation checks if the expected TXT record has been propagated to
-// all authoritative nameservers. If not it waits and retries for some time.
-func checkDnsPropagation(domain, fqdn, value string) error {
-	authoritativeNss, err := lookupNameservers(toFqdn(domain))
+// checkDnsPropagation checks if the expected TXT record has been propagated to all authoritative nameservers.
+func checkDnsPropagation(fqdn, value string) (bool, error) {
+	// Initial attempt to resolve at the recursive NS
+	r, err := dnsQuery(fqdn, dns.TypeTXT, recursiveNameserver, true)
 	if err != nil {
-		return err
+		return false, err
+	}
+	if r.Rcode != dns.RcodeSuccess {
+		return false, fmt.Errorf("Could not resolve %s -> %s", fqdn, dns.RcodeToString[r.Rcode])
 	}
 
-	if err = waitFor(30, 2, func() (bool, error) {
-		return checkAuthoritativeNss(fqdn, value, authoritativeNss)
-	}); err != nil {
-		return err
+	// If we see a CNAME here then use the alias
+	for _, rr := range r.Answer {
+		if cn, ok := rr.(*dns.CNAME); ok {
+			if cn.Hdr.Name == fqdn {
+				fqdn = cn.Target
+				break
+			}
+		}
 	}
 
-	return nil
+	authoritativeNss, err := lookupNameservers(fqdn)
+	if err != nil {
+		return false, err
+	}
+
+	return checkAuthoritativeNss(fqdn, value, authoritativeNss)
 }
 
 // checkAuthoritativeNss queries each of the given nameservers for the expected TXT record.
@@ -98,7 +113,7 @@ func checkAuthoritativeNss(fqdn, value string, nameservers []string) (bool, erro
 		}
 
 		if r.Rcode != dns.RcodeSuccess {
-			return false, fmt.Errorf("%s returned RCode %s", ns, dns.RcodeToString[r.Rcode])
+			return false, fmt.Errorf("NS %s returned %s for %s", ns, dns.RcodeToString[r.Rcode], fqdn)
 		}
 
 		var found bool
@@ -112,7 +127,7 @@ func checkAuthoritativeNss(fqdn, value string, nameservers []string) (bool, erro
 		}
 
 		if !found {
-			return false, fmt.Errorf("%s did not return the expected TXT record", ns)
+			return false, fmt.Errorf("NS %s did not return the expected TXT record", ns)
 		}
 	}
 
@@ -124,6 +139,7 @@ func dnsQuery(fqdn string, rtype uint16, nameserver string, recursive bool) (in 
 	m := new(dns.Msg)
 	m.SetQuestion(fqdn, rtype)
 	m.SetEdns0(4096, false)
+
 	if !recursive {
 		m.RecursionDesired = false
 	}
@@ -137,45 +153,33 @@ func dnsQuery(fqdn string, rtype uint16, nameserver string, recursive bool) (in 
 	return
 }
 
-// lookupNameservers returns the authoritative nameservers for the given domain name.
+// lookupNameservers returns the authoritative nameservers for the given fqdn.
 func lookupNameservers(fqdn string) ([]string, error) {
-	var err error
-	var r *dns.Msg
 	var authoritativeNss []string
-	resolver := "google-public-dns-a.google.com"
 
-	r, err = dnsQuery(fqdn, dns.TypeSOA, resolver, true)
+	r, err := dnsQuery(fqdn, dns.TypeNS, recursiveNameserver, true)
 	if err != nil {
 		return nil, err
 	}
 
-	// If there is a SOA RR in the Answer section then fqdn is the root domain.
 	for _, rr := range r.Answer {
-		if soa, ok := rr.(*dns.SOA); ok {
-			r, err = dnsQuery(soa.Hdr.Name, dns.TypeNS, resolver, true)
-			if err != nil {
-				return nil, err
-			}
-
-			for _, rr := range r.Answer {
-				if ns, ok := rr.(*dns.NS); ok {
-					authoritativeNss = append(authoritativeNss, strings.ToLower(ns.Ns))
-				}
-			}
-
-			return authoritativeNss, nil
+		if ns, ok := rr.(*dns.NS); ok {
+			authoritativeNss = append(authoritativeNss, strings.ToLower(ns.Ns))
 		}
+	}
+
+	if len(authoritativeNss) > 0 {
+		return authoritativeNss, nil
 	}
 
 	// Strip of the left most label to get the parent domain.
 	offset, _ := dns.NextLabel(fqdn, 0)
 	next := fqdn[offset:]
-	// Only the TLD label left. This should not happen if the domain DNS is healthy.
 	if dns.CountLabel(next) < 2 {
-		return nil, fmt.Errorf("Could not determine root domain")
+		return nil, fmt.Errorf("Could not determine authoritative nameservers")
 	}
-	
-	return lookupNameservers(fqdn[offset:])
+
+	return lookupNameservers(next)
 }
 
 // toFqdn converts the name into a fqdn appending a trailing dot.
--- a/acme/dns_challenge.go
+++ b/acme/dns_challenge.go
@@ -6,17 +6,18 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"net"
 	"strings"
 	"time"
 
 	"github.com/miekg/dns"
 )
 
-type preCheckDNSFunc func(domain, fqdn string) bool
+type preCheckDNSFunc func(domain, fqdn, value string) error
 
-var preCheckDNS preCheckDNSFunc = checkDNS
+var preCheckDNS preCheckDNSFunc = checkDnsPropagation
 
-var preCheckDNSFallbackCount = 5
+var recursionMaxDepth = 10
 
 // DNS01Record returns a DNS record which will fulfill the `dns-01` challenge
 func DNS01Record(domain, keyAuth string) (fqdn string, value string, ttl int) {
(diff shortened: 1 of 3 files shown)


acme/dns_challenge.go
```diff
@@ -60,50 +61,121 @@ func (s *dnsChallenge) Solve(chlng challenge, domain string) error {
...
-func checkDNS(domain, fqdn string) bool {
-	// check if the expected DNS entry was created. If not wait for some time and try again.
-	m := new(dns.Msg)
-	m.SetQuestion(domain+".", dns.TypeSOA)
-	c := new(dns.Client)
-	in, _, err := c.Exchange(m, "google-public-dns-a.google.com:53")
+// checkDnsPropagation checks if the expected TXT record has been propagated to
+// all authoritative nameservers. If not it waits and retries for some time.
+func checkDnsPropagation(domain, fqdn, value string) error {
+	authoritativeNss, err := lookupNameservers(toFqdn(domain))
```
willglynn: > I think this should be `lookupNameservers(toFqdn(fqdn))` instead.

The discussion in [a previous thread](https://github.com/xenolf/lego/pull/96#discussion_r52112859) still applies to the current commit. I'm commenting again so it appears here. [ cc: @janeczku ]

willglynn:
I like this strategy. Unfortunately, it is insufficient to mimic the behavior of typical recursive resolvers:

    $ dig txt _acme-challenge.is-a-cname.willglynn.com @8.8.8.8

    ; <<>> DiG 9.8.3-P1 <<>> txt _acme-challenge.is-a-cname.willglynn.com @8.8.8.8
    ;; global options: +cmd
    ;; Got answer:
    ;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 42356
    ;; flags: qr rd ra; QUERY: 1, ANSWER: 2, AUTHORITY: 0, ADDITIONAL: 0

    ;; QUESTION SECTION:
    ;_acme-challenge.is-a-cname.willglynn.com. IN TXT

    ;; ANSWER SECTION:
    _acme-challenge.is-a-cname.willglynn.com. 59 IN	CNAME come-and-find-me.lerfjhax.com.
    come-and-find-me.lerfjhax.com. 59 IN	TXT	"ENTER MY HALL OF MIRRORS AND TREMBLE, MORTAL"

    ;; Query time: 114 msec
    ;; SERVER: 8.8.8.8#53(8.8.8.8)
    ;; WHEN: Tue Feb  9 08:44:37 2016
    ;; MSG SIZE  rcvd: 155

This particular example is rather contrived, but still, this approach isn't 100% equivalent to how `boulder` will be performing lookups.

acme/dns_challenge.go
janeczku: Validation domain (fqdn) and domain are owned by the same DNS authority, hence they share the same set of nameservers. The reason we are using the domain to lookup the authoritative NS is that we expect the domain to exist, while the fqdn may not yet have propagated in the DNS system.

willglynn: It is legal to delegate the validation domain to a different set of nameservers, in which case `lookupNameservers(toFqdn(domain))` and `lookupNameservers(toFqdn(fqdn))` will have different results.

It is also legal to complete a DNS-01 challenge by publishing a TXT record for `fqdn` even if the public authoritative nameservers have no records at all for `domain` (e.g. split horizon DNS), as I mentioned in [a second thread](https://github.com/xenolf/lego/pull/96#discussion_r52114015) on the previous PR.

janeczku:
> This particular example is rather contrived, but still, this approach isn't 100% equivalent to how boulder will be performing lookups.

The example is not only contrived. You're breaking the ACME spec when you create a CNAME record instead of a TXT record at the validation domain.

> The DNS challenge requires the client to provision a TXT record containing a designated value under a specific validation domain name.

https://ietf-wg-acme.github.io/acme/#dns

willglynn:
The spec describes validation as:

> 1. Compute the SHA-256 digest of the key authorization
> 2. Query for TXT records under the validation domain name
> 3. Verify that the contents of one of the TXT records matches the digest value

I read that as "a query for `_acme-challenge.… IN TXT` and see if one of the TXT records matches the digest". Recursive resolvers follow CNAMEs, and return those TXT records in those CNAMEs in the answer section to the query.

`boulder` follows this reading of the spec:

    $ lego --server=https://acme-staging.api.letsencrypt.org/directory -d is-a-cname.willglynn.com -m … --dns=manual -x http-01 -x tls-sni-01 run
    2016/02/09 10:57:13 [INFO][is-a-cname.willglynn.com] acme: Obtaining bundled SAN certificate
    2016/02/09 10:57:13 [INFO][is-a-cname.willglynn.com] acme: Could not find solver for: http-01
    2016/02/09 10:57:13 [INFO][is-a-cname.willglynn.com] acme: Trying to solve DNS-01
    2016/02/09 10:57:13 [INFO] acme: Please create the following TXT record in your DNS zone:
    2016/02/09 10:57:13 [INFO] acme: _acme-challenge.is-a-cname.willglynn.com. 120 IN TXT "ptnyX8Lgo8Qq4AdE-IrEy38jk0xyWT8nJXrfbt7epLM"
    2016/02/09 10:57:13 [INFO] acme: Press 'Enter' when you are done

    <switch terminals>
    $ dig txt _acme-challenge.is-a-cname.willglynn.com @8.8.8.8

    ; <<>> DiG 9.8.3-P1 <<>> txt _acme-challenge.is-a-cname.willglynn.com @8.8.8.8
    ;; global options: +cmd
    ;; Got answer:
    ;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 20282
    ;; flags: qr rd ra; QUERY: 1, ANSWER: 3, AUTHORITY: 0, ADDITIONAL: 0

    ;; QUESTION SECTION:
    ;_acme-challenge.is-a-cname.willglynn.com. IN TXT

    ;; ANSWER SECTION:
    _acme-challenge.is-a-cname.willglynn.com. 59 IN	CNAME come-and-find-me.lerfjhax.com.
    come-and-find-me.lerfjhax.com. 59 IN	TXT	"ENTER MY HALL OF MIRRORS AND TREMBLE, MORTAL"
    come-and-find-me.lerfjhax.com. 59 IN	TXT	"ptnyX8Lgo8Qq4AdE-IrEy38jk0xyWT8nJXrfbt7epLM"

    ;; Query time: 97 msec
    ;; SERVER: 8.8.8.8#53(8.8.8.8)
    ;; WHEN: Tue Feb  9 10:57:39 2016
    ;; MSG SIZE  rcvd: 211

    <switch back, press enter>
    2016/02/09 10:57:44 [INFO][is-a-cname.willglynn.com] The server validated our request
    2016/02/09 10:57:44 [INFO] acme: You can now remove this TXT record from your DNS zone:
    2016/02/09 10:57:44 [INFO] acme: _acme-challenge.is-a-cname.willglynn.com. 120 IN TXT "..."
    2016/02/09 10:57:44 [INFO][is-a-cname.willglynn.com] acme: Validations succeeded; requesting certificates
    2016/02/09 10:57:45 [INFO] acme: Requesting issuer cert from https://acme-staging.api.letsencrypt.org/acme/issuer-cert
    2016/02/09 10:57:45 [INFO][is-a-cname.willglynn.com] Server responded with a certificate.

If the spec intends to require the TXT record to be provisioned directly on the validation name, we should clarify the language in the spec and fix `boulder` to also require that.

acme/dns_challenge.go
janeczku: I disagree. The ACME spec could not be more clearer:

> The client constructs the validation domain name by prepending the label
"_acme-challenge" to the domain name being validated, then provisions a TXT
record with the digest value under that name."

Where does it tell you to delegate the validation domain to another zone?

Also, doing so would undermine the whole objective of the DNS challenge, which is to establish proof of ownership of the certificate domain by the client.

> "The client must demonstrate to the server both (1) that it holds the private key of the account key pair, and (2) that it has authority over the identifier being claimed."

If the validation domain is by itself a root domain (delegated to a different zone), then by provisioning a TXT record under it, you do not prove authority over the (parent) certificate domain, which may be owned by a third party.

janeczku: > It is also legal to complete a DNS-01 challenge by publishing a TXT record for fqdn even if the public authoritative nameservers have no records at all for domain (e.g. split horizon DNS), as I mentioned in a second thread on the previous PR. It's not correct to require domain to exist.

We do not require the certificate domain to exist. E.g, if the certificate domain www.example.com has no records (NXDOMAIN) the DNS check will still pass.

willglynn: > The ACME spec could not be more clearer:
>> The client constructs the validation domain name by prepending the label "_acme-challenge" to the domain name being validated, then provisions a TXT record with the digest value under that name."

Exactly. You don't provision records on the domain name being validated, you provision records for *the validation domain name*. The [DNS-01 section](https://ietf-wg-acme.github.io/acme/#rfc.section.7.5) doesn't concern the domain name at all – only the validation domain name, `fqdn` in this context. DNS-01 doesn't make any requirements of the domain name, only of the *validation* domain name.

> Where does it tell you to delegate the validation domain to another zone?

Where does it tell you I can't delegate it?

> Also, doing so would undermine the whole objective of the DNS challenge, which is to establish proof of ownership of the certificate domain by the client.

The DNS-01 challenge does not require asserting control over the certificate domain. If that were the goal, it could require the TXT record to be published on the certificate domain instead. (It's already specified to ignore non-matching responses, so there's no conflict even if you've already got TXT records on that name.) Why use a separate validation domain name at all?

There are legitimate use cases for delegating `_acme-challenge`  – as I mentioned in the previous thread, the situation where it is impractical to modify `bar.com` from a cron job because of grumpy change control officers (and the compliance reasons behind their grumpiness). It is desirable in such situations to delegate `_acme-challenge.foo.bar.com` out of the `bar.com` zone.

> If the validation domain is by itself a root domain (delegated to a different zone), then by provisioning a TXT record under it, you do not prove authority over the (parent) certificate domain, which may be owned by a third party. (e.g. _acme-challenge.co.uk. -> you, but co.uk. -> ns1.nic.uk)

DNS-01 does not concern itself with the certificate domain name, it concerns itself with the *validation* domain name. If you can control the validation domain name, regardless of which zone or nameservers you use to do it, that is 100% sufficient to satisfy a DNS-01 challenge. Delegation is a part of DNS, and I maintain that it is useful to delegate ACME challenges.

As for `_acme-challenge.co.uk`, please see the complex history of underscores in DNS. (Or just try registering that domain name.) It is not an accident that `_acme-challenge` contains an underscore.

> Lastly, why are we even having this academic discussion? Lego solves DNS challenge by provisioning a TXT record under the certificate domain. No zone delegation is happening, which is why we don't need to handle such case in the DNS check (irrespective of whether that would be technically legal).

It provisions a TXT record under the *validation domain*. `RFC2136_ZONE=_acme-challenge.foo.bar.com lego --dns=rfc2136` can easily create records in a delegated zone, and of course there's `--dns=manual`.

We're having this discussion because this is a situation where this check would fail while `boulder` would succeed, and I consider that a bug in this check.

janeczku:
I agree, there is something to be wished for in terms of clarity of the spec. E.g. the description as it relates to the client behavior is very specific:

> For example, if the domain name being validated is “example.com”, then the client would provision the following DNS record:
`_acme-challenge.example.com. 300 IN TXT "gfj9Xq...Rg85nM"`

I don't see how one could read that as "... the client may alternatively create a CNAME record under the validation domain and a TXT record in the target zone."

But apparently the consensus indeed seems to be that CNAME'ing the validation domain is legal: https://www.ietf.org/mail-archive/web/acme/current/msg00827.html

Anyhow, seeing that currently none of the current Lego DNS providers supports this edge case, i think there is a point to be made that we can always factor in support for this in the DNS check later when necessary.

willglynn:
> I don't see how one could read that as "... the client may alternatively create a CNAME record under the validation domain and a TXT record in the target zone."

I read the text as normative – suggestions, really – and then worked backwards from how the validation section is written.

> Anyhow, seeing that currently none of the current Lego DNS providers supports this edge case, i think there is a point to be made that we can always factor in support for this in the DNS check later when necessary.

Well, the `manual` provider would permit this today.

We could accommodate this case with a flag that says "instead of checking and proceeding automatically, wait for manual confirmation", or by making the automatic check recursive (i.e. following CNAMEs up to a limited depth in the pursuit of TXT records).

acme/dns_challenge.go
janeczku: >> Where does it tell you to delegate the validation domain to another zone?

>Where does it tell you I can't delegate it?

A TXT record can not co-exist with a NS record for the same name if the name is a subdomain and not the root domain. Hence the spec describing you to create a TXT record under the validation domain implies that you should not create a NS record there which would be necessary for your scenario of delegation.

But again, this discussion is out of scope for this PR as Lego's DNS providers don't delegate the validation domain anywhere.

willglynn: > A TXT record can not co-exist with a NS record for the same name if the name is a subdomain and not the root domain. 

It can and does.

Scenario: issuing a certificate for `is-delegated.willglynn.com`.

`willglynn.com IN NS ns1.willglynn.com`, which has `willglynn.com IN SOA ns1.willglynn.com`. Within the `willglynn.com` zone, `is-delegated.willglynn.com` does not exist, but `_acme-challenge.is-delegated.willglynn.com IN NS ns1.lerfjhax.com`.

`ns1.lerfjhax.com` has a `_acme-challenge.is-delegated.willglynn.com IN SOA ns1.lerfjhax.com`, and within *that* zone, `_acme-challenge.is-delegated.willglynn.com IN NS ns1.lerfjhax.com`, agreeing with the delegation, and there exists an `_acme-challenge.is-delegated.willglynn.com IN TXT` record as well.

Asking `lookupNameservers(toFqdn(domain))` about the validation domain returns a referral, since `ns1.willglynn.com` knows it is not an authority for the validation domain:

    $ dig txt _acme-challenge.is-delegated.willglynn.com @ns1.willglynn.com

    ; <<>> DiG 9.8.3-P1 <<>> txt _acme-challenge.is-delegated.willglynn.com @ns1.willglynn.com
    ;; global options: +cmd
    ;; Got answer:
    ;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 988
    ;; flags: qr rd; QUERY: 1, ANSWER: 0, AUTHORITY: 1, ADDITIONAL: 0
    ;; WARNING: recursion requested but not available

    ;; QUESTION SECTION:
    ;_acme-challenge.is-delegated.willglynn.com. IN TXT

    ;; AUTHORITY SECTION:
    _acme-challenge.is-delegated.willglynn.com. 300	IN NS ns1.lerfjhax.com.

    ;; Query time: 62 msec
    ;; SERVER: 205.251.198.58#53(205.251.198.58)
    ;; WHEN: Tue Feb  9 13:25:06 2016
    ;; MSG SIZE  rcvd: 87

Asking the nameserver which *is* actually an authority for the validation domain name – i.e. `lookupNameservers(toFqdn(fqdn))` – returns the TXT record:

    $ dig txt _acme-challenge.is-delegated.willglynn.com @ns1.lerfjhax.com

    ; <<>> DiG 9.8.3-P1 <<>> txt _acme-challenge.is-delegated.willglynn.com @ns1.lerfjhax.com
    ;; global options: +cmd
    ;; Got answer:
    ;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 24113
    ;; flags: qr aa rd; QUERY: 1, ANSWER: 1, AUTHORITY: 1, ADDITIONAL: 2
    ;; WARNING: recursion requested but not available

    ;; QUESTION SECTION:
    ;_acme-challenge.is-delegated.willglynn.com. IN TXT

    ;; ANSWER SECTION:
    _acme-challenge.is-delegated.willglynn.com. 60 IN TXT "1Mvyj8Fkgch0YMz0PyTk83hrPIHrYgR-pbvN3KuevTw"

    ;; AUTHORITY SECTION:
    _acme-challenge.is-delegated.willglynn.com. 259200 IN NS ns1.lerfjhax.com.

    ;; ADDITIONAL SECTION:
    ns1.lerfjhax.com.	86400	IN	A	208.79.90.50
    ns1.lerfjhax.com.	86400	IN	AAAA	2607:f2f8:a428::2

    ;; Query time: 76 msec
    ;; SERVER: 2607:f2f8:a428::2#53(2607:f2f8:a428::2)
    ;; WHEN: Tue Feb  9 13:25:08 2016
    ;; MSG SIZE  rcvd: 187

And yes, this works:

    2016/02/09 13:25:14 [INFO][is-delegated.willglynn.com] The server validated our request
    2016/02/09 13:25:14 [INFO] acme: You can now remove this TXT record from your DNS zone:
    2016/02/09 13:25:14 [INFO] acme: _acme-challenge.is-delegated.willglynn.com. 120 IN TXT "..."
    2016/02/09 13:25:14 [INFO][is-delegated.willglynn.com] acme: Validations succeeded; requesting certificates
    2016/02/09 13:25:14 [INFO] acme: Requesting issuer cert from https://acme-staging.api.letsencrypt.org/acme/issuer-cert
    2016/02/09 13:25:14 [INFO][is-delegated.willglynn.com] Server responded with a certificate.

> But again, this discussion is out of scope for this PR as Lego's DNS providers don't delegate the validation domain anywhere.

…unless you delegate the validation domain ahead of time only need `lego` to create the TXT record in the delegated zone. As I said, `RFC2136_ZONE=_acme-challenge.foo.bar.com lego --dns=rfc2136 ` can do that right now.

janeczku: > …unless you delegate the validation domain ahead of time and only need lego to create the TXT record in the delegated zone. As I said, RFC2136_ZONE=_acme-challenge.foo.bar.com lego --dns=rfc2136 can do that right now.

You are right, thats a valid (if edgy) scenario.

janeczku:
Will,

I have pushed a commit to the PR that adds support for the scenarios discussed yesterday:
* Validation domain is a CNAME
* Validation domain is delegated to another nameserver.

I tested `checkDnsPropagation(fqdn, value)` with your testing records:

`_acme-challenge.is-delegated.willglynn.com` `PASS`
`_acme-challenge.is-delegated.willglynn.com` `PASS`

Please let me know if you are :+1:  with the PR as it is so we can move forward and ask @xenolf to land this.

willglynn:
:+1:

I think this is good to merge. It works better than the current code and it addresses the two cases I had in mind as clear sources of trouble.

This check still makes some assumptions that might be problematic in conceivable situations, like that the set of nameservers doesn't change throughout the operation, but as far as I can tell so far, such issues will shake themselves out just by retrying the command. That's a big improvement over a check that will never succeed and which requires recompiling `lego` to bypass.

I've been hacking on a way to test code like this (DNS fixtures modeling a constellation of authoritative servers which you can update on command), but it's not ready yet, and in any case seems like the subject of a separate PR.

janeczku:
:yum: :tada: Yipeeeeh

> I've been hacking on a way to test code like this (DNS fixtures modeling a constellation of authoritative servers which you can update on command), but it's not ready yet, and in any case seems like the subject of a separate PR.

Looking forward to check that out once it's ready.

PR #118, opened 2016-02-10
Added a --webroot option for HTTP challenge

adriencarbonne:
When using this option, the challenge will be written in a file in
".well-known/acme-challenge/" inside the given webroot folder.

This allows lego to work without binding any port at all.


xenolf:
Hello and thank you for your contribution :)

I see that you modeled your provider after the HTTP-01 provider. HTTP-01 and TLS-SNI-01 solvers are a "special" case as they are the default and are handled a bit differently.

Other providers should be used with the `SetChallengeProvider` function and should supply a `New` function. In your case this would be `NewHTTPProviderWebroot` taking the path as an argument.

```go
provider := NewHTTPProviderWebroot("/var/www/mysite/.well-known/acme-challenge")
client.SetChallengeProvider(acme.HTTP01, provider)
```

adriencarbonne:
I hope this will be better now, if not could you please give me more hints? Thank you!

(I am also working on adding 2 more actions to lego cli: --register to do only the account registration part, no TOS, no cert ; and --tos --url to display the TOS url for use in another program ; and --tos --agree to agree the TOS => would such a PR interest you, or do I keep it on my fork?)

xenolf:
This is looking good now :+1: 

We are currently discussing how we want to handle external providers going forward in #100 and #111, this will delay merging this a bit.

Concerning the switches, I can see register being useful to just generate an account. About the TOS display... couldn't the external program just parse the JSON we save and use the URL from there?

adriencarbonne:
Thank you!

About the switches, see #125

README.md
```diff
@@ -42,10 +42,11 @@ When using the standard `--path` option, all certificates and account configurat
 
 #### Sudo
 The CLI does not require root permissions but needs to bind to port 80 and 443 for certain challenges. 
-To run the CLI without sudo, you have two options:
+To run the CLI without sudo, you have three options:
 
 - Use setcap 'cap_net_bind_service=+ep' /path/to/program
 - Pass the `--http` or/and the `--tls` option and specify a custom port to bind to. In this case you have to forward port 80/443 to these custom ports (see [Port Usage](#port-usage)).
+- Pass the `--webport` option and specify the path to your webroot folder. In this case the challenge will be written in a file in `.well-known/acme-challenge/` inside your webroot.
```
moritzheiber: I reckon `--webport` is a typo here?

adriencarbonne:
This was indeed a typo.

@xenolf : I just changed one word in readme.md, but it seems the build is now failing due to an external dependency. Most probably this is something that has already been fixed in your master I guess.

xenolf:
@adriencarbonne I think this is because of the ECC changes which were merged since you submitted this PR. Try to rebase your change on top of current master.

xenolf:
To expand further - now that #144 is merged, this provider should relocate to `lego/providers/http/webroot` :)

adriencarbonne:
Please merge now ;) Thank you!

xenolf:
While going over your code I was wondering if `--webroot` should also exclude TLS-SNI-01 just as `--dns` excludes HTTP-01 and TLS-SNI-01 because you specifically request the challenge. Any thoughts on this?

adriencarbonne:
Good idea indeed :)

PR #122, opened 2016-02-12
Support existing CSRs

willglynn:
Some systems (embedded ones mostly) can produce CSRs and accept signed certificate chains but cannot exchange private keys. This changeset allows `lego` to support such things. The workflow looks something like:

1. User asks device to generate a private key. This key is stored on the device and inaccessible to the user.
2. User asks device to generate a certificate signing request for the desired name. The device generates and signs a CSR, presenting it to the user.
3. User runs `lego --csr=/path/to/csr.pem` instead of `lego --domain=foo.bar`. The user probably also includes `--dns=provider`, since it's likely that DNS will be the easiest validation method.
4. `lego` performs an ACME exchange on behalf of the CSR and the domain(s) listed in the CSR. (Note that this does not require the CSR's private key.) `lego` writes the resulting certificate to disk.
5. User uploads the certificate to the device.

Works for me:

    $ AWS_PROFILE=… AWS_REGION=us-east-1 lego -m … --dns=route53 --csr=Downloads/myreq.csr run
    2016/02/11 18:58:22 [INFO][zd.horizon.willglynn.com] acme: Obtaining bundled SAN certificate given a CSR
    2016/02/11 18:58:23 [INFO][zd.horizon.willglynn.com] acme: Trying to solve DNS-01
    2016/02/11 18:58:55 [INFO][zd.horizon.willglynn.com] Checking DNS record propagation...
    2016/02/11 18:58:59 [INFO][zd.horizon.willglynn.com] The server validated our request
    2016/02/11 18:59:27 [INFO][zd.horizon.willglynn.com] acme: Validations succeeded; requesting certificates
    2016/02/11 18:59:27 [INFO] acme: Requesting issuer cert from https://acme-v01.api.letsencrypt.org/acme/issuer-cert
    2016/02/11 18:59:27 [INFO][zd.horizon.willglynn.com] Server responded with a certificate.

The [resulting certificate](https://crt.sh/?id=12831499) refers to the same name and key as [the CSR](https://gist.github.com/willglynn/6cdbd6bb6c5873b6f3c2), as it must, and Let's Encrypt stripped out the extraneous identifiers in the CSR, [as they are permitted to do](https://tools.ietf.org/html/draft-ietf-acme-acme-01#page-32):

     The values provided in the CSR are only a request, and are not
     guaranteed.  The server or CA may alter any fields in the certificate
     before issuance.  For example, the CA may remove identifiers that are
     not authorized for the account key that signed the request.


willglynn:
I rebased to `master`, resolving a merge conflict in the README.

xenolf:
Using a CSR should also be supported for renewal, right? In the current code, if you do not have the private key, lego will generate a new key for you and renew domains based on the existing domains it reads from the certificate.

willglynn:
Hmm, good point. If we couldn't renew directly (i.e. polling a new certificate), renewals should be attempted by issuing a new certificate from the original CSR.

We can:

1. Make `RenewCertificate()` fail if a) the certificate could not be directly renewed and b) no private key is available, forcing the caller to `ObtainCertificateForCSR()` again,
2. Make a `RenewCertificateForCSR()` which takes a CSR and force people to invoke it explicitly,
3. Extend `CertificateResource` to include a CSR alongside the private key and make `RenewCertificate()` aware of both CSRs and private keys, or
4. Extend `CertificateResource` to include a CSR alongside the private key, update `ObtainCertificate()` to store both the private key and the CSR it generated, and change `RenewCertificate()` to favor using CSRs

Option 1 is least intrusive, but option 4 seems like it might be ideal, especially since it means we don't need to touch the private key for any renewal operation.

What is your preference?

xenolf:
> If we couldn't renew directly (i.e. polling a new certificate)

This renewal technique is defined in the ACME spec (and implemented as per the spec) but does never come to play in practice because boulder chose to not support it. It may even get removed pending a current PR to the spec. This also means that option one is not possible.

This leaves us with three possible scenarios we need to keep in mind:
- The user supplies us with a private key and we use that to renew the cert
- The user does not supply us with a private key. Thus we generate a new one and "renew" the cert with the new key.
- The user supplies us with a CSR and we renew from that.

Maybe the best option is to just handle this in the CLI and to not use the `Renew`function at all. If we get a CSR passed to renew, we directly pass it into `ObtainCertificateForCSR`.

xenolf:
Do you have an update on this? :)

willglynn:
I plan to take a look again later today.

mholt:
@willglynn I now have an interest in this too. :+1: I can cheer you on or something, just let me know. :checkered_flag:

vancluever:
Guys, I've updated this to support option 4, in PR #224.

Will be giving this a spin shortly and will update the thread there when I know it works (in API mainly, if anyone wants to try it on CLI that would be huge).

xenolf:
Closing via #224

PR #123, opened 2016-02-12
--dns=foo means we specifically intend to fulfill a DNS challenge

willglynn:
This irks me, particularly because challenge selection is non-deterministic:

    $ AWS_PROFILE=… AWS_REGION=us-east-1 lego -m … --dns=route53 --server=https://acme-stang.api.letsencrypt.org/directory -d foo.willglynn.com run
    2016/02/11 19:48:49 [INFO][foo.willglynn.com] acme: Obtaining bundled SAN certificate
    2016/02/11 19:48:49 [INFO][foo.willglynn.com] acme: Trying to solve HTTP-01
    2016/02/11 19:48:49 [foo.willglynn.com] Could not obtain certificates
    	[foo.willglynn.com] error presenting token: Could not start HTTP server for challenge -> listen tcp :80: bind: permission denied

If I say `lego --dns=foo`, I want `lego` to solve a DNS challenge. That's why I told it which provider I want it to use when solving a DNS challenge.

The commit below makes `--dns=foo` also imply `-x http-01 -x tls-sni-01`.

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -112,7 +112,7 @@ func main() {
 		},
 		cli.StringFlag{
 			Name: "dns",
-			Usage: "Enable the DNS challenge for solving using a provider." +
+			Usage: "Solve a DNS challenge using the specified provider." +
 				"\n\tCredentials for providers have to be passed through environment variables." +
 				"\n\tFor a more detailed explanation of the parameters, please see the online docs." +
 				"\n\tValid providers:" +
--- a/cli.go
+++ b/cli.go
@@ -112,7 +112,7 @@ func main() {
 		},
 		cli.StringFlag{
 			Name: "dns",
-			Usage: "Solve a DNS challenge using the specified provider." +
+			Usage: "Solve a DNS challenge using the specified provider. Disables all other challenges." +
 				"\n\tCredentials for providers have to be passed through environment variables." +
 				"\n\tFor a more detailed explanation of the parameters, please see the online docs." +
 				"\n\tValid providers:" +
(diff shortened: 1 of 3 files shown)


mholt:
Yeah, this is probably better. People wouldn't go to the trouble of specifying DNS credentials if they could just use http or tls-sni challenges.

xenolf:
Could you add this to the documentation as part of your pull request? I agree that using DNS-01 probably means you want only DNS-01 but I want to make sure people know of this implicit exclude.

willglynn:
I rebased to `master`, re-worded the message in `--help` about `--dns`, and added a line to the README.

willglynn:
…oh. And now I changed another line in the README to drop the now-unnecessary `--exclude` directives.

cli.go
```diff
@@ -112,7 +112,7 @@ func main() {
 		},
 		cli.StringFlag{
 			Name: "dns",
-			Usage: "Enable the DNS challenge for solving using a provider." +
+			Usage: "Solve a DNS challenge using the specified provider." +
```
xenolf: Could you add the "Note ... " you added to the README to here as well?

willglynn: I went for brevity: "Disables all other solvers". It's a really long line already. 3bceed427a63ab1789d7aa87e0d2360bc14da8c6

xenolf: Yeah that's fine as well. I just wanted to make sure that it clearly makes this point everywhere.

PR #124, opened 2016-02-12
Reflect result of renew action in cli exit code.

fenderle:
Since after renewing a certificate most probably some action has to occur like restarting webservers and so on it is very useful to reflect a successful renewal with a zero and a not yet required renewal with a non-zero exit code.

This way you can do something like this in a cronjob or shell script:

lego ... renew --days="30" && echo "now restart the daemons" 

(Note: I did not use logger.Fatal on purpose to make clear that this is not actually an error)

### DIFF
--- a/cli_handlers.go
+++ b/cli_handlers.go
@@ -244,7 +244,7 @@ func renew(c *cli.Context) {
 
 		if int(expTime.Sub(time.Now()).Hours()/24.0) > c.Int("days") {
 			logger().Printf("Certification expiration not reached for domain %s", domain);
-			os.Exit(1)
+			os.Exit(2)
 		}
 	}
 
--- a/cli_handlers.go
+++ b/cli_handlers.go
@@ -243,7 +243,8 @@ func renew(c *cli.Context) {
 		}
 
 		if int(expTime.Sub(time.Now()).Hours()/24.0) > c.Int("days") {
-			return
+			logger().Printf("Certification expiration not reached for domain %s", domain);
+			os.Exit(1)
 		}
 	}
 


fenderle:
Maybe we should replace exit(1) with exit(2) to be able to check for an error (which would be exit code 1) or a not yet expired message (which would be 2)

xenolf:
A non-zero exit code usually means the program encountered an error along the way. In this example the binary does exactly what you asked it to do though. I see where you're coming from but maybe the exit status is not the right way to propagate that information. Others might depend on the non-zero return code only being there on error.

mholt:
Instead of relying on exit code to report whether a certificate was renewed, why not just reload the server config if the exit code is 0?

fenderle:
Because the exit code is always zero, even if the cert has not been renewed.

The patch returns non-zero exit code in case the cert has not been renewed, which could be considered as an error for the renew operation with --days specified. To differentiate the result codes it is quite usual to set different exit codes.

mholt:
The --days flag, however, says to lego, "Only renew if expiration is within 30 days." If the certificate is not expiring soon, I don't think that's a program error. It's just a no-op.

If the question is, "How do I know when to reload my server config?" you could just always reload it after running lego --renew, which will always be sure to load the new certificate if there is one.

fenderle:
Reloading is not possible with all types of daemons - some only reread the certs on restart, thus interrupting operation. So one might want to limit impact.

mholt:
Interesting, I was under the impression that servers can reload with 0 downtime.

Well anyway, maybe the docs for --days could be changed to state that if it's not time to renew, that's an error. Then an error exit code makes sense. It can be ignored if the parent doesn't use it.

fenderle:
That would be great. You should then use a different exit code (maybe 2) to be able to distinguish this from other errors. My patch uses 1 as exit code which might be a dumb idea.

mischapeters:
I created a simple shell script around lego for OpenBSD. Which should be easily changed for another system. See the line with rcctl.

```shell
#!/bin/sh

domains="example.tld -d www.example.tld"
days=5
email=<email>

location=<path to lego folder>
certificates=${location}/certificates/
lego=<path to lego bin>

md5 -h /tmp/md5hashes ${certificates}*.crt
${lego} -d ${domains} --email ${email} --http :8080 --tls :8443 --path ${location} renew --days ${days} > /dev/null 2>&1

if [ $? = "1" ]
then
        echo "Something went wrong with lego."
        exit 1;
fi

cd ${certificates} && set -- *.crt
for _c; do
        md5 -C /tmp/md5hashes ${certificates}${_c} > /dev/null 2>&1
done

if [ $? = "1" ]
then
        echo "New certificates installed."
        rcctl restart httpd
else
        echo "No new certificates installed."
fi

rm -rf /tmp/md5hashes
```

The reason for the http and tls port is I am doing some magic with pf so it's easier to get Let's Encrypt to talk to the process. Source: https://blog.netsight.org/archives/2015/12/05/openbsd__lets_encrypt_the_webserver/index.html

```
table <letsencrypt> persist { 66.133.109.36, 64.78.149.164, 2.20.25.220 }

pass in quick inet proto tcp from <letsencrypt> to port http rdr-to egress port 8080
pass in quick inet proto tcp from <letsencrypt> to port https rdr-to egress port 8443
```

PR #125, opened 2016-02-12
Added "register" and "agree" actions

adriencarbonne:
`lego register` only does the registration part of the `lego run`
process.
`lego agree` only does the TOS agreement part of the `lego run` process.

This is great for using with a 3rd party tool/GUI while still using the
standard lego CLI binary.

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -72,6 +72,16 @@ func main() {
 				},
 			},
 		},
+		{
+			Name:   "register",
+			Usage:  "Register an account",
+			Action: register,
+		},
+		{
+			Name:   "agree",
+			Usage:  "Agree to CA Terms Of Service",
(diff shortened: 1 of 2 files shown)


willglynn:
It makes sense to me that `lego` should have a non-interactive way of completing account setup.

The official `letsencrypt` Python CLI has `--email` and `--agree-tos` switches, enabling one to register for an account if needed, agree to the TOS if needed, and obtain a certificate in a single command invocation. Would pattern this be preferable to having separate verbs? Does it matter to anyone?

xenolf:
I have added a global `--agree-tos` flag to the CLI.

d-Rickyy-b:
The CLI flag is now called `--accept-tos`.

PR #127, opened 2016-02-14
Tweaks

mholt:
This does 3 things, mostly inconsequential:

- Limit HTTP requests to 10 seconds
- Fixes vet errors and lint warnings
- Adds `go vet` to CI as a way to be reminded to keep the code clean and correct

I also have a small tweak to the GetOCSPForCert function ready to go pending your response on gitter.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,14 @@
 language: go
 
 go:
-    - 1.4.3
-    - 1.5.3
-    - tip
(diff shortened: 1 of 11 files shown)


mholt:
Bah humbug. The tests pass locally... not really sure what's going on there. And it's the same travis manifest I use for Caddy which works fine...

PR #132, opened 2016-02-16
Refactor CloudFlare DNS provider to have no 3rd party dependencies

janeczku:
Since there is no officially maintained (or widely accepted) library for CloudFlare it is better to implement the API calls using just the standard net/http library.

### DIFF
--- a/acme/dns_challenge_cloudflare.go
+++ b/acme/dns_challenge_cloudflare.go
@@ -163,9 +163,7 @@ func (c *DNSProviderCloudFlare) makeRequest(method, uri string, body io.Reader) 
 	req.Header.Set("X-Auth-Key", c.authKey)
 	req.Header.Set("User-Agent", userAgent())
 
-	client := http.DefaultClient
-	client.Timeout = time.Duration(30 * time.Second)
-
+	client := http.Client{Timeout: 30 * time.Second}
 	resp, err := client.Do(req)
 	if err != nil {
 		return nil, fmt.Errorf("Error querying API -> %v", err)
--- a/acme/dns_challenge_cloudflare.go
+++ b/acme/dns_challenge_cloudflare.go
@@ -161,6 +161,7 @@ func (c *DNSProviderCloudFlare) makeRequest(method, uri string, body io.Reader) 
 
 	req.Header.Set("X-Auth-Email", c.authEmail)
 	req.Header.Set("X-Auth-Key", c.authKey)
+	req.Header.Set("User-Agent", userAgent())
 
 	client := http.DefaultClient
 	client.Timeout = time.Duration(30 * time.Second)
--- a/acme/dns_challenge_cloudflare.go
+++ b/acme/dns_challenge_cloudflare.go
@@ -1,23 +1,26 @@
 package acme
 
 import (
+	"bytes"
+	"encoding/json"
 	"fmt"
+	"io"
+	"net/http"
 	"os"
 	"strings"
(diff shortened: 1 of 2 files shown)


acme/dns_challenge_cloudflare.go
```diff
@@ -54,85 +66,126 @@ func (c *DNSProviderCloudFlare) Present(domain, token, keyAuth string) error {
...
+		Result  json.RawMessage `json:"result"`
+	}
+
+	req, err := http.NewRequest(method, fmt.Sprintf("%s%s", CloudFlareApiURL, uri), body)
+	if err != nil {
+		return nil, err
+	}
+
+	req.Header.Set("X-Auth-Email", c.authEmail)
+	req.Header.Set("X-Auth-Key", c.authKey)
```
mholt: Maybe we should set the User-Agent here like we do for all requests in http.go.

mholt:
Huge +1 to simplifying!

acme/dns_challenge_cloudflare.go
janeczku: Sure. What string do you suggest? :yum:

acme/dns_challenge_cloudflare.go
```diff
@@ -54,85 +66,127 @@ func (c *DNSProviderCloudFlare) Present(domain, token, keyAuth string) error {
...
+	req.Header.Set("User-Agent", userAgent())
+
+	client := http.DefaultClient
+	client.Timeout = time.Duration(30 * time.Second)
```
mholt: Yikes, let's not change the DefaultClient. Smells like a race condition.

janeczku: `DefaultClient` is actually just a shortcut for `&http.Client{}`. So it's not dangerous but probably a bit odd to customize the default client. Changed.

mholt: It is dangerous because that variable is shared by all code that uses
net/http Post and Get functions and anyone else using http.DefaultClient,
but the value is not protected by a mutex.

acme/dns_challenge_cloudflare.go
```diff
@@ -26,26 +29,35 @@ func NewDNSProviderCloudFlare(cloudflareEmail, cloudflareKey string) (*DNSProvid
...
 		return err
 	}
 
-	record := newTxtRecord(zoneID, fqdn, value, ttl)
-	err = c.client.Records.Create(c.ctx, record)
+	rec := cloudFlareRecord{
+		Type:    "TXT",
+		Name:    unFqdn(fqdn),
+		Content: value,
+		TTL:     120,
```
xenolf: Why are we using a static TTL of 120 here instead of the supplied one above?

janeczku: That's the minimum valid TTL CloudFlare allows. So this was meant as a precaution in case `DNS01Record` will be changed to return a lower value anytime in the future. If you want i can bring back the clamp function from the current CloudFlare provider: https://github.com/xenolf/lego/blob/master/acme/dns_challenge_cloudflare.go#L127-L136

xenolf: Ah, alright. I just wanted to know the reason. In this case it's fine I think.

acme/dns_challenge_cloudflare.go
janeczku: Right, good catch!

PR #135, opened 2016-02-19
DNS Challenge: Do not join TXT record values

xi2:
ACME spec states that the contents of **one** of the TXT records should
match the digest value, not their concatenation.

I am looking at <https://ietf-wg-acme.github.io/acme/> section 7.5. We appear to be doing it wrongly at present? Am I looking at the right spec?


xi2:
I see now, upon closer reading of the spec for TXT records that my pull request actually in error, and that the existing join is not for multiple records, but for multiple strings within the same record. My apologies. However there is still a tiny bug in the error reporting, for which I will submit a new PR. Closing.

PR #136, opened 2016-02-19
DNS Challenge: Fix handling of CNAMEs

xi2:
In #133 I was seeing unnecessary delays in checking/confirming the DNS propagation for my Gandi challenge provider. 

When I examined `checkDNSPropagation` I found that it was exiting
early if the TXT record could not be found on the recursive
nameserver, and thus the authoritative nameservers were not being
queried until after the record showed up on the recursive nameserver
causing the delay.

If I am understanding things correctly (a big if!) if the recursive nameserver returns not found for the CNAME check we should just press on and examine the authoritative nameservers. This commit does that and it *appears* to fix my problem.

Is this correct? Thoughts?

### DIFF
--- a/acme/dns_challenge.go
+++ b/acme/dns_challenge.go
@@ -82,16 +82,14 @@ func checkDNSPropagation(fqdn, value string) (bool, error) {
 	if err != nil {
 		return false, err
 	}
-	if r.Rcode != dns.RcodeSuccess {
-		return false, fmt.Errorf("Could not resolve %s -> %s", fqdn, dns.RcodeToString[r.Rcode])
-	}
-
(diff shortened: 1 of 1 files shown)


janeczku:
Well the reasoning for retrying on a NXDOMAIN was that the CNAME might just have been created in the process. But you are definitely right that this leads to delays as Google DNS caches NXDOMAIN results quite aggressively.
I think we can expect users who require these kind of edge setups (CNAME, delegation) to put that in place beforehand.
So +1 for merging this.

xi2:
Thanks for the explanation @janeczku. Much appreciated.

PR #138, opened 2016-02-19
CLI: Give helpful error message if --http/--tls is given without colon

### LINKED ISSUE #134: --tls flag not working
Hi,

I tried to generate a certificate using the --tls flag to change the https challenge server port but it appear to have no effect and try to bind at port 443.

```
go get -u github.com/xenolf/lego
cd go/bin
./lego --path /etc/lego --domains domain.com -m my@email.com  --http 8080 --tls 8081 run
```

Gives me the error:
```
error presenting token: Could not start HTTPS server for challenge -> listen tcp :443: bind: address already in use
```

Am I doing something wrong or is it a bug?

Thanks!

xi2:
Fixes #134

### DIFF
--- a/cli_handlers.go
+++ b/cli_handlers.go
@@ -44,10 +44,16 @@ func setup(c *cli.Context) (*Configuration, *Account, *acme.Client) {
 	}
 
 	if c.GlobalIsSet("http") {
+		if strings.Index(c.GlobalString("http"), ":") == -1 {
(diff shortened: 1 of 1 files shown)


xenolf:
Could you re-write the error strings to be more in-line with the other strings?
Something like `The --http switch only accepts interface:port or :port for its argument.`

xi2:
Done :)

PR #139, opened 2016-02-26
Some small textual fixes and adding a flag to supress bundle creation.

middelink:
Used a negative flag, --nobundle, to make sure the default would not change.


xenolf:
Thank you for your contribution. 
Could you please rename the new option to be in line with the other multi word switches?

mholt:
Meaning, I think to use a hyphen: --no-bundle

middelink:
Done.

PR #141, opened 2016-02-26
Add DNS challenge provider for Google Cloud DNS

peterwald:
Use GCE_PROJECT to designate your GCE project.

Authentication is automatically picked up from gcloud credentials if
running locally and from GCE metadata if run within Google Cloud.

Requires at least permission scope
"https://www.googleapis.com/auth/ndev.clouddns.readwrite"

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -163,6 +163,7 @@ Here is an example bash command using the CloudFlare DNS provider:
 	fmt.Fprintln(w, "\tdigitalocean:\tDO_AUTH_TOKEN")
 	fmt.Fprintln(w, "\tdnsimple:\tDNSIMPLE_EMAIL, DNSIMPLE_API_KEY")
 	fmt.Fprintln(w, "\tgandi:\tGANDI_API_KEY")
+	fmt.Fprintln(w, "\tgcloud:\tGCE_PROJECT")
 	fmt.Fprintln(w, "\tmanual:\tnone")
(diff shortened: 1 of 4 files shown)


xenolf:
Thank you for this!

I see that you are providing a live test only, please make sure that the live test is not run when no credentials are supplied.

Please note that the package which holds the 3rd party DNS providers will change tomorrow so it will require another change.

peterwald:
OK, the tests are passing when not run live. Let me know if you need any further integration with the new DNS provider package.

peterwald:
@xenolf I've refactored the code to account for #144

xenolf:
@peterwald Thank you! I'm sorry that I've been a bit slow on progress with #144 but it should get merged in soon.

xenolf:
@peterwald something seems to be acting up in here. Could you please rebase this PR onto current master now that #144 is merged? Thanks!

peterwald:
This merge looks a lot worse than it is due to the resyncing. Even though it looks like a big change because all of the intermediate commits were included, the actual differences between the starting and ending state are small. If you'd prefer I can close out this PR and create another one that shows the changes more cleanly.

LukeHandle:
I'd probably suggest a rebase/squash to tidy this?

xenolf:
Could this PR get rebased on top of #148? This allows you to customize the timeout and interval of the dns pre-check and thus removes the need to export the function.

peterwald:
Rebased on #148 and squashed.

PR #142, opened 2016-02-27
Use zone name when talking to DNS APIs

LukeHandle:
I noticed that the CloudFlare lookup didn't lookup directly based on the zone name (which seemed silly).

I Adjusting lookupNameservers() to return the fqdn at point it finds the NS as this seemed like a better idea vs. making a new function to repeat the same process.

I also changed the other providers to use the new method vs. the suffix method. I've adjusted where appropriate for differing uses of FQDNs in zone names.

### DIFF
--- a/providers/dns/route53/route53.go
+++ b/providers/dns/route53/route53.go
@@ -104,16 +104,19 @@ func (r *DNSProvider) getHostedZoneID(fqdn string) (string, error) {
 		zones = append(zones, zoneResp.HostedZones...)
 	}
 
+	authZone, err := acme.FindZoneByFqdn(fqdn, acme.RecursiveNameserver)
+	if err != nil {
+		return "", err
+	}
+
 	var hostedZone route53.HostedZone
 	for _, zone := range zones {
-		if strings.HasSuffix(fqdn, zone.Name) {
-			if len(zone.Name) > len(hostedZone.Name) {
-				hostedZone = zone
-			}
+		if zone.Name == authZone {
+			hostedZone = zone
 		}
 	}
 	if hostedZone.ID == "" {
-		return "", fmt.Errorf("No Route53 hosted zone found for domain %s", fqdn)
+		return "", fmt.Errorf("Zone %s not found in Route53 for domain %s", authZone, fqdn)
 	}
 
 	return hostedZone.ID, nil
--- a/acme/dns_challenge.go
+++ b/acme/dns_challenge.go
@@ -21,7 +21,7 @@ var (
 	fqdnToZone                  = map[string]string{}
 )
 
-var recursiveNameserver = "google-public-dns-a.google.com:53"
+var RecursiveNameserver = "google-public-dns-a.google.com:53"
 
 // DNS01Record returns a DNS record which will fulfill the `dns-01` challenge
 func DNS01Record(domain, keyAuth string) (fqdn string, value string, ttl int) {
@@ -90,7 +90,7 @@ func (s *dnsChallenge) Solve(chlng challenge, domain string) error {
 // checkDNSPropagation checks if the expected TXT record has been propagated to all authoritative nameservers.
 func checkDNSPropagation(fqdn, value string) (bool, error) {
 	// Initial attempt to resolve at the recursive NS
-	r, err := dnsQuery(fqdn, dns.TypeTXT, recursiveNameserver, true)
+	r, err := dnsQuery(fqdn, dns.TypeTXT, RecursiveNameserver, true)
 	if err != nil {
 		return false, err
 	}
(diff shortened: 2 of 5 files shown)


janeczku:
On a first look i don't see how doing a series of UDP queries is an improvement both in terms of reliability and simplicity over the current code that iterates the slice of zones returned from the API . If any, you would definitely need to add some retry logic to `lookupNameservers()` as currently a single lost UDP packet will cause it to error.
Are there benefits i don't see?

LukeHandle:
I had assumed (uh-oh) that the DNS package would be handling the lost packets/reliability issues of DNS. If it does not, then we should be implementing that any time we perform a dnsQuery().

The first reason I created this was so that for the APIs that supported it, you could perform a lookup directly referencing the correct zone name, grab the ID and continue with the record changes. This seems much more logical and scales much better when you have a large number of hosted zones.

The reason I then extended it for the other APIs that did not support direct querying was because it then simplified the search for the zone ID within the respective slices. Instead of messing around with getting suffix and checking if this version was the longest out of all the ones returned, we could just find the zone that exactly matches.

NB. Route53 does actually support lookups by name, though our library does not - [ListHostedZonesByName](http://docs.aws.amazon.com/Route53/latest/APIReference/api-list-hosted-zones-by-name.html). This might be something we may wish to PR there, or implement ourselves without the library?

janeczku:
Pinned because it isn't obvious: https://github.com/miekg/dns/blob/master/client.go#L40-L41

LukeHandle:
I've just remembered, don't we use TCP anyway: https://github.com/LukeHandle/lego/blob/patch-dns-providers/acme/dns_challenge.go#L147 ?

We will still see issues with failed queries, though the rate at which this occurs when it shouldn't?

Either way, these are changes needed as part of dnsQuery().

janeczku:
`if err == dns.ErrTruncated {`

This only fires when there was an UDP response in the first place (with the truncated bit set in the header).

janeczku:
This seems much more logical
> For me it is logical to query at the source (API) rather than to rely on a third party (Google DNS server).  :stuck_out_tongue: 

> and scales much better when you have a large number of hosted zones.
I do agree.

I think another argument is, that RFC2136 requires this kind of DNS zone lookup anyway (https://github.com/xenolf/lego/pull/143) so why not implement it once and let it be used by all the DNS providers as they see fit.

LukeHandle:
> For me it is logical to query at the source (API) rather than to rely on a third party (Google DNS server).
An interesting one. The argument of whether the "source" is what is returned to us by DNS (and consequently clients) or what may be configured in the API by the owners?

Honestly, the only time that it may be misleading is if a zone has been configured in the API that is not in use (either being prepared or decommissioned). Pretty much all of the time I would expect the current method to work as expected.

If the RFC2136 is merged I will of course update this, though I expect the DNS provider package re-order will happen soon as well.

janeczku:
> Honestly, the only time that it may be misleading is if a zone has been configured in the API that is not in use (either being prepared or decommissioned).

DNS propagation delay is a real thing -  the data returned by the caching recursive nameserver is not necessarily a true representation of the current state of the DNS as configured in the API or even already provisioned to the authoritative nameservers. The scenarios where this might cause trouble (creating the record fails because the zone name is incorrect) may be limited but it is still a valid argument for looking up the enclosing zone "at the source".

xenolf:
I assume this is no longer needed as #143 was merged?

LukeHandle:
I don't believe #143 made the changes to the other challenges. I would suggest we still went ahead with that (and I'm happy to update the PR this weekend). 

Is there any opinion against such a change? Now we have a function, the PR would be similar but swaping out old for new.

xenolf:
Alright, I will wait for you to update this PR then :smile: 

The only issue I could see with this is a desync between the API and the DNS server as @janeczku already mentioned.

LukeHandle:
Hey @xenolf, are you able to take a look and advise on any changes if required?

xenolf:
@LukeHandle I'm sorry... I didn't get notified of the changes you made :worried: 
Will look at it ASAP!

providers/dns/route53/route53.go
```diff
@@ -104,16 +104,19 @@ func (r *DNSProvider) getHostedZoneID(fqdn string) (string, error) {
...
+		return "", err
+	}
+
 	var hostedZone route53.HostedZone
 	for _, zone := range zones {
-		if strings.HasSuffix(fqdn, zone.Name) {
-			if len(zone.Name) > len(hostedZone.Name) {
-				hostedZone = zone
-			}
+		if zone.Name == authZone {
```
xenolf: Why aren't we querying route53 with the hosted zone we determined before we get here? From what I gathered from a quick look at the docs it seems like we can pass a marker to the `ListHostedZones` function.

xenolf:
Apart from my comment I'm in favor of this. @janeczku any opinions? :)

LukeHandle:
@xenolf I don't believe the marker functions like that:
> For the value of marker, specify the value of the NextMarker element that was returned in the previous response.

I mentioned before, Route53 does actually support lookups by name, though our library does not - ListHostedZonesByName. Annoyingly, the returned result is in reverse order (eg. com.example.subdomain). I could not see any other ability to "search" or specifically lookup using a domain.

janeczku:
Besides the inability to lookup the zone by name It also seems that the currently used library is not maintained anymore. Hence the Route53 provider should probably be changed to use the official AWS library in the near future (the large size of the library isn't as big as a deal now that the providers are in their own packages). 

I think this can be merged.

xenolf:
> I don't believe the marker functions like that

My bad :worried: 
Yeah I guess we should switch over to the official provider.

PR #143, opened 2016-02-28
Issue/140: Using RFC2136 dns challenge does not allow me to create multi domain certs.

middelink:
- Removal of RFC2136_ZONE from help text
 - Query nameserver directly to find zone we have to update
 - During insert, make sure the new record is the ONLY challenge.
   (I had a few panics, hence 3 challenges left. Not good.)

### DIFF
--- a/acme/dns_challenge.go
+++ b/acme/dns_challenge.go
@@ -18,7 +18,7 @@ type preCheckDNSFunc func(fqdn, value string) (bool, error)
 
 var (
 	preCheckDNS preCheckDNSFunc = checkDNSPropagation
-	fqdn2zone                   = map[string]string{}
+	fqdnToZone                  = map[string]string{}
 )
 
 var recursiveNameserver = "google-public-dns-a.google.com:53"
@@ -185,7 +185,7 @@ func lookupNameservers(fqdn string) ([]string, error) {
 // findZoneByFqdn determines the zone of the given fqdn
 func findZoneByFqdn(fqdn, nameserver string) (string, error) {
 	// Do we have it cached?
-	if zone, ok := fqdn2zone[fqdn]; ok {
+	if zone, ok := fqdnToZone[fqdn]; ok {
 		return zone, nil
 	}
 
@@ -210,9 +210,9 @@ func findZoneByFqdn(fqdn, nameserver string) (string, error) {
 				// If we ended up on one of the TLDs, it means the domain did not exist.
 				publicsuffix, _ := publicsuffix.PublicSuffix(unFqdn(zone))
 				if publicsuffix == unFqdn(zone) {
-					return "", fmt.Errorf("Could not determine zone authoritively")
+					return "", fmt.Errorf("Could not determine zone authoritatively")
 				}
-				fqdn2zone[fqdn] = zone
+				fqdnToZone[fqdn] = zone
 				return zone, nil
 			}
 		}
@@ -225,9 +225,9 @@ func findZoneByFqdn(fqdn, nameserver string) (string, error) {
 			// If we ended up on one of the TLDs, it means the domain did not exist.
 			publicsuffix, _ := publicsuffix.PublicSuffix(unFqdn(zone))
 			if publicsuffix == unFqdn(zone) {
-				return "", fmt.Errorf("Could not determine zone authoritively")
+				return "", fmt.Errorf("Could not determine zone authoritatively")
 			}
-			fqdn2zone[fqdn] = zone
+			fqdnToZone[fqdn] = zone
 			return zone, nil
 		}
 	}
@@ -250,6 +250,11 @@ func unFqdn(name string) string {
 		return name[:n-1]
 	}
 	return name
+}
+
+// clearFqdnCache clears the cache of fqdn to zone mappings. Primarily used in testing.
+func clearFqdnCache() {
+	fqdnToZone = map[string]string{}
 }
 
 // waitFor polls the given function 'f', once every 'interval' seconds, up to 'timeout' seconds.
--- a/acme/dns_challenge.go
+++ b/acme/dns_challenge.go
@@ -11,13 +11,17 @@ import (
 	"time"
 
 	"github.com/miekg/dns"
+	"golang.org/x/net/publicsuffix"
 )
 
 type preCheckDNSFunc func(fqdn, value string) (bool, error)
 
-var preCheckDNS preCheckDNSFunc = checkDNSPropagation
+var (
+	preCheckDNS preCheckDNSFunc = checkDNSPropagation
+	fqdn2zone                   = map[string]string{}
+)
 
-var recursiveNameserver = "google-public-dns-a.google.com"
+var recursiveNameserver = "google-public-dns-a.google.com:53"
 
 // DNS01Record returns a DNS record which will fulfill the `dns-01` challenge
 func DNS01Record(domain, keyAuth string) (fqdn string, value string, ttl int) {
@@ -105,7 +109,7 @@ func checkDNSPropagation(fqdn, value string) (bool, error) {
 // checkAuthoritativeNss queries each of the given nameservers for the expected TXT record.
 func checkAuthoritativeNss(fqdn, value string, nameservers []string) (bool, error) {
 	for _, ns := range nameservers {
-		r, err := dnsQuery(fqdn, dns.TypeTXT, ns, false)
+		r, err := dnsQuery(fqdn, dns.TypeTXT, net.JoinHostPort(ns, "53"), false)
 		if err != nil {
 			return false, err
 		}
@@ -133,6 +137,7 @@ func checkAuthoritativeNss(fqdn, value string, nameservers []string) (bool, erro
 }
 
 // dnsQuery sends a DNS query to the given nameserver.
+// The nameserver should include a port, to facilitate testing where we talk to a mock dns server.
 func dnsQuery(fqdn string, rtype uint16, nameserver string, recursive bool) (in *dns.Msg, err error) {
 	m := new(dns.Msg)
 	m.SetQuestion(fqdn, rtype)
@@ -142,7 +147,7 @@ func dnsQuery(fqdn string, rtype uint16, nameserver string, recursive bool) (in 
 		m.RecursionDesired = false
 	}
 
-	in, err = dns.Exchange(m, net.JoinHostPort(nameserver, "53"))
+	in, err = dns.Exchange(m, nameserver)
 	if err == dns.ErrTruncated {
 		tcp := &dns.Client{Net: "tcp"}
 		in, _, err = tcp.Exchange(m, nameserver)
@@ -155,7 +160,12 @@ func dnsQuery(fqdn string, rtype uint16, nameserver string, recursive bool) (in 
 func lookupNameservers(fqdn string) ([]string, error) {
 	var authoritativeNss []string
 
-	r, err := dnsQuery(fqdn, dns.TypeNS, recursiveNameserver, true)
+	zone, err := findZoneByFqdn(fqdn, recursiveNameserver)
+	if err != nil {
+		return nil, err
+	}
+
+	r, err := dnsQuery(zone, dns.TypeNS, recursiveNameserver, true)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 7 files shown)


acme/dns_challenge_rfc2136.go
```diff
@@ -99,3 +101,55 @@ func (r *DNSProviderRFC2136) changeRecord(action, fqdn, value string, ttl int) e
...
+	// in the authority section of the reply it will have the SOA of the
+	// containing zone. rfc2308 has this to say on the subject:
+	//   Name servers authoritative for a zone MUST include the SOA record of
+	//   the zone in the authority section of the response when reporting an
+	//   NXDOMAIN or indicating that no data (NODATA) of the requested type exists
+	m := new(dns.Msg)
+	m.SetQuestion(fqdn, dns.TypeSOA)
+	m.SetEdns0(4096, false)
+	m.RecursionDesired = true
+	m.Authoritative = true
```
janeczku: `Authoritative` is an answer flag. Should not be set in a query.

acme/dns_challenge_rfc2136.go
```diff
@@ -99,3 +101,55 @@ func (r *DNSProviderRFC2136) changeRecord(action, fqdn, value string, ttl int) e
...
+	// Query the authorative nameserver for a hopefully non-existing SOA record,
+	// in the authority section of the reply it will have the SOA of the
+	// containing zone. rfc2308 has this to say on the subject:
+	//   Name servers authoritative for a zone MUST include the SOA record of
+	//   the zone in the authority section of the response when reporting an
+	//   NXDOMAIN or indicating that no data (NODATA) of the requested type exists
+	m := new(dns.Msg)
+	m.SetQuestion(fqdn, dns.TypeSOA)
+	m.SetEdns0(4096, false)
+	m.RecursionDesired = true
```
janeczku: `m.RecursionDesired = false` - We are talking to the authoritative NS.

janeczku:
Looking good. Just two nits with the DNS message headers and the unit test needs to be updated with the changes, too.

LukeHandle:
The findZone() function would probably be suited as shared between all the DNS providers. It is pretty much what I was aiming for in #142. You could go one step further and adjust lookupNameservers() to use query the NS of the result of findZone().

CloudFlare would certainly benefit from being able to directly lookup the zone. Rackspace also supports that in their API (which I was hoping to implement after the restructure).

acme/dns_challenge_rfc2136.go
middelink: Actually, I tried that, didn't work. I got a answer back listing all the root NSs.

Let me try again to see what goes wrong, because I run "dig +norecurse SOA smtp.my.domani" I do get a proper reply, so i'm missing something.

acme/dns_challenge_rfc2136.go
middelink: D'oh. Removed.

acme/dns_challenge_rfc2136.go
middelink: Right, found it. Not happy. I have a split view bind setup, an internal one with, well all the internal hosts and an external view with all the other domains I have, exposed to the outside world. Turns out that when I query the NS from inside my network, it correctly picks the internal view which knows nothing about the external domains. So without recursion I get the root NSs as hint to look at.

As adding those zones into the internal view  is a no-no due to the dynamic updates we want to do later on, I'm opting for leaving the recursion on.

middelink:
@LukeHandle Interesting, I would have thought that finding the zone would be more useful as a new function on the dns challengers. So they can each implement in a way they see fit? But glancing over the current implementations, they all iterate over the list of zones they manage, to pick out the right one. Offering the zone to them in Present() might indeed improve life for them.

acme/dns_challenge_rfc2136.go
janeczku: If i understand correctly, your DNS internal view is setup so that it does `recursive queries` against domains it itself is authoritative for? That doesn't sound right...

middelink:
I don't believe this single query SOA answer approach is working out.

Reason: invalid domains mess things up. We were trying to use the SOA of the authoritative zone in the AUTHORITY section of the reply. This works fine for valid, existing domains. For invalid domains, they return "nl." or "com." or even just "." as /that/ NS has authoritative knowledge over the non-existence. Sadly however it is NOT the NS we want to talk to... But I have no way of discerning both type of answers. Using the number of dots > 1 is no good, invalid-domain.co.uk. would slip through.
So lookupZone("invalid-domain.co.uk") > "co.uk"; lookupNss("co.uk") = ["nsa.co.uk"...] and this is where we send the dyn-updates to and query to check if the propagation has succeeded... Not getting  enthusiastic :)

Am I missing anything?

PS. The current implementation of lookupNameservers also misses the boat for .co.uk domains.

middelink:
Ahum. Turns out there is a thing called https://godoc.org/golang.org/x/net/publicsuffix. I'm using that the verify the NXDOMAIN SOA record is not on that list.

LukeHandle:
@middelink Each of them would benefit from this as at some point or another they have to establish which zone to use. With this method we can say without any doubt which zone the record should be added to. Iterating through the returned list is inefficient when you have hundreds of hosted domains and DNS should always give the actual answer.

CloudFlare and Rackspace APIs also allow you to specify the zone name in the lookup simplifying things there.

If we use the SOA method then I would agree a new function would make sense in dns_providers. The lookupNameservers() function can call the lookup zone and perform an NS query on the zone vs. iterating through each level. Supplying it with the Present would also make sense considering they would all benefit from it.

middelink:
@LukeHandle I agree with you fully :) The approach I'm taking now (having found a solution to the invalid domains; see above) is the one of least resistance and have findZone() cache, so if dns providers call it again, it does not have to go out and do the dns thing again.

The other, slightly more intrusive approach would be to get the zone in Solve() and plumb it through to all the providers. then no caching would be needed (making testing a tad easier too)

As I have no strong feelings about one over the other, feel free to +1  the later ^^

janeczku:
> For invalid domains, they return "nl." or "com." or even just "." as /that/ NS has authoritative knowledge over the non-existence. Sadly however it is NOT the NS we want to talk to... 

@middelink 
Thats a good example for why setting the `recursion desired` flag when talking to an authoritative NS is a bad idea (see my comment above). Without the recursion flag the SOA method in any case either returns the appropriate NS or nothing at all.

**Valid domain**

> dig +norecurse www.rancher.com SOA @emma.ns.cloudflare.com

```
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 39531
;; flags: qr aa; QUERY: 1, ANSWER: 0, AUTHORITY: 1, ADDITIONAL: 0
;; QUESTION SECTION:
;www.rancher.com.		IN	SOA
;; AUTHORITY SECTION:
rancher.com.		86400	IN	SOA	brad.ns.cloudflare.com. dns.cloudflare.com.
```

=> lookupZoneByFqdn() would return the correct zone name

**Invalid subdomain**

> dig +norecurse nonexistant.subdomain.rancher.com SOA @emma.ns.cloudflare.com

```
;; ->>HEADER<<- opcode: QUERY, status: NXDOMAIN, id: 26264
;; flags: qr aa; QUERY: 1, ANSWER: 0, AUTHORITY: 1, ADDITIONAL: 0
;; QUESTION SECTION:
;nonexistant.subdomain.rancher.com. IN	SOA
;; AUTHORITY SECTION:
rancher.com.		86400	IN	SOA	brad.ns.cloudflare.com. dns.cloudflare.com.
```

=> lookupZoneByFqdn() would return the correct zone name

**Invalid root domain**

> dig +norecurse nowaythatdomainexists.com SOA @emma.ns.cloudflare.com

```
;; ->>HEADER<<- opcode: QUERY, status: REFUSED, id: 3982
;; flags: qr; QUERY: 1, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 0
;; QUESTION SECTION:
;nowaythatdomainexists.com.	IN	SOA
```

=> lookupZoneByFqdn() would return an error

LukeHandle:
@middelink Personally, :+1: the latter :stuck_out_tongue: 

I think that each provider benefits from it and it reduces duplicated code as they each try and find out which zone to use.

middelink:
@janeczku True, it won't help if we do not know the authoritative nameservers. So what would only work for the rfc2136 provider (even assuming all the domains I want to request a cert for are under the same RFC2136_NAMESERVER btw). And of course it would break 100% reliable in split view nameserver setups.

In all other cases we still do the chopping off labels thing (lookupNameservers) which than still ends up misdiagnosing invalid domains.

middelink:
@LukeHandle I decided to not go for the plumbing in this PR, it is already touching too much code and changing the interface on Present/Cleanup does not seem like a hot idea right now.

janeczku:
> For invalid domains, they return "nl." or "com." or even just "." as /that/ NS has authoritative knowledge over the non-existence. Sadly however it is NOT the NS we want to talk to... 

To which nameserver *do* you want to talk regarding a non-existent domain??? The authority for nonexistandomain.nl *is* the `nl.` TLD nameserver.

> PS. The current implementation of lookupNameservers also misses the boat for .co.uk domains.

I can't reproduce any of the behavior you were describing in relation to the current implementation of lookupNameservers() (to check propagation). Were you referring to the ability of the logic used in `lookupNameservers` to lookup the zone of the domain?

janeczku:
I still think that this PR is on the right track.

Here is the RFC recommendation on how Dynamic DNS update clients should discover the zone name:
https://tools.ietf.org/html/rfc4697#section-2.8.1

```TEXT
Dynamic update agents SHOULD send SOA or NS queries to progressively
   higher-level names to find the closest enclosing zone for a given
   name to update.
```

middelink:
> To which nameserver do you want to talk regarding a non-existent domain??? The authority for nonexistandomain.nl is the nl. TLD nameserver.

Sure. badly phrased sentence. My bad. The point was we don't have a authoritative nameserver when we do Solve(). lookupNameservers() tries to find them by doing a recursive NS search on the recursive nameserver, and cutting off the labels till it gets an answer. It only stops when it has less than two labels left, so it stops at xxx.TLD. Therefore "in-valid.co.uk" will result in the "co.uk" nameservers as it will cut away the "in-valid" part and recurses since dns.LabelCount("co.uk.") >= 2.

So while I don't object to the code, though it will result in potentially multiple DNS queries, it cutoff methodology is too simple. Hence my use of publicsuffix.

> Here is the RFC recommendation on how Dynamic DNS update clients should discover the zone name: https://tools.ietf.org/html/rfc4697#section-2.8.1

Funny how the RFC is about "Observed DNS Resolution Misbehavior" :)
So what are you saying, should we not use the SOA name in the authoritative section to directly jump to the zone we look for? It seems to pass the dns_challenge_test.go fine.

janeczku:
> Funny how the RFC is about "Observed DNS Resolution Misbehavior" :)

It seems we are not the only ones who get their mind blown by the pitfalls and intricacies of dealing with DNS. :laughing: 

> So what are you saying, should we not use the SOA name in the authoritative section to directly jump to the zone we look for?

Good question. Could the RFC statement relate to the cases where the authoritative NS is not known in advance and you have to do a iterative query down from the root domain?

I think with this requirement (https://tools.ietf.org/html/rfc2308):

Name servers authoritative for a zone MUST include the SOA record of
   the zone in the authority section of the response when reporting an
   NXDOMAIN or indicating that no data (NODATA) of the requested type exists

we can expect to always receive the SOA in the authority section on the first query, right? So my answer would be that we should use the SOA record in the first answer and we don't need to chop of names.

middelink:
Well, I guess my PR stands than ^^

BTW, if this is in, I will do some work on more hermetic testing in dns_challenge_test, up and including adding in-valid.co.uk :)

acme/dns_challenge.go
```diff
@@ -169,15 +179,59 @@ func lookupNameservers(fqdn string) ([]string, error) {
...
+			return "", fmt.Errorf("NS %s returned %s for %s", nameserver, dns.RcodeToString[in.Rcode], fqdn)
+		}
+		// We have a success, so one of the answers has to be a SOA RR
+		for _, ans := range in.Answer {
+			if soa, ok := ans.(*dns.SOA); ok {
+				zone := soa.Hdr.Name
+				// If we ended up on one of the TLDs, it means the domain did not exist.
+				publicsuffix, _ := publicsuffix.PublicSuffix(unFqdn(zone))
+				if publicsuffix == unFqdn(zone) {
+					return "", fmt.Errorf("Could not determine zone authoritively")
```
xenolf: Please don't light the pitchforks in case I'm wrong, but isn't it "authoritatively" ? :dash:

acme/dns_challenge.go
```diff
@@ -11,13 +11,17 @@ import (
...
 	"github.com/miekg/dns"
+	"golang.org/x/net/publicsuffix"
 )
 
 type preCheckDNSFunc func(fqdn, value string) (bool, error)
 
-var preCheckDNS preCheckDNSFunc = checkDNSPropagation
+var (
+	preCheckDNS preCheckDNSFunc = checkDNSPropagation
+	fqdn2zone                   = map[string]string{}
```
xenolf: Please change this name to use camel case. (fqdnToZone)

acme/dns_challenge.go
middelink: No pitchforks needed: http://grammarist.com/usage/authoritative-authoritive/ :)

acme/dns_challenge.go
middelink: Done. Also added a convenience function clearFqdnCache() so tests do not need to know how to clear the cache.

Issue #106, opened 2016-02-07
Make lego more search friendly with godoc.org


xi2:
Hi. Firstly, thank you for making lego. It's working great for me. This is not a bug report, but merely a suggestion.

When I go looking for Go stuff my first try is normally to type the thing I am looking for into godoc.org (e.g. "lets encrypt" or "acme"). I actually did this twice over the last couple of months and failed to notice lego. I eventually found it a day or two ago via a blog post.

This could be easily remedied by adding single line package comments as discussed at https://golang.org/doc/effective_go.html#commentary for github.com/xenolf/lego and github.com/xenolf/lego/acme. I can't believe I'm the only one looking on godoc.org before looking elsewhere!

Best wishes and thanks again for lego.

mholt:
I can work on this.

xenolf:
So we just need to add `Let's encrypt and ACME CLI package` to the main package and `ACME library implementation` to the acme package? (for example)

xi2:
Yes, I believe so. Then if I type "let's encrypt" into Godoc search, because that string is in the package description, it should come up. And more importantly it shows what lego is used for. :)

xenolf:
Well, that shouldn't be too hard to add then :wink: 

Thanks for bringing it to our attention!

PR #144, opened 2016-02-29
Move DNS providers out of ACME package.

xenolf:
Open for comment.

I moved all existing providers into their own cozy packages to get some dependency burden off the ACME package.

The proposed structure is `providers/dns/$provider`. This lets us add providers for other challenges in a consistent manner by just adding another folder. I also moved support functions for the providers into the `providers/dns` package.

I did not remove the dependency on `miekg/dns` from `dns_challenge` because I couldn't find an elegant solution for ripping the dns check function out.


middelink:
Can we then also have a ChallengeDNSProvider for these? (can extend ChallengeProvider). I am working on a tool to write out TLSA records, and beeing able to add an interface for just the DNS providers would be extremely useful as that functionality can't be expressed in terms of Present/Cleanup :)

> I did not remove the dependency on miekg/dns because I couldn't find an elegant solution for ripping the dns check function out.

providers/dns/rfc2136 will always have that dependency, so the code will be brought in. We might as well use it elsewhere.

xenolf:
> providers/dns/rfc2136 will always have that dependency, so the code will be brought in. We might as well use it elsewhere.

Yes, I should have added a "from dns_challenge" there :smile: The plan is to somehow get rid of the `preCheckDNS` function and its supporting functions and move them somewhere else.

middelink:
> The plan is to somehow get rid of the preCheckDNS function and its supporting functions and move them somewhere else.

Are you sure you want to do that, I mean move the code into providers/dns/dns.go sure, but it seems to me you need to follow each Present up with a check to see if the change propagated all the way up. So calling it from the Solve function seems natural. Unless you want to change the "contract" on Present/Cleanup, making them do the checkDNS?

xenolf:
That's to be seen. The issue at hand is that the `miekg/dns` package is a rather large dependency which gets pulled in by default even if you are not using DNS at all.

xi2:
I have what might seem a trivial comment regarding naming. If the the providers are now in their own packages, I believe best practice as advocated by the Go team [1] would be to name the "New..." functions `NewDNSProvider` rather than `NewDNSProviderXXX` (e.g. `NewDNSProvider`, not `NewDNSProviderDigitalOcean`). Likewise the type in the digitalocean package would be just `type DNSProvider` rather than `type DNSProviderDigitalOcean`, since it would be addressed as `digitalocean.DNSProvider`.

Not a huge deal, but I believe things are slightly neater this way.

[1] <https://xi2.org/r/go/doc/effective_go.html#package-names> (see comment on "stutter")

mholt:
@xi2 That's probably a good idea, now that each provider is in its own package.

middelink:
Well, except for FindZoneByFqdn() you can probably use the std net.LookupNS and net.LookupTXT but then we are back where we were before PR 140, doing the removal of the labels in front of the fqdn till we find an authoritative NS.

xenolf:
> would be to name the "New..." functions NewDNSProvider rather than NewDNSProviderXXX

You are right, I will change the code accordingly.

> Well, except for FindZoneByFqdn() you can probably use the std net.LookupNS and net.LookupTXT but then we are back where we were before PR 140, doing the removal of the labels in front of the fqdn till we find an authoritative NS.

I don't want to remove the functionality at all. I just want to try to move it into another package.

xi2:
I have just noticed that the cloudflare, digitalocean, dnsimple and route53 providers all state in their public comment text that

    DNSProviderXXX is an implementation of the DNSProvider interface

But there is no such interface type called `DNSProvider` that I can find? They seem to mean the `ChallengeProvider` interface type. I would have put this in a separate report, but it seemed appropriate to correct it here, since you are moving and modifying the DNS code anyway.

xi2:
As a reminder, they'll also need to say `implementation of the acme.ChallengeProvider interface`, since `ChallengeProvider` is now out of package.

xi2:
Also ideally, for the reasons given in #106 and for local Godoc use, the new `dns` and provider packages should have brief package level comments.

providers/dns/digitalocean/digitalocean.go
```diff
@@ -6,6 +6,8 @@ import (
 	"fmt"
 	"net/http"
 	"sync"
+
+	"github.com/xenolf/lego/acme"
 )
 
 // DNSProviderDigitalOcean is an implementation of the DNSProvider interface
```
janeczku: This should be corrected to "an implementation of the `acme.ChallengeProvider` interface" (in the other DNS provider packages too.)

janeczku: This should be corrected to "an implementation of the `acme.ChallengeProvider` interface" (in the other DNS provider packages too.)

providers/dns/utils.go
```diff
@@ -0,0 +1,47 @@
...
+func UnFqdn(name string) string {
+	n := len(name)
+	if n != 0 && name[n-1] == '.' {
+		return name[:n-1]
+	}
+	return name
+}
+
+// WaitFor polls the given function 'f', once every 'interval' seconds, up to 'timeout' seconds.
+func WaitFor(timeout, interval int, f func() (bool, error)) error {
```
janeczku: This function is now duplicated, declared in both `providers/dns/utils.go` and `acme/dns_challenge.go`. Since it is not specific to DNS providers i suggest to declare this once in `acme/utils.go`.

janeczku:
> That's to be seen. The issue at hand is that the miekg/dns package is a rather large dependency which gets pulled in by default even if you are not using DNS at all.

Wouldn't this be solved by moving the solvers in their respective packages? 

So the client would have a method like this:

`func (c *Client) AddSolver(challenge Challenge, s Solver)`

And a program that needs an acme client with just support for dns01 challenge would do:

```Golang
import (
	github.com/xenolf/lego/acme
	github.com/xenolf/lego/acme/dns
	github.com/xenolf/lego/acme/dns/cloudflare
)

dnsSolver := dns.DnsChallenge{
        Provider: cloudflare.NewDNSProvider(...),
}
client.AddSolver(acme.DNS01, dnsSolver)

```

xenolf:
> Wouldn't this be solved by moving the solvers in their respective packages?

Well, I'm not sure that moving them out of the ACME package is the right approach. This adds further complexity to the package and makes it less obvious to use. As these issues only come into play with the DNS challenge I think the problem would get solved by just moving this functionality out into a shared package and not the whole challenge(s).

xi2:
Sorry, one more comment. As noted above by @janeczku the `WaitFor` function is duplicated. If the exported version in the `dns` package is to be kept, it would be nice if the `timeout` and `interval` parameters were of type `time.Duration`. This matches what I have used in #148. If `WaitFor` is kept unexported (i.e. `waitFor`), it would still be nice but is less important.

xenolf:
I'm sorry this took me so long but the last two weeks were hella busy :)

I added the requested changes to the PR.
What I am unsure about is if the [ClearFqdnCache](https://github.com/xenolf/lego/blob/refactor-providers/acme/dns_challenge.go#L237) function really needs to be exported as its just used in tests or if there is another way to achieve this.

xi2:
You could put `ClearFqdnCache` in an internal package (<https://golang.org/cmd/go/#hdr-Internal_Directories>)? e.g. `lego/internal/testing.go` or `lego/internal/testing/testing.go`. That way it can only be imported by other lego packages. Or maybe that's overengineering things a bit!

xi2:
By the way, thank you for implementing all the suggestions!

janeczku:
Bravo :thumbsup: :clap:

xi2:
Before you merge, xenolf, I think I've spotted a small oversight with the new `time.Duration` unit used in `acme.WaitFor`.

`time.Duration` is nanoseconds, so a caller to `waitFor` would use something like e.g. `WaitFor(3*time.Second, 1*time.Second, func() ...`. This means all calls to `WaitFor` should be changed to look like this instead of `WaitFor(3, 1 ,func() ...`.

Also, within `WaitFor` itself, you would use `timeup := time.After(timeout)` instead of `timeup := time.After(timeout * time.Second)`, and `time.Sleep(interval)` instead of `time.Sleep(interval * time.Second)`

Finally, the comment for `WaitFor` would read something like

    // WaitFor polls the given function 'f', once every 'interval', up to 'timeout'.

instead of

    // WaitFor polls the given function 'f', once every 'interval' seconds, up to 'timeout' seconds.

acme/dns_challenge.go
```diff
@@ -69,7 +68,7 @@ func (s *dnsChallenge) Solve(chlng challenge, domain string) error {
 
 	logf("[INFO][%s] Checking DNS record propagation...", domain)
 
-	err = waitFor(30, 2, func() (bool, error) {
+	err = WaitFor(30, 2, func() (bool, error) {
```
janeczku: Can we bump this up to 60 seconds while you are at it? The current value has proven to be a bit too optimistic.

xenolf:
Changed the `WaitFor` stuff around.

> You could put ClearFqdnCache in an internal package

Then I'd also have to put `fqdnToZone` into that package although this variable belongs where it is atm.

PR #150, opened 2016-03-10
Add DNS challenge support for namecheap

beevik:
This PR adds DNS challenge support for [namecheap](https://www.namecheap.com/).

If you decide to accept this PR, please feel free to delay merging it until the changes being made in https://github.com/xenolf/lego/pull/144 are complete.

### DIFF
--- a/acme/client.go
+++ b/acme/client.go
@@ -504,6 +504,7 @@ func (c *Client) requestCertificate(authz []authorizationResource, bundle bool, 
 			if len(cert) > 0 {
 
 				cerRes.CertStableURL = resp.Header.Get("Content-Location")
+				cerRes.AccountRef = c.user.GetRegistration().URI
 
 				issuedCert := pemEncode(derCertificateBytes(cert))
 				// If bundle is true, we want to return a certificate bundle.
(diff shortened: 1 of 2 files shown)


xenolf:
Thank you for this :)

I hope to be able to land #144 soon(tm) now that I've got a bit more time on my hands.

xi2:
Hi beevik. I had a quick read through of your code and I wasn't 100% sure it handles TLDs containing a period correctly (such as co.uk or or org.uk). What made me wonder is that given the domain `example.co.uk`, the `splitDomain` function would return `"co", "uk", nil` when I'd imagine it should return `"example", "co.uk", nil`. These values are then sent through to `namecheap.domains.dns.getHosts` as `SLD` and `TLD`. Am I missing something?

For an alternative way to get TLD and SLD, you can see how I do it in #133 by chopping bits off the front of the FQDN until I hit something that the provider acknowledges, or I believe you can do it by querying nameservers which is what #142 is about?

beevik:
Ah good catch.  I just added code to fix this bug.  I decided to use a namecheap API to help determine the TLD/SLD.

xi2:
Ah cool. That's a nice way of doing it.

We'll see what @xenolf thinks is the best for #148, and then we can both change our code accordingly for the timeout issue.

xenolf:
Hey @beevik! Could you please rebase this PR onto current master and move your provider into `lego/providers/dns/namecheap`. Thanks!

PR #153, opened 2016-03-14
Add DNS challenge provider for namecheap.

beevik:
Ok here is a new pull request with the namecheap DNS challenge code.  It's based on the new dns provider code organization.

Note, I had to expose the checkDNSPropagation function in acme/dns_challenge.go in order to use a custom WaitFor delay within Present.


providers/dns/namecheap/namecheap.go
```diff
@@ -0,0 +1,406 @@
...
+	"encoding/xml"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"strings"
+	"time"
+
+	"github.com/xenolf/lego/acme"
+)
```
mholt: YAY for using the standard library! :tada:

mholt:
I've only had a chance to glance at your PR, but it looks pretty solid so far! :+1:

xi2:
@beevik, if #148 gets merged you will be able to specify the custom timeout via a `Timeout` method as you did in your old PR, instead of exposing `checkDNSPropagation`. I think @xenolf is mostly happy with #148 now.

beevik:
Ok I rebased my PR onto this PR.  I also removed the change that exposed checkDNSPropagation.

xi2:
In #148 I have removed the interval growth, and instead allowed specifying the interval via the `Timeout()` method. For Gandi, I've just set it to 60 seconds. Let me know if you have objections to this.

beevik:
Rebased and updated.

Issue #60, opened 2015-12-22
Manual validation feature for http-01 and tls-sni-01 challenges


glkz:
When we cannot run an ACME client on the target server a "manual" mode for the ```http-01``` and  ```tls-sni-01``` challenges would be helpful.

Adding ```--manual``` flag to ```run``` command can be a solution imho.

Btw, Let's Encrypt client has a [plugin](https://letsencrypt.readthedocs.org/en/latest/using.html#manual) for this.

xenolf:
I had a quick look at what the "manual" plugin does for the official client.
Am I right with the assumption that lego would need to output curl commands for example which a user could use to run them on a distant machine?

glkz:
Curl commands are not necessary. I think just printing the challenge with some instructions and waiting user's action to continue would be sufficient.

For example, something like this for http-01:
```
$ lego --domains example.com --email admin@example.com run --manual
...
A HTTP get request to url
    http://example.com/.well-known/acme-challenge/{token}
must return the following string in the response body
    {jws string here}

Press ENTER when your server is ready. 
```

fenderle:
+1

uwe:
+1

gaillard:
Is there a way to use this for an automated process of what an admin _would_ do with it?

xenolf:
@gaillard Why not get a SAN cert for the 20 subdomains and then distribute them across your machines?

mholt:
Wondering if this is still necessary anymore; and even if so, if it is a good idea at all. The goal is to _automate_ the process of managing certificates. If there a manual step is possible, then we fall short of that and people continue to rely on manual ways.

PR #156, opened 2016-03-17
Add manual validation for http-01 challenge

xi2:
Since it didn't seem like a lot of code, I added support for manual validation for http-01. The requesting issue (#60) has two +1s.

I based the file `http_challenge_manual.go` on `dns_challenge_manual.go`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,11 +42,14 @@ When using the standard `--path` option, all certificates and account configurat
 
 #### Sudo
 The CLI does not require root permissions but needs to bind to port 80 and 443 for certain challenges. 
-To run the CLI without sudo, you have three options:
+To run the CLI without sudo, you have five options:
 
 - Use setcap 'cap_net_bind_service=+ep' /path/to/program
 - Pass the `--http` or/and the `--tls` option and specify a custom port to bind to. In this case you have to forward port 80/443 to these custom ports (see [Port Usage](#port-usage)).
 - Pass the `--webroot` option and specify the path to your webroot folder. In this case the challenge will be written in a file in `.well-known/acme-challenge/` inside your webroot.
+- Pass the `--manual` option. In this case, instructions for
+fulfilling the http-01 challenge manually will be output to the screen.
+- Pass the `--dns` option and specify a dns provider.
 
 #### Port Usage
 By default lego assumes it is able to bind to ports 80 and 443 to solve challenges.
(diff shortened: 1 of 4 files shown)


beevik:
Just out of curiosity, is there any reason why this provider isn't in `providers/http`?  (Same question for manual DNS)

xi2:
@beevik, I put `http_challenge_manual.go` in acme, since I just took `dns_challenge_manual.go` and modified it. I believe manual DNS is in acme, not providers, for reasons discussed in #111. Referring to manual DNS there, @xenolf said

> The manual provider should remain in the package as I think every challenge should be solvable by just importing the package.

However, that is not true of manual HTTP, so I can see an argument for putting that in providers. However I can also see an argument for keeping it in acme for symmetry with DNS manual, and the fact acme has HTTPProviderServer too.

I have no strong feelings on where it should live, so will move it wherever necessary.

xenolf:
It is not about manual providers being in the ACME package but the "default" ones. For HTTP-01 this is serving the challenge on port 80, for TLS-SNI-01 that's serving it on port 443 and for DNS-01 it's the manual one. All other providers should go into `providers/challenge/provider`.

xenolf:
> (Same question for manual DNS)

@beevik I believe that just by importing the ACME package all challenges we support should be solvable.

xi2:
Understood @xenolf. I'll move it to providers/http/manual.

xi2:
It is now living happily in providers :)

xenolf:
I'm thinking about a way of not having to add a new stand-alone CMD switch for every HTTP-01 provider :laughing:

We have `--http`, `--webroot` and with this we get `--manual`. IMO those should be brought together and all have "http" somewhere in their name to label them as belonging to the HTTP-01 challenge.

xi2:
I agree. That though crossed my mind too.

xi2:
Just a suggestion without giving it too much thought. Free up the `--http` and `--tls` flags to make them symmetrical with `--dns`. So you can then specify `--http manual` or `--http webroot`. Then specify the ports some other way, such as `--http-port` and `--tls-port`. I quite like the idea that `--dns`, `--http`, and `--dns` all do that same thing logically i.e. specify a provider.

xenolf:
Yes, I guess having a uniform interface to specify providers with `--http` `--tls` and `--dns` seems logical. For the address:port we could use `--http-address` and `--tls-address`. (Is there a name defined for the ip:port tuple?)

xi2:
Yes, sorry. I meant `--http-address`, not `--http-port`. Need more coffee. I'm not aware of a special name for ip:port. I personally think address is fine.

xenolf:
>  Need more coffee.

Yes... I do as well. Today is a productive day ;)

xi2:
:) :) If you want, I can whip up a separate PR for this. Otherwise, I'm fine to wait, since I don't use this. It just seemed that #60 seemed an easy issue to knock off.

xi2:
First attempt at flag renaming in #160.

egeland:
Really hoping for `--manual` mode as my whole workflow relies on it (I push my static content to a cloud provider bucket-as-a-website).
I can do it with the `certbot`, but `lego` seems slicker.. :grinning:

PR #157, opened 2016-03-17
Make DNS provider credential-handling more consistent.

beevik:
Different DNS providers were handling credentials in different ways.
Some were reading credential environment variables in cli_handlers.go
and then passing them into the NewDNSProvider function, while others
were reading the environment variables within their NewDNSProvider
functions.

This change moves all credential environment handling into the
providers, thereby decentralizing the handling of credentials.

### DIFF
--- a/cli_handlers.go
+++ b/cli_handlers.go
@@ -14,12 +14,12 @@ import (
 	"github.com/xenolf/lego/providers/dns/cloudflare"
 	"github.com/xenolf/lego/providers/dns/digitalocean"
 	"github.com/xenolf/lego/providers/dns/dnsimple"
+	"github.com/xenolf/lego/providers/dns/dyn"
 	"github.com/xenolf/lego/providers/dns/gandi"
 	"github.com/xenolf/lego/providers/dns/googlecloud"
 	"github.com/xenolf/lego/providers/dns/namecheap"
 	"github.com/xenolf/lego/providers/dns/rfc2136"
 	"github.com/xenolf/lego/providers/dns/route53"
-	"github.com/xenolf/lego/providers/dns/dyn"
 	"github.com/xenolf/lego/providers/http/webroot"
 )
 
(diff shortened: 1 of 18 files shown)


xi2:
I like the idea of being consistant, and I noticed that different providers were doing different things. My instinct is that it is better to keep the reading of environment variables to the lego binary, and keep the library free of that, but I don't feel strongly about it. I just don't know many libraries that read environment variables, but maybe I need to widen my horizons!

beevik:
I don't know that I really care which way it's done either, as long as it's consistent.

Is the providers/dns library intended to be something that will eventually stand alone?

mholt:
I appreciate consistency, but i do not think this is the right change.

I still think that the library should not obtain credentials, it should have them passed in by the caller.

I'm going to be using this library in an app that may receive credentials through a secure API request, not env vars.

Instead, I would need a way to pass in credentials. Why limit it to just env vars? They're easy enough to read in Go anyway, it definitely doesn't require a library to call os.Getenv.

beevik:
@mholt, my change doesn't prevent credentials from being passed in as parameters.  It simply uses the environment variables as fallbacks when you pass in empty strings for the credentials.  This is how the some of the other DNS providers -- cloudflare and dnsimple -- worked.

mholt:
@beevik Okay, gotcha. I was on my phone when I saw and wrote that comment, so I didn't give the code much of a look.

xenolf:
The driving question here is if we want to have "magic" inside the providers or not. One does kinda expect magic from the route53 provider as well as from the GCD in #141 as their clients provide that magic. For other providers I'm not sure if we should have these silent defaults.

mholt:
Even the route53 and GCD providers could maybe have a method like "LoadCredentials" to take away the magic.

Although, thinking on it more, the way it is with this PR (in that credentials that are passed in have priority, and the env var is only a fallback) is kind of nice... I, the caller, can just pass in credentials if I have them, otherwise get them from the environment. Operator error if the credentials aren't set, kind of thing. Plus it kind of standardizes the env var names across different programs that use this library.

So, now I'm +1.

xi2:
I'm kind of on the fence, but if we go this route the behaviour should probably be documented in the NewDNSProvider comment for each provider so it is unmissable?

xenolf:
> if we go this route the behaviour should probably be documented in the NewDNSProvider comment for each provider so it is unmissable?

Yes, it would need documentation. I'm fine with the providers having these ENV fallbacks but wouldn't it be more idiomatic to have a seperate `New...` function which does not accept parameters for this?

beevik:
>I'm fine with the providers having these ENV fallbacks but wouldn't it be more idiomatic to have a seperate New... function which does not accept parameters for this?

Any recommendations on what to call them?  I could call them `NewDNSProvider` for the parameterless version and `NewDNSProviderCredentials` for the version with parameters.

mholt:
@beevik That works. From what I've seen that's an idiomatic way to do it.

beevik:
Ok this is done.

PR #158, opened 2016-03-18
Optional Logger Interface

dkumor:
Having the optional logger as an interface allows plugging in custom logging libraries such as logrus, and allows integration with an app's custom logging machinery.

Also added quotes around user registration log, since without them it isn't clear that when generating an anonymous user it is just using empty string.

### DIFF
--- a/acme/client.go
+++ b/acme/client.go
@@ -161,7 +161,7 @@ func (c *Client) Register() (*RegistrationResource, error) {
 	if c == nil || c.user == nil {
 		return nil, errors.New("acme: cannot register a nil client or user")
 	}
-	logf("[INFO] acme: Registering account for %s", c.user.GetEmail())
+	logf("[INFO] acme: Registering account for \"%s\"", c.user.GetEmail())
 
 	regMsg := registrationMessage{
 		Resource: "new-reg",
--- a/acme/client.go
+++ b/acme/client.go
@@ -17,9 +17,15 @@ import (
 	"time"
 )
 
+// LoggerInterface usually used for *log.Logger, but allows support for
+// custom loggers such as Logrus
+type LoggerInterface interface {
+	Printf(string, ...interface{})
+}
+
 var (
 	// Logger is an optional custom logger.
-	Logger *log.Logger
+	Logger LoggerInterface
 )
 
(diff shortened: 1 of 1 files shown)


xenolf:
Hello @dkumor and thank you for your contribution!

I'm not sure that changing it to an interface will solve the base issue as we again add a constraint that the external logger must implement a `Printf` function. After giving this a bit of thought - wouldn't accepting an `io.Writer` solve this problem in a more general way?

dkumor:
In that case, no change is needed at all, since you can set up the stdlib logger to use an `io.Writer`:

```go
var mywriter io.Writer
acme.Logger = log.New(mywriter, "", 0),
```

Also note that the log lines are manually prefixed with [acme] in the code, but a logger can be set to use the [acme] prefix automatically.

xenolf:
>  you can set up the stdlib logger to use an io.Writer

Yes that was my thought. By allowing this to be passed in we can continue to use our internal logger but log to whatever output the user wants. Would this be a solution for you?

dkumor:
Yes - I didn't think of setting up the logger with a custom io.Writer. As it stands, this means that there are no changes needed to the code at all.

I think this PR can be closed - using an io.Writer proxy is a bit more roundabout than what I was thinking, but it will work.

xenolf:
>  using an io.Writer proxy is a bit more roundabout than what I was thinking, but it will work.

Well I guess we could make this a bit more convenient to do. As it stands now the user needs to create the logger himself and manage the io.Writer. Maybe it would make sense to just allow passing in of an `io.Writer` and let the library use it transparently.

dkumor:
Sorry, I meant it the other way around. My specific use case is with Logrus, which fits a logger interface compatible with the stdlib logger.

I was aiming at directly passing in my logger with this PR, since it would fit the interface.

As it stands, it is just a 2 step process:
```go
acme.Logger = log.New(mylogger.Writer(),"",0)
```
vs
```go
acme.Logger = mylogger
```

I don't know if replacing the optional logger with an io.Writer would be useful, especially when going from writer to logger is a one liner.

xenolf:
You are right, that was a bad thought :smile: 
I'll close this PR then as this is resolved.

PR #159, opened 2016-03-18
refactored for authzone lookup - DO CNAME fix

willseward:

### DIFF
--- a/acme/dns_challenge.go
+++ b/acme/dns_challenge.go
@@ -1,38 +1,18 @@
 package acme
 
 import (
-	"crypto/sha256"
-	"encoding/base64"
 	"errors"
 	"fmt"
 	"log"
-	"net"
-	"strings"
 	"time"
-
-	"github.com/miekg/dns"
-	"golang.org/x/net/publicsuffix"
 )
 
-type preCheckDNSFunc func(fqdn, value string) (bool, error)
+type preCheckDNSFunc func(fqdn *Domain, value string) (bool, error)
 
 var (
 	preCheckDNS preCheckDNSFunc = checkDNSPropagation
 	fqdnToZone                  = map[string]string{}
 )
-
-var recursiveNameserver = "google-public-dns-a.google.com:53"
-
-// DNS01Record returns a DNS record which will fulfill the `dns-01` challenge
-func DNS01Record(domain, keyAuth string) (fqdn string, value string, ttl int) {
-	keyAuthShaBytes := sha256.Sum256([]byte(keyAuth))
-	// base64URL encoding without padding
-	keyAuthSha := base64.URLEncoding.EncodeToString(keyAuthShaBytes[:sha256.Size])
-	value = strings.TrimRight(keyAuthSha, "=")
-	ttl = 120
-	fqdn = fmt.Sprintf("_acme-challenge.%s.", domain)
-	return
-}
 
 // dnsChallenge implements the dns-01 challenge according to ACME 7.5
 type dnsChallenge struct {
@@ -44,6 +24,8 @@ type dnsChallenge struct {
 func (s *dnsChallenge) Solve(chlng challenge, domain string) error {
 	logf("[INFO][%s] acme: Trying to solve DNS-01", domain)
 
+	verifyDomain := NewDomain(domain)
+
 	if s.provider == nil {
 		return errors.New("No DNS Provider configured")
 	}
@@ -54,18 +36,20 @@ func (s *dnsChallenge) Solve(chlng challenge, domain string) error {
 		return err
 	}
 
-	err = s.provider.Present(domain, chlng.Token, keyAuth)
+	err = s.provider.Present(verifyDomain, chlng.Token, keyAuth)
 	if err != nil {
 		return fmt.Errorf("Error presenting token %s", err)
 	}
 	defer func() {
-		err := s.provider.CleanUp(domain, chlng.Token, keyAuth)
+		err := s.provider.CleanUp(verifyDomain, chlng.Token, keyAuth)
 		if err != nil {
-			log.Printf("Error cleaning up %s %v ", domain, err)
+			log.Printf("Error cleaning up %s %v ", verifyDomain.Domain, err)
 		}
 	}()
 
-	fqdn, value, _ := DNS01Record(domain, keyAuth)
+	fqdn, value, _ := verifyDomain.GetDNS01Record(keyAuth)
+
+	txtRecord := NewDomain(fqdn)
 
 	logf("[INFO][%s] Checking DNS record propagation...", domain)
 
@@ -78,7 +62,7 @@ func (s *dnsChallenge) Solve(chlng challenge, domain string) error {
 	}
 
 	err = WaitFor(timeout, interval, func() (bool, error) {
-		return preCheckDNS(fqdn, value)
+		return preCheckDNS(txtRecord, value)
 	})
 	if err != nil {
 		return err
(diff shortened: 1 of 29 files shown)


xenolf:
You should rebase your changes on top of current master. Since you forked this repo stuff moved around quite a bit.

mholt:
Seriously tho, the last 18 minutes have been madness in here! :smile:

willseward:
Looks like it. I'll try to keep up

willseward:
Whoops. Didn't even try the tests...

willseward:
For some reason `TestDNSValidServerResponse` is still failing. Any idea why?

xi2:
I've only glanced at your code, @willseward, so my apologies. But something I noticed is that in the original `TestDNSValidServerResponse` the variable `preCheckDNS` is overridden in the test with this code:

```go
preCheckDNS = func(fqdn, value string) (bool, error) {
    return true, nil
}
```
This matters in the `dnsChallenge.Solve` method which gets called by the test, because it contains

```go
err = WaitFor(timeout, interval, func() (bool, error) {
    return preCheckDNS(fqdn, value)
})
```

With `preCheckDNS` being overridden, this WaitFor call returns immediately without error. However in your PR, this code is changed to

```go
err = WaitFor(timeout, interval, func() (bool, error) {
    return txtRecord.CheckDNSPropagation(value)
})
```

In the test in your PR, this is not overridden, so it is my guess that `WaitFor` is failing with a timeout here in the test.

willseward:
Rebased and tested!

xenolf:
Thanks @willseward! I will review this today but please allow for some time as this is quite the change :)

willseward:
Thanks. I large change wasn't really my intention, but I wanted to avoid duplicating logic as much as possible. I also tried to change the other providers as little as possible since I have no means of testing them.

acme/domain.go
```diff
@@ -0,0 +1,247 @@
...
+	if len(authoritativeNss) > 0 {
+		return authoritativeNss, nil
+	}
+	return nil, fmt.Errorf("Could not determine authoritative nameservers")
+}
+
+// findZoneByFqdn determines the zone of the given fqdn
+func (d *Domain) GetAuthoritativeZone() (string, error) {
+
+	nameserver := recursiveNameserver
```
middelink: you can replace the 2 occurrences of `nameserver` below with `recursiveNameserver`

acme/domain.go
```diff
@@ -0,0 +1,247 @@
...
+		}
+	}
+
+	if len(authoritativeNss) > 0 {
+		return authoritativeNss, nil
+	}
+	return nil, fmt.Errorf("Could not determine authoritative nameservers")
+}
+
+// findZoneByFqdn determines the zone of the given fqdn
```
middelink: // GetAuthoritativeZone determines ....

acme/domain.go
```diff
@@ -0,0 +1,247 @@
...
+
+
+type Domain struct {
+	Domain            string
+	authoritativeZone string
+	nameServers       []string
+}
+
+func NewDomain(domain string) *Domain {
+	return &Domain{Domain: domain}
```
middelink: Why not store the fqdn here once and make GetFqdn a no-op?
  return &Domain{Domain: toFqdn(domain)}

acme/domain.go
```diff
@@ -0,0 +1,247 @@
...
+	// base64URL encoding without padding
+	keyAuthSha := base64.URLEncoding.EncodeToString(keyAuthShaBytes[:sha256.Size])
+	value = strings.TrimRight(keyAuthSha, "=")
+	ttl = 120
+	fqdn = fmt.Sprintf("_acme-challenge.%s.", d.Domain)
+	return
+}
+
+func (d *Domain) GetFqdn() string {
+	return ToFqdn(d.Domain)
```
middelink: see above
  return d.Domain

acme/domain.go
```diff
@@ -0,0 +1,247 @@
...
+	}
+
+	r, err := DnsQuery(zone, dns.TypeNS, recursiveNameserver, true)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, rr := range r.Answer {
+		if ns, ok := rr.(*dns.NS); ok {
+			authoritativeNss = append(authoritativeNss, strings.ToLower(ns.Ns))
```
middelink: I was always confused as to why we lowercase the nameservers here. NSs are support to be case independent.

middelink: Also, can we add the ":53" here? So the resulting list of nameservers have a proper port.

acme/domain.go
```diff
@@ -0,0 +1,247 @@
...
+	ttl = 120
+	fqdn = fmt.Sprintf("_acme-challenge.%s.", d.Domain)
+	return
+}
+
+func (d *Domain) GetFqdn() string {
+	return ToFqdn(d.Domain)
+}
+
+// LookupNameservers returns the authoritative nameservers for the given fqdn.
```
middelink: // GetNameservers returns...

acme/domain.go
```diff
@@ -0,0 +1,247 @@
...
+	authoritativeNss, err := d.LookupNameservers()
+	if err != nil {
+		return false, err
+	}
+
+	return checkAuthoritativeNss(fqdn, value, authoritativeNss)
+}
+
+// LookupNameservers returns the authoritative nameservers for the given fqdn.
+func (d *Domain) LookupNameservers() ([]string, error) {
```
middelink: This looks suspiciously like GetNameservers()

acme/domain.go
```diff
@@ -0,0 +1,247 @@
...
+	if len(authoritativeNss) > 0 {
+		return authoritativeNss, nil
+	}
+	return nil, fmt.Errorf("Could not determine authoritative nameservers")
+}
+
+// checkAuthoritativeNss queries each of the given nameservers for the expected TXT record.
+func checkAuthoritativeNss(fqdn, value string, nameservers []string) (bool, error) {
+	for _, ns := range nameservers {
+		r, err := DnsQuery(fqdn, dns.TypeTXT, net.JoinHostPort(ns, "53"), false)
```
middelink: and this JoinHostPort should not be here, to facilitate testing.

providers/dns/rfc2136/rfc2136.go
```diff
@@ -48,27 +48,29 @@ func NewDNSProvider(nameserver, tsigAlgorithm, tsigKey, tsigSecret string) (*DNS
 }
 
 // Present creates a TXT record using the specified parameters
-func (r *DNSProvider) Present(domain, token, keyAuth string) error {
-	fqdn, value, ttl := acme.DNS01Record(domain, keyAuth)
-	return r.changeRecord("INSERT", fqdn, value, ttl)
+func (r *DNSProvider) Present(domain *acme.Domain, token, keyAuth string) error {
+	fqdn, value, ttl := domain.GetDNS01Record(keyAuth)
+	txtRecord := acme.NewDomain(fqdn)
```
middelink: you know, you could have done the fqdn > acme.Domain conversion in changeRecord() and saved yourself a bit of work ^^

willseward:
Thanks @middelink  for the suggestions. The code is updated.

There was some funniness with the namecheap provider, but it looks like that one is a sort of "outliar" as far as the implementation goes.

acme/domain.go
```diff
@@ -0,0 +1,225 @@
...
+		return name[:n-1]
+	}
+	return name
+}
+
+
+type Domain struct {
+	Domain            string
+	authoritativeZone string
+	nameServers       []string
```
middelink: is this field actually used?

acme/domain.go
middelink: I still think moving adding the port to the discovery in line 187 is a worthwhile change, as it would allow for testing of checkAuthorativeNss() with a mocked out nameserver (which will have a random port on localhost)

acme/dns_challenge.go
```diff
@@ -88,179 +72,29 @@ func (s *dnsChallenge) Solve(chlng challenge, domain string) error {
...
-		}
-	}
-
-	authoritativeNss, err := lookupNameservers(fqdn)
-	if err != nil {
-		return false, err
-	}
-
-	return checkAuthoritativeNss(fqdn, value, authoritativeNss)
+func checkDNSPropagation(fqdn *Domain, value string) (bool, error) {
```
xenolf: I'm wondering if we actually still need this when all logic is elsewhere.

acme/dns_challenge.go
```diff
@@ -88,179 +72,29 @@ func (s *dnsChallenge) Solve(chlng challenge, domain string) error {
...
-		var found bool
-		for _, rr := range r.Answer {
-			if txt, ok := rr.(*dns.TXT); ok {
-				if strings.Join(txt.Txt, "") == value {
-					found = true
-					break
-				}
-			}
+// waitFor polls the given function 'f', once every 'interval' seconds, up to 'timeout' seconds.
+func waitFor(timeout, interval int, f func() (bool, error)) error {
```
xenolf: Does this serve any purpose? I don't see it used anywhere and it really only is a duplication of the function in util.go.

acme/dns_challenge.go
```diff
@@ -78,7 +62,7 @@ func (s *dnsChallenge) Solve(chlng challenge, domain string) error {
 	}
 
 	err = WaitFor(timeout, interval, func() (bool, error) {
-		return preCheckDNS(fqdn, value)
```
willseward: @xenolf It's used here. It's possible that is was a merge mistake on my part.

acme/dns_challenge.go
willseward: That function exists to permit testing with no-op functions

willseward:
Replaced by #176

PR #160, opened 2016-03-18
CLI: Reorganize flags

xi2:
First attempt at flag renaming, discussed in #156.

This commit renames the `--http` flag to `--http-address` and `--tls` to `--tls-address`.
    
The `--webroot` flag is removed and instead webroot is specified by using `--http webroot`. The webroot path is now given in an environment variable, in the same style as the dns credentials.
    
An `httphelp` command is added to mirror the `dnshelp` command.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,15 +42,16 @@ When using the standard `--path` option, all certificates and account configurat
 
 #### Sudo
 The CLI does not require root permissions but needs to bind to port 80 and 443 for certain challenges. 
-To run the CLI without sudo, you have three options:
+To run the CLI without sudo, you have four options:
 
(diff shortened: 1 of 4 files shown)


xenolf:
I'm not sure that pushing the webroot path into the ENV is a good thing to do but I can't think of a better way too atm. Gotta think about this a bit more :smile:

xi2:
Sure. This was just meant as a starting point for some ideas, and to separate them out from #156 :) 

I'm not super keen on environment variables in the CLI at all, but I'm not sure why. It might be irrational!

xi2:
Since this PR is now well out of date and there are no more ideas on CLI flags at the moment, I'm closing it. Further comments can be left in #156.

PR #163, opened 2016-03-20
Add DNS challenge provider for Vultr.

rgl:


providers/dns/vultr/vultr.go
```diff
@@ -0,0 +1,121 @@
...
+)
+
+// DNSProvider is an implementation of the acme.ChallengeProvider interface.
+type DNSProvider struct {
+	client *vultr.Client
+}
+
+// NewDNSProvider returns a DNSProvider instance with a configured Vultr client.
+// Authentication uses the VULTR_API_KEY environment variable.
+func NewDNSProvider() (*DNSProvider, error) {
```
beevik: Could you split this into two functions: NewDNSProvider and NewDNSProviderCredentials?  That would make it consistent with all the other dns providers.

rgl: Done. But someone needs to kick the Travis builds because they are failing for some unrelated reasons: https://travis-ci.org/xenolf/lego/builds/117906594.

xenolf: I fixed the building in latest master. Please re-trigger a build by force pushing to your fork. Re-triggering the build on my end does not work as the merge commit created by github does not yet have these changes. Thanks!

mholt:
Turns out that fix doesn't work for PRs :frown: You'll have to do [something like this](https://github.com/mholt/caddy/pull/691/files) to make PRs pass again... unfortunately this doesn't raise the rate limits but it's all we can do I guess.

xenolf:
@mholt That's what I already did :)

mholt:
My bad! You got it. :+1:

providers/dns/vultr/vultr.go
rgl: Done. The build has now succeeded.

PR #164, opened 2016-03-21
Gandi DNS: Unexport Endpoint variable

xi2:
Prior to 47219ad, the Gandi DNS tests ran in a separate package from the provider itself, called `gandi_test`. Following that commit they were changed to run in the provider's package.

The `Endpoint` variable was only exported in order to be usable from `gandi_test`. Since the tests now run in `gandi`, it should be unexported and its comment changed, since the comment is now inaccurate.


xi2:
Tests are passing locally, so not sure what's up with CI.

xi2:
CI says `remote: Daily bandwidth rate limit exceeded for IP 52.0.240.122.`. I guess that could be it.

PR #168, opened 2016-03-23
Updated timeout for cloudflare dns challenge

JorritSalverda:
Updated timeout for cloudflare dns challenge record propagation to fix issue as described in https://github.com/xenolf/lego/issues/167

I haven't written a test, but the current value of 30 seconds isn't tested either. I'm not familiar enough with golang to do write it within reasonable time, so hope that's okay.


nunofgs:
I'm having problems with the current timeout, as described here https://github.com/janeczku/rancher-letsencrypt/issues/27.

Any chance the cloudflare timeout could be configurable?

PR #169, opened 2016-03-24
Gandi DNS: use acme.FindZoneByFqdn function

xi2:
Commit 8512fab modified the cloudflare, dnsimple and route53 providers to use the `acme.FindZoneByFqdn` function. This commit makes the corresponding change to the Gandi provider.

Benefits are a small reduction in code size and keeping code style similar to the other providers.

### DIFF
--- a/providers/dns/gandi/gandi.go
+++ b/providers/dns/gandi/gandi.go
@@ -20,20 +20,30 @@ import (
 // Gandi API reference:       http://doc.rpc.gandi.net/index.html
 // Gandi API domain examples: http://doc.rpc.gandi.net/domain/faq.html
 
+var (
+	// endpoint is the Gandi XML-RPC endpoint used by Present and
+	// CleanUp. It is overridden during tests.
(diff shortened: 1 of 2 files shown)


LukeHandle:
Guilty as charged. Looks good though, much cleaner :+1:

xi2:
Thanks @LukeHandle. I hope so :)

xi2:
Tests failing on Go 1.4 only. I don't think it's something introduced in this PR?

xi2:
The test failure cause would appear to be [this commit](https://github.com/GoogleCloudPlatform/gcloud-golang/commit/f95c875f2894b735501920cce2ab858f2110b23a) in `google.golang.org/cloud/compute/metadata` which is used by `google.golang.org/api/dns/v1` which is used by the `googlecloud` provider. It uses `http.Request.Cancel` which is Go 1.5+. It's being discussed in [this issue](https://github.com/GoogleCloudPlatform/gcloud-golang/issues/230).

xenolf:
> Looks good though, much cleaner

I agree :)

PR #170, opened 2016-03-24
Update README usage Section

kennyballou:
The usage output seems to have drifted and was causes confusion when
attempting to use as a reference.


xenolf:
Thanks for pointing it out and fixing! :+1:

PR #172, opened 2016-03-24
Update README example section

xi2:
The example in README no longer compiles, due to new signature of `acme.NewClient` (taking a `KeyType`). This PR hopefully fixes that.


xenolf:
My bad! Totally forgot about the README :worried: 
Thanks for fixing :)

PR #176, opened 2016-03-27
DigitalOcean uses FindZoneByFqdn

willseward:
It looks like there's been enough refactoring so that my last PR is unnecessary. Here's a quick fix using the new solution.

I'll go ahead and close the other PR

### DIFF
--- a/providers/dns/digitalocean/digitalocean.go
+++ b/providers/dns/digitalocean/digitalocean.go
@@ -63,7 +63,14 @@ func (d *DNSProvider) Present(domain, token, keyAuth string) error {
 
 	fqdn, value, _ := acme.DNS01Record(domain, keyAuth)
 
-	reqURL := fmt.Sprintf("%s/v2/domains/%s/records", digitalOceanBaseURL, domain)
+	authZone, err := acme.FindZoneByFqdn(acme.ToFqdn(domain), acme.RecursiveNameserver)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


LukeHandle:
Before this change, was DO unable to handle verifying multi-level sub domains? I remember looking at it before and thinking it could based on their API?

Eg. Auth zone is example.com, you want to verify this.is.a.subdomain.example.com

Either way, this PR brings it in-line with the recent changes and it makes more sense this way :+1:

willseward:
Yes, DO could not verify CNAMEs. If the user wants to verify a.example.com but the zone was setup as so that a.exmple.com was a CNAME in the zone example.com, the provider would make an API request for a.example.com which returned a 404.

janeczku:
Nit, but this bug affects any type of subdomain record, not just CNAME type records.

willseward:
Agreed. Just citing a trivial example. This should work for all types of subdomain records.

xenolf:
LGTM :+1:

willseward:
Is there any timeline for a release with this change?

xenolf:
@willseward I'd like to get #179 fixed and then I'll issue a point release.

LukeHandle:
Hey @xenolf, have you had a chance to look at #186  which would fix #179?

xenolf:
FYI v0.3.1 has been released.

willseward:
Thanks @xenolf

PR #177, opened 2016-03-27
Build with go-jose.v1 (instead of master)

csstaub:
Changes the import path for go-jose to pin to the `v1` branch (via gopkg.in). I maintain backwards-compatibility for the `v1` branch, whereas `master` can diverge over time. This is to make sure I don't end up breaking your build :-). Thanks.


xenolf:
Thanks for this! :)

PR #182, opened 2016-04-04
Only try to parse JSON documents

### LINKED ISSUE #181: Fail gracefully when directory is not available
First: Thanks for writing lego!

I was just running lego and it gave me the following error message:

```
$ lego --email user@domain.tld run
2016/04/04 19:12:55 No key found for account user@domain.tld. Generating a curve P384 EC key.
2016/04/04 19:12:55 Saved key to /home/user/fd0/.lego/accounts/acme-v01.api.letsencrypt.org/user@domain.tld/keys/user@domain.tld.key
2016/04/04 19:12:56 Could not create client: get directory at 'https://acme-v01.api.letsencrypt.org/directory': invalid character '<' looking for beginning of value
[1]    16230 exit 1     lego --email user@domain.tld run
```

This is apparently caused by maintenance mode of the directory, requesting the URL with curl yields:

```
$ curl -i https://acme-v01.api.letsencrypt.org/directory
HTTP/1.1 502 Bad Gateway
Server: nginx
Content-Type: text/html
Content-Length: 1488
ETag: "55f255cc-5d0"
Expires: Mon, 04 Apr 2016 17:13:13 GMT
Cache-Control: max-age=0, no-cache, no-store
Pragma: no-cache
Date: Mon, 04 Apr 2016 17:13:13 GMT
Connection: keep-alive

<!DOCTYPE html>

<html lang="en">
<head>
  <meta charset="utf-8">
  <meta http-equiv="X-UA-Compatible" content="IE=edge">
  <meta name="viewport" content=
  "width=device-width, initial-scale=1">

  <title>Maintenance</title>
  <link href=
  "//maxcdn.bootstrapcdn.com/bootstrap/3.2.0/css/bootstrap.min.css"
  rel="stylesheet" type="text/css">
  <link href=
  "//maxcdn.bootstrapcdn.com/font-awesome/4.2.0/css/font-awesome.min.css"
  rel="stylesheet" type="text/css">
</head>

<body>
  <div class="container-fluid">
    <div class="row">
      <div class="col-md-6 text-right">
        <p style="font-size: 90px;">
<i class="fa fa-coffee"></i></p>
        </div>

      <div class="col-md-6 text-left">
        <h1>502<br>
        <small>Maintenance Time</small></h1>
      </div>
    </div>

    <div class="row">
      <div class="col-xs-12 text-center">
        <p>Sorry! We're doing some cleanup at the moment. We'll be back shortly.</p>
      </div>
    </div>
    <div class="row">
      <div class="col-xs-4 col-xs-offset-2 text-center">
        <p><a href="https://letsencrypt.status.io" title="Twitter">
          <i class="fa fa-area-chart"></i>
          Service Status (letsencrypt.status.io)
        </a></p>
      </div>
      <div class="col-xs-4 text-center">
        <p><a href="https://twitter.com/letsencrypt" title="Twitter">
          <i class="fa fa-twitter"></i>
          Check with us on Twitter
        </a></p>
      </div>
    </div> <!-- row -->
  </div>


</body>
</html>
```

I think it's a good idea to fail gracefully in this case, e.g. by looking at the HTTP response code.

fd0:
This patch adds code to only parse the HTTP response body as JSON if the content-type header advertises the content as JSON. In my case, the directory server was unavailable: it returned a 503 HTTP response code with an HTML document, and the only thing lego reported was:

    2016/04/04 19:12:56 Could not create client: get directory at 'https://acme-v01.api.letsencrypt.org/directory': invalid character '<' looking for beginning of value

This was caused by trying to parse the document body (HTML) as JSON, without looking at the content-type header and returning the JSON parse error.

Closes #181

### DIFF
--- a/acme/error.go
+++ b/acme/error.go
@@ -52,10 +52,14 @@ func (c challengeError) Error() string {
 
 func handleHTTPError(resp *http.Response) error {
 	var errorDetail RemoteError
-	decoder := json.NewDecoder(resp.Body)
-	err := decoder.Decode(&errorDetail)
-	if err != nil {
-		return err
+
+	// try to decode the content as JSON
+	if resp.Header.Get("Content-Type") == "application/json" {
+		decoder := json.NewDecoder(resp.Body)
+		err := decoder.Decode(&errorDetail)
+		if err != nil {
+			return err
+		}
 	}
 
 	errorDetail.StatusCode = resp.StatusCode


mholt:
It'd be nice if the Detail field could be populated with the HTML content or a status message or something, in the case that the page isn't JSON.

fd0:
Yes, that'd be nice indeed. Maybe it's a second issue to add sugar to it, I'm satisfied with the error message as it is now (after adding my commit).

Maybe extract the title from the HTML document (if it's HTML), does code for that already exist in the stdlib? I don't think it's a good idea to dump the source code of the error document.

mholt:
You may be right. It's kinda too bad their error page is HTML but their working page is JSON...

fd0:
Btw, the error message looks like this now (for a 404 error with an HTML error page):
```
2016/04/04 23:31:13 Could not create client: get directory at 'http://localhost:3000/`': acme: Error 404 -  - 
```

fd0:
We could add the HTTP code description (the text after the status code) though, in this case that'd be `Bad Gateway`, which is not really better than `502` alone...

xenolf:
Thanks for bringing this issue up and fixing it!

I think this is fine as implemented here, HTML parsing would probably go one step too far.

PR #183, opened 2016-04-07
Add Dockerfile

koep:
This enables users to build, ship and run lego with Docker.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,17 @@
+FROM alpine:3.3
+
+ENV GOPATH /go
+
+RUN apk update && apk add ca-certificates go git && \
+    rm -rf /var/cache/apk/*
+
+COPY . /go/src/github.com/xenolf/lego
+
+RUN cd /go/src/github.com/xenolf/lego && \
+	go get ./... && \
+	go build -o /usr/bin/lego . && \
+	apk del ca-certificates go git && \
+	rm -rf /var/cache/apk/* && \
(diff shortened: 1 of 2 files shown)


mholt:
Cool. But, why? (Not trolling. Honest question.) I don't use Docker, a lot of people do, but what's the benefit of shipping a completely self-contained, static Go binary in a Docker container? Why introduce the overhead, why layer on the complexity?

*Note: CI build failure is unrelated to your PR (obviously).*

koep:
Hi,

this blog post should answer your questions ;-) 

https://medium.com/@kelseyhightower/optimizing-docker-images-for-static-binaries-b5696e26eb07#.p5bnz779z

mholt:
Thanks for the link!

So basically:

> Deploying applications with Docker brings the benefits of Linux containers. I essentially gain the ability to leverage Linux namespaces and cgroups for free.

I still don't see why that is valuable for a program like lego. :confused:

To be clear, I'm not opposed to your contribution -- very grateful for it! I just want to be sure I can appreciate it as much as the next user...

koep:
I personally use docker to build lego from source without installing golang on my host.

It may also be possible to use lego on platforms that use docker containers exclusively (I haven't had the chance to test this though)
- CoreOS
- Kubernetes

etc.

xenolf:
Well, I see no harm in having a docker file in case anyone wants to use it :)

Issue #173, opened 2016-03-25
Improvement: setting dns-01 challenge record for multiple domains in parallel


JorritSalverda:
When running lego for multiple domains it will set the dns TXT record and check propagation in sequence. This can take a long time if the number of domains is large. Would it be possible to parallelize this?

xenolf:
Well it sure is possible. @mholt and me already had a talk about this and we came to agree that it would be a great feature. The ACME challenges are good to run concurrent as the context for each of them should be seperate. What we need to do though is look all over to be sure that no state is shared.

So yes, this is something I'd like to add in the future :)

lenovouser:
Yeah, this would be very helpful if you have a lot of domains.

mthoretton:
Just another comment to add you some additional pressure about developing this great functionality ;) Btw lego works fine with route53! Thanks

xenolf:
I've started to work on this (for all challenges) locally and will push once I have something to show ;)

xi2:
Thank you for working on this. I will be happy to test and/or update the gandi provider as required. :)

aaronb-cg:
I would like to see this also.
Lego is working great for us, just takes a long time with the large number of domains we're processing.

lenovouser:
Any progress?

jasonwbarnett:
Would love to collaborate to get this done. I looked over the code tonight and identified where the modification should go, but haven't had time to brainstorm the best design, e.g. worker pool, go routines, etc...

jasonwbarnett:
I leveraged #237 as a starting point and was able to make things more efficient. I was able to obtain a SSL certificate with 86 SANs in 6 minutes and 24 seconds.

I was using the following:

Challenge Type: DNS-01
DNS Provider: Google DNS
TXT Record TTL: 10 seconds
zone negative cache TTL: 10 seconds

Thi

xenolf:
Hey @jasonwbarnett and thanks for your time!
I had a quick look at your implementation. Please keep in mind that many of the DNS providers currently shipped with lego do have state per instance of the provider (for caching the domain name for example). Your implementation only calls the `solve()` function in parallel which will overwrite this state.

jasonwbarnett:
@xenolf Where did you get with your implementation? Do you have it published anywhere?

Wondering if it makes sense to implement the concurrency in the provider itself or what... Or if it's just a matter of refactoring the providers which currently have a shared state.

xenolf:
[...]

PR #185, opened 2016-04-10
Fix out of range

rekby:
Some time (near once per 50-100 certificates) lego fails with stack:
panic: runtime error: index out of range

goroutine 1906 [running]:
panic(0x8781a0, 0xc8200100b0)
        C:/Go/src/runtime/panic.go:464 +0x3e6
github.com/xenolf/lego/acme.(*jws).Nonce(0xc820374300, 0x0, 0x0, 0x0, 0x0)
        c:/Users/Tim/Documents/go/src/github.com/xenolf/lego/acme/jws.go:105 +0x176
gopkg.in/square/go-jose%2ev1.(*genericSigner).Sign(0xc8203923c0, 0xc8205a6240, 0x6b, 0x8b, 0x6d0778650c48, 0x0, 0x0)
        c:/Users/Tim/Documents/go/src/gopkg.in/square/go-jose.v1/signing.go:157 +0x376
github.com/xenolf/lego/acme.(*jws).signContent(0xc820374300, 0xc8205a6240, 0x6b, 0x8b, 0x481c4b, 0x0, 0x0)
        c:/Users/Tim/Documents/go/src/github.com/xenolf/lego/acme/jws.go:70 +0x1f5
github.com/xenolf/lego/acme.(*jws).post(0xc820374300, 0xc8200144c0, 0x33, 0xc8205a6240, 0x6b, 0x8b, 0x0, 0x0, 0x0)
        c:/Users/Tim/Documents/go/src/github.com/xenolf/lego/acme/jws.go:35 +0x69
github.com/xenolf/lego/acme.postJSON(0xc820374300, 0xc8200144c0, 0x33, 0x8bd040, 0xc8205a61b0, 0x79dc80, 0xc8205a6120, 0x0, 0x0, 0x0)
        c:/Users/Tim/Documents/go/src/github.com/xenolf/lego/acme/http.go:96 +0x1df
github.com/xenolf/lego/acme.(*Client).getChallenges.func1(0xc820024510, 0xc82022c780, 0xc82022c720, 0xc8204342b9, 0xb)
        c:/Users/Tim/Documents/go/src/github.com/xenolf/lego/acme/client.go:408 +0x1a7
created by github.com/xenolf/lego/acme.(*Client).getChallenges
        c:/Users/Tim/Documents/go/src/github.com/xenolf/lego/acme/client.go:421 +0x17c

### DIFF
--- a/acme/jws.go
+++ b/acme/jws.go
@@ -101,7 +101,9 @@ func (j *jws) Nonce() (string, error) {
 		if err != nil {
 			return nonce, err
 		}
-		return "", errors.New("Can't get nonce")
+		if len(j.nonces) == 0 {
+			return "", errors.New("Can't get nonce")
+		}
 	}
 
 	nonce, j.nonces = j.nonces[len(j.nonces)-1], j.nonces[:len(j.nonces)-1]
--- a/acme/jws.go
+++ b/acme/jws.go
@@ -10,6 +10,7 @@ import (
 	"net/http"
 
 	"gopkg.in/square/go-jose.v1"
+	"errors"
 )
 
 type jws struct {
@@ -100,6 +101,7 @@ func (j *jws) Nonce() (string, error) {
 		if err != nil {
 			return nonce, err
 		}
+		return "", errors.New("Can't get nonce")
 	}
 
 	nonce, j.nonces = j.nonces[len(j.nonces)-1], j.nonces[:len(j.nonces)-1]
--- a/acme/jws.go
+++ b/acme/jws.go
@@ -10,6 +10,7 @@ import (
 	"net/http"
 
 	"gopkg.in/square/go-jose.v1"
+	"errors"
 )
 
 type jws struct {
@@ -99,6 +100,9 @@ func (j *jws) Nonce() (string, error) {
 		err := j.getNonce()
 		if err != nil {
 			return nonce, err
+		}
+		if len(j.nonces) == 0 {
+			return "", errors.New("Can't get nonce")
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


xenolf:
Hey and thanks for figuring this out and fixing it! :)

I've noticed that this change broke the tests. Would you mind fixing them up?

rekby:
Done

xenolf:
Thanks for fixing the tests.
I'm afraid the retry you added has no effect tho. We only ever have one running request at any given time and if there is no nonce for this request, there will never be with the current code. It would be a better idea to call `getNonce()` if we are starved of nonces to get a fresh one.

rekby:
I have checked the sources:
getNonce has unconditional call httpHead at first line.
httpHead create new request and get header every time.

re-call getNonce will try get http headers every time.

Or I don't understand what you want say to me.

xenolf:
No you are right, `getNonce` will make a HEAD request to the ACME server. This is intentional as new nonces only come from the server. In order to get a new one we have to hit the server and extract the new nonce from the headers.

rekby:
Ok. I revert patch to simple - without retry, but it is workaround only.

I research the code and I see - the error impossible in single thread code. It may be only race condition.
calls getNonce -> getNonceFromResponse always return error or j.nonces length more then zero by append nonce in getNonceFromResponse
	j.nonces = append(j.nonces, nonce)

May be many goroutines work with the nonces same time and it need guard by mutex or redesign working with nonces?

rekby:
I use lego as library in own code. My code is single thread/goroutine, but it is long running loop: I retrieving many hundreds certificates in a loop without restart program.

xenolf:
> My code is single thread/goroutine, but it is long running loop: I retrieving many hundreds certificates in a loop without restart program.

That shouldn't be a problem. The client instance itself isn't thread safe yet tho. See #173.

Issue #179, opened 2016-03-31
DNS providers fail to present token (read udp i/o timeout)


janeczku:
Since #142  i am getting intermittent errors with DNS providers:

`level=error msg="[xxx.xxx.xxx] Error obtaining certificate: Error presenting token read udp xxx.xxx.xxx:57113->8.8.8.8:53: i/o timeout"`

As discussed in https://github.com/xenolf/lego/pull/142#issuecomment-189884283 the `acme.FindZoneByFqdn` function doesn't handle connection errors when sending a query to the nameserver. Since DNS uses UDP a single lost/dropped packet will cause the connection to fail. This doesn't matter when the function is already wrapped in a retry loop but #142 uses this function directly to lookup the zone used in API queries.

@LukeHandle

LukeHandle:
As before, I suggest we add such functionality in the `dnsQuery` function.

How many times would you consider best for retrying here, would you suggest switching to TCP in the event of an issue for some more reliability?

Finally, do you suggest we only retry when the error is specifically connection issues, or is it worth retrying on any error?

LukeHandle:
Another thing, where abouts was this error being generated? I was struggling to find the strings in the message eg. https://github.com/xenolf/lego/search?utf8=%E2%9C%93&q=Error+obtaining+certificate&type=Code

```
grep -R "certificate:" .                                                        
/acme/client.go:                                logf("[ERROR][%s] acme: Could not bundle issuer certificate: %v", cert.Domain, err)
/acme/client.go:                                                logf("[WARNING][%s] acme: Could not bundle issuer certificate: %v", commonName.Domain, err)
README.md:Obtain a certificate:
README.md:To renew the certificate:
```

janeczku:
> How many times would you consider best for retrying here, would you suggest switching to TCP in the event of an issue for some more reliability?

1. I would rather fallback to the alternate Google DNS (8.8.4.4) when the initial query failed. DNS over TCP is not really purposed for redundancy. It's also worth noting that Google's Public DNS servers throttles clients exceeding a certain amount of queries per minute. They do this by silently dropping the request which looks like a lost packet from our side. If we are rate limited switching to TCP won't help but switching to the alternate server will.
I don't have a strong opinion on how many times we should retry. Maybe max three tries total?

> Finally, do you suggest we only retry when the error is specifically connection issues, or is it worth retrying on any error?

[...]

PR #186, opened 2016-04-10
Retry logic for dnsQuery

LukeHandle:
With regard to the issues seen in #179

 * Logic to retry the DNS query up 3 times (respects truncated errors)
 * Added an alternate recursive nameserver
 * Logic to use alt NS after 1st failure (only when the default NS is used)
 * Adjusted formatting of 2 errors to differentiate the returned error string

### DIFF
--- a/acme/dns_challenge.go
+++ b/acme/dns_challenge.go
@@ -21,7 +21,10 @@ var (
 	fqdnToZone                  = map[string]string{}
 )
 
-var RecursiveNameserver = "google-public-dns-a.google.com:53"
+var RecursiveNameservers = []string{
+	"google-public-dns-a.google.com:53",
+	"google-public-dns-b.google.com:53",
+}
 
 // DNS01Record returns a DNS record which will fulfill the `dns-01` challenge
 func DNS01Record(domain, keyAuth string) (fqdn string, value string, ttl int) {
@@ -56,12 +59,12 @@ func (s *dnsChallenge) Solve(chlng challenge, domain string) error {
 
 	err = s.provider.Present(domain, chlng.Token, keyAuth)
 	if err != nil {
-		return fmt.Errorf("Error presenting token %s", err)
+		return fmt.Errorf("Error presenting token: %s", err)
 	}
 	defer func() {
 		err := s.provider.CleanUp(domain, chlng.Token, keyAuth)
 		if err != nil {
-			log.Printf("Error cleaning up %s %v ", domain, err)
+			log.Printf("Error cleaning up %s: %v ", domain, err)
 		}
 	}()
 
@@ -90,7 +93,7 @@ func (s *dnsChallenge) Solve(chlng challenge, domain string) error {
 // checkDNSPropagation checks if the expected TXT record has been propagated to all authoritative nameservers.
 func checkDNSPropagation(fqdn, value string) (bool, error) {
 	// Initial attempt to resolve at the recursive NS
-	r, err := dnsQuery(fqdn, dns.TypeTXT, RecursiveNameserver, true)
+	r, err := dnsQuery(fqdn, dns.TypeTXT, RecursiveNameservers, true)
 	if err != nil {
 		return false, err
 	}
@@ -117,7 +120,7 @@ func checkDNSPropagation(fqdn, value string) (bool, error) {
 // checkAuthoritativeNss queries each of the given nameservers for the expected TXT record.
 func checkAuthoritativeNss(fqdn, value string, nameservers []string) (bool, error) {
 	for _, ns := range nameservers {
-		r, err := dnsQuery(fqdn, dns.TypeTXT, net.JoinHostPort(ns, "53"), false)
+		r, err := dnsQuery(fqdn, dns.TypeTXT, []string{net.JoinHostPort(ns, "53")}, false)
 		if err != nil {
 			return false, err
 		}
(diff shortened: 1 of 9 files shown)


acme/dns_challenge.go
```diff
@@ -155,12 +157,28 @@ func dnsQuery(fqdn string, rtype uint16, nameserver string, recursive bool) (in
...
+			tcp := &dns.Client{Net: "tcp"}
+			// The following should reset the err to be nil
+			in, _, err = tcp.Exchange(m, nameserver)
+		}
+
+		if err == nil {
+			return
+		}
+
+		logf("[WARNING][%s] acme: DNS Query #%d failed: %s", fqdn, i, err)
```
janeczku: A dropped UDP packet isn't really something i would warn the user about as long as we are still able to solve by retrying.

acme/dns_challenge.go
```diff
@@ -144,7 +145,8 @@ func checkAuthoritativeNss(fqdn, value string, nameservers []string) (bool, erro
 	return true, nil
 }
 
-// dnsQuery sends a DNS query to the given nameserver.
+// dnsQuery will attempt 3 times, a query to the given nameserver.
+// If the given nameserver is `RecursiveNameserver`, it will attempt the `altRecursiveNameserver` on the 2nd and 3rd attempts.
 // The nameserver should include a port, to facilitate testing where we talk to a mock dns server.
 func dnsQuery(fqdn string, rtype uint16, nameserver string, recursive bool) (in *dns.Msg, err error) {
```
janeczku: Just an idea: But since we now use multiple recursive server, what do you think about making dnsQuery() and FindZoneByFqdn() accept a slice of nameservers and declare the default servers like this:

```Golang
var RecursiveNameservers = []string{
	"google-public-dns-a.google.com:53",
	"google-public-dns-b.google.com:53",
}
```

This would make it also quite simple to actually alternate between servers on each retry:

```Golang
	// Try up to 3 times (returning early if nil errors)
	for i := 1; i <= 3; i++ {
		ns := nameservers[i % len(nameservers)]
		in, err = dns.Exchange(m, ns)
```

acme/dns_challenge.go
LukeHandle: Fair enough, originally I added it for debugging type info but kinda liked it in the end. I guess the important bit is that we are passing the error message when it does fail the 3 times.

acme/dns_challenge.go
LukeHandle: I considered it as an option. In the end for the sake of simplicity (and less changes)  I just went ahead as I did. I didn't think the proper alternation was worth it (it's just neater?). 

The slice 🍕 does add more flexibility when calling `FindZoneByFqdn()` which is nice, but almost all the providers I think use the default nameserver anyway?

I guess, my final opinion is "This is a better way to do it, and therefore we should" - even if my short-sightedness doesn't necessarily appreciate its application ;) 

In making this change, how about also adding either a variable retry, or set it to len(nameservers)+1?

And it'll be a couple of days before I can take another proper look at this.

janeczku: > I guess, my final opinion is "This is a better way to do it, and therefore we should" - even if my short-sightedness doesn't necessarily appreciate its application ;)

Glad we are on the same page here (minus the short-sightedness) 😄 
rfc2136 is the only provider that passes a custom nameserver to `FindZoneByFqdn()`.

> In making this change, how about also adding either a variable retry, or set it to len(nameservers)+1?

Either is fine with me.

LukeHandle:
Okay changes made as discussed.

Minor point about maybe returning the nameserver used which I will highlight in a second. Otherwise, looks good?

Also, a slightly strange, though functionally, non-issue is the order in which the nameservers are attempted. See [here](https://play.golang.org/p/McT5BgW23c).

Any more feedback appreciated.

acme/dns_challenge.go
```diff
@@ -203,13 +215,13 @@ func FindZoneByFqdn(fqdn, nameserver string) (string, error) {
...
 	//   NXDOMAIN or indicating that no data (NODATA) of the requested type exists
-	in, err := dnsQuery(fqdn, dns.TypeSOA, nameserver, true)
+	in, err := dnsQuery(fqdn, dns.TypeSOA, nameservers, true)
 	if err != nil {
 		return "", err
 	}
 	if in.Rcode != dns.RcodeNameError {
 		if in.Rcode != dns.RcodeSuccess {
-			return "", fmt.Errorf("NS %s returned %s for %s", nameserver, dns.RcodeToString[in.Rcode], fqdn)
+			return "", fmt.Errorf("The NS returned %s for %s", dns.RcodeToString[in.Rcode], fqdn)
```
LukeHandle: If dnsQuery were to return the nameserver it used to make the query, we could allow the error to be more informative again. Do we consider this as issue or not?

For the sake of committing something, I've gone with not specifically mentioning the nameserver here and [below](https://github.com/xenolf/lego/pull/186/files#diff-3dd431e0b068070def9558d0daeef06cR240)

acme/dns_challenge.go
```diff
@@ -155,25 +158,34 @@ func dnsQuery(fqdn string, rtype uint16, nameserver string, recursive bool) (in
 		m.RecursionDesired = false
 	}
 
-	in, err = dns.Exchange(m, nameserver)
-	if err == dns.ErrTruncated {
-		tcp := &dns.Client{Net: "tcp"}
-		in, _, err = tcp.Exchange(m, nameserver)
-	}
+	// Will rety the request based on the number of servers (n+1)
```
janeczku: retry

acme/dns_challenge.go
janeczku: OK

janeczku:
@LukeHandle Just ran a couple of live tests with Route53 and CloudFlare. I also made sure that the retry works by dropping packets to one of the nameservers. Everything checks out.

@xenolf LGTM

kunalpatilabg:
NGROK Unable to connect .. Any insight?

image

lenovouser:
@kunalpatilabg wrong repository? 😄

PR #187, opened 2016-04-11
Webroot cleanup

rekby:
webroot provides has bad cleanup: the provider create folders and token file, but remove token file only.

The patch remove created folders too (if they are empty).


xenolf:
I do not really approve of the loop this PR is using as we only need to remove the file and the folder.
Furthermore, the changes which will hit in the next few days won't work with this either as there could be more then one webroot request active at any given time.

PR #193, opened 2016-04-17
Fix broken DNSimple provider

janeczku:
Requesting a certificate using DNSimple provider results in the  following error:
> Error obtaining certificate: Error presenting token dns: domain must be fully qualified


mholt:
:+1: Glad you caught this. LGTM

PR #200, opened 2016-04-20
Expose some internals to allow building webui.

codefx9:
Hi,
We would like to build a webui around DNS challenge type. The idea is that users will be able to see the DNS challenges they need to configure manually in a web page instead of from command line. A background process can periodically check for DNS record and call Validate() when `CheckDNSPropagation()` is successful.

It will be great if you can give some feedback on this pull request. I can cleanup the comments if the pull request is directionally ok.

Thanks.

### DIFF
--- a/acme/client.go
+++ b/acme/client.go
@@ -42,10 +42,10 @@ type User interface {
 
 // Interface for all challenge solvers to implement.
 type solver interface {
-	Solve(challenge challenge, domain string) error
+	Solve(challenge IDChallenge, domain string) error
 }
 
-type validateFunc func(j *jws, domain, uri string, chlng challenge) error
+type validateFunc func(j *jws, domain, uri string, chlng IDChallenge) error
 
 // Client is the user-friendy way to ACME
 type Client struct {
@@ -230,7 +230,7 @@ func (c *Client) ObtainCertificate(domains []string, bundle bool, privKey crypto
 
 	logf("[INFO][%s] acme: Validations succeeded; requesting certificates", strings.Join(domains, ", "))
 
-	cert, err := c.requestCertificate(challenges, bundle, privKey)
+	cert, err := c.RequestCertificate(challenges, bundle, privKey)
 	if err != nil {
 		for _, chln := range challenges {
 			failures[chln.Domain] = err
@@ -445,7 +445,7 @@ func (c *Client) getChallenges(domains []string) ([]authorizationResource, map[s
 	return challenges, failures
 }
 
-func (c *Client) requestCertificate(authz []authorizationResource, bundle bool, privKey crypto.PrivateKey) (CertificateResource, error) {
+func (c *Client) RequestCertificate(authz []authorizationResource, bundle bool, privKey crypto.PrivateKey) (CertificateResource, error) {
 	if len(authz) == 0 {
 		return CertificateResource{}, errors.New("Passed no authorizations to requestCertificate!")
 	}
@@ -601,8 +601,8 @@ func parseLinks(links []string) map[string]string {
 
 // validate makes the ACME server start validating a
 // challenge response, only returning once it is done.
-func validate(j *jws, domain, uri string, chlng challenge) error {
-	var challengeResponse challenge
+func validate(j *jws, domain, uri string, chlng IDChallenge) error {
+	var challengeResponse IDChallenge
 
 	hdr, err := postJSON(j, uri, chlng, &challengeResponse)
 	if err != nil {
--- a/acme/dns_challenge.go
+++ b/acme/dns_challenge.go
@@ -17,7 +17,7 @@ import (
 type preCheckDNSFunc func(fqdn, value string) (bool, error)
 
 var (
-	preCheckDNS preCheckDNSFunc = checkDNSPropagation
+	preCheckDNS preCheckDNSFunc = CheckDNSPropagation
 	fqdnToZone                  = map[string]string{}
 )
 
@@ -44,7 +44,7 @@ type dnsChallenge struct {
 	provider ChallengeProvider
 }
 
-func (s *dnsChallenge) Solve(chlng challenge, domain string) error {
+func (s *dnsChallenge) Solve(chlng IDChallenge, domain string) error {
 	logf("[INFO][%s] acme: Trying to solve DNS-01", domain)
 
 	if s.provider == nil {
@@ -87,11 +87,11 @@ func (s *dnsChallenge) Solve(chlng challenge, domain string) error {
 		return err
 	}
 
-	return s.validate(s.jws, domain, chlng.URI, challenge{Resource: "challenge", Type: chlng.Type, Token: chlng.Token, KeyAuthorization: keyAuth})
+	return s.validate(s.jws, domain, chlng.URI, IDChallenge{Resource: "challenge", Type: chlng.Type, Token: chlng.Token, KeyAuthorization: keyAuth})
 }
 
-// checkDNSPropagation checks if the expected TXT record has been propagated to all authoritative nameservers.
-func checkDNSPropagation(fqdn, value string) (bool, error) {
+// CheckDNSPropagation checks if the expected TXT record has been propagated to all authoritative nameservers.
+func CheckDNSPropagation(fqdn, value string) (bool, error) {
 	// Initial attempt to resolve at the recursive NS
 	r, err := dnsQuery(fqdn, dns.TypeTXT, RecursiveNameservers, true)
 	if err != nil {
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 lego.exe
 lego
 .lego
+.idea
--- a/acme/client_test.go
+++ b/acme/client_test.go
@@ -129,12 +129,12 @@ func TestValidate(t *testing.T) {
 		case "POST":
 			st := statuses[0]
 			statuses = statuses[1:]
-			writeJSONResponse(w, &challenge{Type: "http-01", Status: st, URI: "http://example.com/", Token: "token"})
+			writeJSONResponse(w, &IDChallenge{Type: "http-01", Status: st, URI: "http://example.com/", Token: "token"})
 
 		case "GET":
 			st := statuses[0]
 			statuses = statuses[1:]
-			writeJSONResponse(w, &challenge{Type: "http-01", Status: st, URI: "http://example.com/", Token: "token"})
+			writeJSONResponse(w, &IDChallenge{Type: "http-01", Status: st, URI: "http://example.com/", Token: "token"})
 
 		default:
 			http.Error(w, r.Method, http.StatusMethodNotAllowed)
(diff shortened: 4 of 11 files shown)


acme/challenges.go
```diff
@@ -1,16 +1,16 @@
 package acme
 
 // Challenge is a string that identifies a particular type and version of ACME challenge.
-type Challenge string
+type ChallengeType string
```
mholt: I like this rename, but please don't forget the godoc; this'll emit a lint warning as-is.

acme/client.go
```diff
@@ -42,10 +42,14 @@ type User interface {
 
 // Interface for all challenge solvers to implement.
 type solver interface {
-	Solve(challenge challenge, domain string) error
+	Present(challenge Challenge, domain string) (string, error)
+	Validate(j *jws, domain, uri string, chlng Challenge) error
+
+	// Present + Validate
+	Solve(challenge Challenge, domain string) error
```
mholt: I'm not sure how I feel about an interface method that is redundant.

acme/client.go
```diff
@@ -89,22 +93,22 @@ func NewClient(caDirURL string, user User, keyType KeyType) (*Client, error) {
...
 	return &Client{directory: dir, user: user, jws: jws, keyType: keyType, solvers: solvers}, nil
 }
 
 // SetChallengeProvider specifies a custom provider that will make the solution available
-func (c *Client) SetChallengeProvider(challenge Challenge, p ChallengeProvider) error {
+func (c *Client) SetChallengeProvider(challenge ChallengeType, p ChallengeProvider) error {
 	switch challenge {
 	case HTTP01:
-		c.solvers[challenge] = &httpChallenge{jws: c.jws, validate: validate, provider: p}
+		c.solvers[challenge] = &httpChallenge{jws: c.jws, vf: vf, provider: p}
```
mholt: I'd prefer more explicit field name like validateFunc.

acme/client.go
```diff
@@ -209,6 +213,33 @@ func (c *Client) AgreeToTOS() error {
 // your issued certificate as a bundle.
 // This function will never return a partial certificate. If one domain in the list fails,
 // the whole certificate will fail.
+func (c *Client) ObtainAuthorization(domains []string) (map[string]Challenge, map[string]error) {
```
mholt: Is it just me reading this wrong or is the godoc still for ObtainCertificate?

acme/dns_challenge.go
```diff
@@ -87,11 +76,35 @@ func (s *dnsChallenge) Solve(chlng challenge, domain string) error {
 		return err
 	}
 
-	return s.validate(s.jws, domain, chlng.URI, challenge{Resource: "challenge", Type: chlng.Type, Token: chlng.Token, KeyAuthorization: keyAuth})
+	return s.Validate(s.jws, domain, chlng.URI, Challenge{Resource: "challenge", Type: chlng.Type, Token: chlng.Token, KeyAuthorization: keyAuth})
+}
+
+func (s *dnsChallenge) Present(chlng Challenge, domain string) (string, error) {
```
mholt: Not really a fan of returning two values here. Why can't return just an error?

mholt:
Can you explain specifically why these changes are needed to accomplish what you want? Especially explain the changes that affect the external API of the package.

codefx9:
@mholt , Thanks for taking a look into this pull request. I am essentially building a webui that will display to users the dns changes they need to make. So, I am thinking about:

1. Writing a custom DNS01 challengeProvider, whose `Present` method writes  a db and later used to build an ui. Since the db schema is not generic, I have not included it. It just follows the `DNSProviderManual`.

2. Call `client.ObtainAuthorization()` to store the dns challenges into the db.

3. A cron-like process will periodically call CheckDNSPropagation() using information from  `Challenge struct`. If the dns is propagated, then it will call the new Validate() method.

4. When all the challenges are validated, it will call `client.RequestCertificate()`.

I have left the Solve() method as-is to avoid breaking the existing code. But I think it can be removed by moving remaining login as part the Validate method. I am not sure that makes sense.

acme/dns_challenge.go
codefx9: If I change `Present` to  `Present(chlng *Challenge, domain string)`, then I can set keyAuth directly in it. Then the return can be just error. Should I do that?

acme/challenges.go
codefx9: Fixed.

acme/client.go
codefx9: Fixed. I had not changed it before.

acme/client.go
```diff
@@ -354,6 +385,33 @@ func (c *Client) RenewCertificate(cert CertificateResource, bundle bool) (Certif
 }
 
 // Looks through the challenge combinations to find a solvable match.
+// Then presents the challenges in series and returns.
+func (c *Client) presentChallenges(challenges []authorizationResource) (map[string]Challenge, map[string]error) {
```
codefx9: @mholt, is there a way I can just return a single map here? Challenge has a `RemoteError` but no `error`.

janeczku:
@codefx9 I am pretty sure this PR is not necessary to accomplish your task: 
1. The functions required to build the DNS change strings for the UI are already exposed: `DNS01Record()`, `FindZoneByFqdn()`.
2. You can write your own provider based on DNS-manual whose Present() is a no-op.
3. checkDNSPropagation() is currently not exposed, but that can be changed.

xenolf:
I agree with @janeczku. Many of these changes do seem to be unnecessary.
The `Present()` hook of a provider has no timeout in itself so it could easily be used to display it in the UI.

codefx9:
@xenolf and @janeczku, thanks for your ideas. I am going to close this pull request once I get  a chance to try out your suggestions.

codefx9:
@janeczku/@xenolf, I have looked into the code and I have  a few more questions:
1. The solve() method in dns_challenge seems to checking for dns propagation without returning. That seems inadequate in  a server process. What if my server dies and I need to start checking dns propagation again on server restart?

2. I am guessing that I need to store `challenges` so that I can validate in case of server restart. challenge class needs to be public in that case. (which conflicts with the Challenge enum).

3. As you mentioned the checkDNSPropagation() needs to be public,

4. I created a ObtainAuthorization() method similarly to handle restart scenarios.

I thinking that the "library" should have independent methods that does one thing at a time. So, in a server process, we can do our own co-ordination. Then, the library can provide some `facades` that co-ordinates these interactions into a simple interface appropriate for clis.

I am fairly new to Go and very new to this codebase. So, it is very much possible that i misunderstood some parts of the code. Looking forward to your feedback/suggestions.

codefx9:
Hi folks, I have removed unnecessary changes. PTAL.

lenovouser:
What does this do besides renaming stuff? I don't get the point of this PR now.

LukeHandle:
@lenovouser as @codefx9 explained in his previous reply, he is now exporting the functions (so they can be called from external packages using the lego library).

The specifics and whether this is required I'll leave to more capable hands.

codefx9:
@janeczku / @xenolf , I have one more issue that I am not sure how to handle. In my background worker  process ( a gearman worker function), I call `CheckDNSPropagation()`. But if the propagation is successful, the only option I have is to call `ObtainCertificate()` again. I see 2 problems with it:

1. If I have multiple certificate in the authorization phase, for which one should I call `ObtainCertificate()` again?
2. Since the Present() method does not get the IdChallenge struct, I can't really call Validate() or RequestCertificate().

Do you have any suggestion on how to handle this?

Thank again for bearing with my native questions.

xenolf:
> The solve() method in dns_challenge seems to checking for dns propagation without returning. That seems inadequate in a server process. What if my server dies and I need to start checking dns propagation again on server restart?

The part where it does the propagation checking is wrapped into a timeout. The code will by default check every 2 seconds for 60 seconds and after that it will abort with an error. This default can be overridden by the provider using the `ChallengeProviderTimeout` interface.
The last line (the call to the `validate` function) does not return until we receive the final status update from the CA.

> I am guessing that I need to store challenges so that I can validate in case of server restart. challenge class needs to be public in that case. (which conflicts with the Challenge enum).

That depends on where in the process you want to get involved. The `authorization` type is what we get from the CA and which is used to pick a challenge we want to solve. If you only want to save a certain challenge (DNS-01) in your case, then yes you'd want to save the `challenge`.

> I thinking that the "library" should have independent methods that does one thing at a time. So, in a server process, we can do our own co-ordination. Then, the library can provide some facades that co-ordinates these interactions into a simple interface appropriate for clis.

I agree and I think we pretty much do this although we don't export these functions as to this date nobody needed them exported :) 
I'm fine with exporting more low level parts of this library as I can think of multiple use cases where this might be favorable. What we need to make sure though is to document these functions as being low level so people using this library can easily see which functions they want to use.

xenolf:
>  But if the propagation is successful, the only option I have is to call ObtainCertificate() again.

You are missing an integral step. You need to call the `validate` function or your own implementation of it. Otherwise the challenge is never submitted to the ACME server for checking.
`CheckDNSPropagation()` is only used by this library to make it's best effort to check if the validation would succeed.

codefx9:
> If you only want to save a certain challenge (DNS-01) in your case, then yes you'd want to save the challenge.
@xenolf, How should I go about that? Pass it to Present() method and store in my custom provider?

acme/client.go
```diff
@@ -42,10 +42,10 @@ type User interface {
 
 // Interface for all challenge solvers to implement.
 type solver interface {
-	Solve(challenge challenge, domain string) error
+	Solve(challenge IdChallenge, domain string) error
```
mholt: "ID" should be capitalized, as it is an initialism: https://github.com/golang/lint/blob/39d15d55e9777df34cdffde4f406ab27fd2e60c0/lint.go#L706 - otherwise lint is going to throw warnings.

codefx9: Fixed

mholt:
> What we need to make sure though is to document these functions as being low level so people using this library can easily see which functions they want to use.

If you go this route, this is very important :+1:  - do not forget it :wink:

xenolf:
> How should I go about that? Pass it to Present() method and store in my custom provider? I am only interested in DNS challenges.

You would not call `Present()` at all but implement your own provider as you are saying the default implementation of the DNS-01 provider is not suited for your use case.

Basically you'd get the challenges, choose the one you want to solve, save it somewhere permanent. Then you'd present the challenge to your user through your web UI and have him configure his DNS. After he told you he did that, you start checking the DNS propagation in the background. Once the DNS propagated, you call validate and wait for the ACME server to tell you the outcome. If the outcome was positive you are clear to request the certificate.

codefx9:
@xenolf , thanks for your detailed answer. I was thinking something along the same lines. I was hoping that I don't have to implement my own provider to do so. I don't see anything reusable to be add to this library with custom provider approach. I am going to close this pull request. Thanks again for answering all my questions.

PR #204, opened 2016-04-27
Fixes #199

### LINKED ISSUE #199: Dyn Provider returns 404 for * subdomains
I use a two A records for my domain. One is the standard TLD and another is a ' * ' to catch any subdomain. I then handle the subdomain inside of nginx. 

When trying to generate a certificate for www -- or any other subdomain -- which does not have an explicit record, I receive a 404 HTTP status code. 

I'm hoping the code can be written so that it validates either the named subdomain OR a ' * ' record to accommodate my situation.

aebruno:
Dyn provider was using incorrect zone in Present and Cleanup functions.
This uses FindZoneByFqdn() to find the correct zone for use in the
subsequent API calls.

### DIFF
--- a/providers/dns/dyn/dyn.go
+++ b/providers/dns/dyn/dyn.go
@@ -172,12 +172,17 @@ func (d *DNSProvider) logout() error {
 
 // Present creates a TXT record using the specified parameters
 func (d *DNSProvider) Present(domain, token, keyAuth string) error {
-	err := d.login()
+	fqdn, value, ttl := acme.DNS01Record(domain, keyAuth)
+
+	authZone, err := acme.FindZoneByFqdn(fqdn, acme.RecursiveNameservers)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


LukeHandle:
Failure appears unrelated:

```
=== RUN   TestCheckAuthoritativeNss

--- FAIL: TestCheckAuthoritativeNss (4.11s)
	dns_challenge_test.go:149: 8.8.8.8.asn.routeviews.org.: got false; want true
=== RUN   TestCheckAuthoritativeNssErr
--- FAIL: TestCheckAuthoritativeNssErr (4.02s)
	dns_challenge_test.go:161: #8.8.8.8.asn.routeviews.org.: expected "did not return the expected TXT record" (error); got "read udp 172.17.0.225:34150->128.223.51.29:53: i/o timeout"
```

Changes look good though 👍

PR #209, opened 2016-05-14
Report error from command line parser to user.

slon:
When I made a typo in command line, e.g. `--emial` instead of `--email`. Lego printed general `Incorrect Usage.` message. I expected to see something more useful.

This patch report error from command line parser to log. In case above:
 `2016/05/14 18:19:20 flag provided but not defined: -emal
`

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -141,7 +141,10 @@ func main() {
 		},
 	}
 
-	app.Run(os.Args)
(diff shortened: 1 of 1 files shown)


xenolf:
Thank you for this. I missed to implement it when I did the adjustments for the new CLI action signatures.

PR #213, opened 2016-05-16
Allow setting "--path" even if os.Getwd() fails

ByStones:
If you run the lego binary with sudo in a directory the user has no access to os.Getwd() will fail with a very missleading error message. It is currently not possible to use --path to set the working directory because the os.Getwd() check fails before you get the chance. This pull request should fix that.

How to reproduce:
`
sudo -i

cd /root

sudo -u nobody /tmp/lego --path="/tmp"
`

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -41,11 +41,18 @@ func main() {
 
 	acme.UserAgent = "lego/" + app.Version
 
+	defaultPath := ""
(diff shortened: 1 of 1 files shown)


xenolf:
Thanks for catching this :D

Issue #214, opened 2016-05-25
Add dns-timeout support


zealic:
I am trying to use DNS challenge with Route53 from cli run command.

But a great chance to failed with following error:

```log
2016/05/25 02:32:58 [www.example.com] Could not obtain certificates
	Error presenting token: Failed to determine Route 53 hosted zone ID: read udp 10.0.2.15:44956->8.8.4.4:53: i/o timeout
2016/05/25 02:32:58 [example.com] Could not obtain certificates
	Error presenting token: Failed to determine Route 53 hosted zone ID: read udp 10.0.2.15:36664->8.8.4.4:53: i/o timeout
```

Because of network latency and the [GFW](https://en.wikipedia.org/wiki/Great_Firewall) reason, dns request is very slow and unstable.
So we need apply dns timeout support for dns-challenges.

xenolf:
Thank you for bringing this up and fixing it right away! I hope this improves things for other people as well.
Closed via #215

PR #215, opened 2016-05-25
Add dns-timeout support.

zealic:
For issue #214


xenolf:
Thanks for taking the time to implement this :)

PR #222, opened 2016-06-06
In case of conflict during registration, the old registration is now recovered

dereulenspiegel:
If a client tries to register and encounters a 409 (Conflict) response the client now uses the old registration provided in the Location header of the 409 response. This makes the registration a bit more robust.

### DIFF
--- a/acme/client.go
+++ b/acme/client.go
@@ -167,15 +167,29 @@ func (c *Client) Register() (*RegistrationResource, error) {
 	}
 
 	var serverReg Registration
+	var regURI string
 	hdr, err := postJSON(c.jws, c.directory.NewRegURL, regMsg, &serverReg)
 	if err != nil {
-		return nil, err
+		remoteErr, ok := err.(RemoteError)
+		if ok && remoteErr.StatusCode == 409 {
+			regURI = hdr.Get("Location")
+			regMsg.Resource = "reg"
+			if hdr, err = postJSON(c.jws, regURI, regMsg, &serverReg); err != nil {
+				return nil, err
+			}
+		} else {
+			return nil, err
+		}
 	}
 
 	reg := &RegistrationResource{Body: serverReg}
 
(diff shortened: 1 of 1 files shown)


acme/client.go
```diff
@@ -167,15 +167,29 @@ func (c *Client) Register() (*RegistrationResource, error) {
...
 	var serverReg Registration
+	var regURI string
 	hdr, err := postJSON(c.jws, c.directory.NewRegURL, regMsg, &serverReg)
 	if err != nil {
-		return nil, err
+		remoteErr, ok := err.(RemoteError)
+		if ok && remoteErr.StatusCode == 409 {
+			regURI = hdr.Get("Location")
+			regMsg.Resource = "reg"
+			if hdr, err = postJSON(c.jws, regURI, regMsg, &serverReg); err != nil {
```
xenolf: Is this post to update the existing account with (possibly) new information?

xenolf:
Thank you for taking the time to implement this!
I have added a note inline.

acme/client.go
dereulenspiegel: Nope, at least to my knowledge this is just to recover existing registration information as stated in the [ACME protocol specification](https://letsencrypt.github.io/acme-spec/#registration). 
In case you use a new account key a new registration should be created and we don't get a 409 back.

dereulenspiegel:
I hope that I answered your question to your satisfaction. If there is anything else blocking the merge (formatting, coding style etc.) just give me a hint and I will fix it.

acme/client.go
xenolf: In that case the POST should only contain `{“resource”:”reg”}` as per the last paragraph in the part of the ACME spec you linked to.

dereulenspiegel: I think you're right, I updated the PR accordingly.

xenolf:
Great! Thank you :)

PR #223, opened 2016-06-07
reg: Add Query and Delete functions

vancluever:
This PR adds 2 new functions for registration support in `acme.Client`:

 * `QueryRegistration`: This performs a POST on the client
   registration's URI and gets the updated registration info.
 * `DeleteRegistration`: This deletes the registration as currently
   configured in the client.

The latter, while a part of the IETF draft, may not be 100% functional in LE yet, my tests showed that resources were still available after deletion. If anyone knows for sure would be nice to know :)

### DIFF
--- a/acme/client.go
+++ b/acme/client.go
@@ -189,6 +189,68 @@ func (c *Client) Register() (*RegistrationResource, error) {
 	return reg, nil
 }
 
+// DeleteRegistration deletes the client's user registration from the ACME
+// server.
+func (c *Client) DeleteRegistration() error {
+	if c == nil || c.user == nil {
+		return errors.New("acme: cannot unregister a nil client or user")
+	}
+	logf("[INFO] acme: Deleting account for %s", c.user.GetEmail())
+
+	regMsg := registrationMessage{
+		Resource: "reg",
+		Delete:   true,
+	}
+
+	_, err := postJSON(c.jws, c.user.GetRegistration().URI, regMsg, nil)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// QueryRegistration runs a POST request on the client's registration and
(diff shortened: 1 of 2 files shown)


acme/client.go
```diff
@@ -189,6 +189,67 @@ func (c *Client) Register() (*RegistrationResource, error) {
...
+	// existing URI
+	reg.URI = c.user.GetRegistration().URI
+	if links["terms-of-service"] != "" {
+		reg.TosURL = links["terms-of-service"]
+	}
+
+	if links["next"] != "" {
+		reg.NewAuthzURL = links["next"]
+	} else {
+		return nil, errors.New("acme: The server did not return 'next' link to proceed")
```
xenolf: This should probably be a different error message here to not be confused with the one in registration.

acme/client.go
```diff
@@ -189,6 +189,67 @@ func (c *Client) Register() (*RegistrationResource, error) {
...
+	logf("[INFO] acme: Deleting account for %s", c.user.GetEmail())
+
+	regMsg := registrationMessage{
+		Resource: "reg",
+		Delete:   true,
+	}
+
+	_, err := postJSON(c.jws, c.user.GetRegistration().URI, regMsg, nil)
+	if err != nil {
+		return nil
```
xenolf: Probably meant to return err here, not nil.
Also it could maybe make sense to check the HTTP status code returned by the server in case of an error.

xenolf:
Thank you for this pull request.
I've added a few comments inline.

vancluever:
Hey @xenolf great thanks for the review! I will update this code and re-push ASAP.

vancluever:
@xenolf all updated and ready for follow-up review!

PR #224, opened 2016-06-08
Support existing CSRs (update to #122)

vancluever:
This builds on #122, implementing the requested renewal support, by extending the `acme.CertificateResource` struct to include the CSR, and adding a hook in `RenewCertificate` that checks for the CSR before attempting to re-generate from scratch.

Will remove the WIP tag once I have proof of functionality.


vancluever:
I want to adjust the `ObtainCertificateForCSR()` API a bit so that it takes an `x509.CertificateRequest` object versus a DER block - this would make it consistent with `ObtainCertificate()` taking a `crypto.PrivateKey` and make it clear to developers that they need to handle decoding the certificate from whatever format it's in before sending it in.

vancluever:
WIP removed, ready for review. Tested this via API, new cert and renewal.

acme/client.go
```diff
@@ -200,6 +200,66 @@ func (c *Client) AgreeToTOS() error {
...
+// the whole certificate will fail.
+func (c *Client) ObtainCertificateForCSR(csr x509.CertificateRequest, bundle bool) (CertificateResource, map[string]error) {
+	// figure out what domains it concerns
+	// start with the common name
+	domains := []string{csr.Subject.CommonName}
+
+	// loop over the SubjectAltName DNS names
+DNSNames:
+	for _, sanName := range csr.DNSNames {
+		// /
```
xenolf: Empty comment?

cli_handlers.go
```diff
@@ -232,11 +268,35 @@ func run(c *cli.Context) error {
...
+			// we couldn't read the CSR
+			failures = map[string]error{"csr": err}
+		} else {
+			// obtain a certificate for this CSR
+			cert, failures = client.ObtainCertificateForCSR(*csr, true)
+		}
 	}
 
-	cert, failures := client.ObtainCertificate(c.GlobalStringSlice("domains"), !c.Bool("no-bundle"), nil)
+	cert, failures = client.ObtainCertificate(c.GlobalStringSlice("domains"), !c.Bool("no-bundle"), nil)
```
xenolf: This should get removed, we don't want to obtain certs twice.
Also both calls to `ObtainCertificate(ForCSR)` should use the `no-bundle` boolean as this line does.

acme/client.go
```diff
@@ -326,6 +386,18 @@ func (c *Client) RenewCertificate(cert CertificateResource, bundle bool) (Certif
 		return cert, nil
 	}
 
+	// If the certificate is the same, than we need to request a new certificate.
```
xenolf: then we need

acme/crypto.go
```diff
@@ -258,9 +261,22 @@ func pemDecodeTox509(pem []byte) (*x509.Certificate, error) {
 		return nil, err
 	}
 
+	if pemBlock.Type != "CERTIFICATE REQUEST" {
```
xenolf: This block is in the wrong function, isn't it? Should be in `pemDecodeTox509CSR`.

xenolf:
Thanks for the updated PR!
I went through it and added a few annotations.

This does only handle requesting a cert via an existing CSR in the CLI, right?

vancluever:
@xenolf thanks for the review - annotations addressed, let me know if there's anything else!

vancluever:
@xenolf I forgot to mention, that yeah to my understanding that this is the only real change here on the CLI side. API is mainly what I have been working with though. There's no breaking changes to existing API and all I've really added to what was already in #122 is solving the renewal issue so we could get it merged.

I've made my contributions as part of my efforts on getting ACME support into Terraform (you can check the progress at hashicorp/terraform#7058), which is nearing completion, I'm looking forward to this being merged so that I have a clean package to vendor!

vancluever:
Thanks for the merge @xenolf!

Issue #225, opened 2016-06-09
undefined: acme.TLSSNI01ChallengeCertDomain


sn0bored:
I am working on a project using https://github.com/rsc/letsencrypt and when I run

`go get rsc.io/letsencrypt` 

The output I get is:
`# rsc.io/letsencrypt`
`../../../rsc.io/letsencrypt/lets.go:703: undefined: acme.TLSSNI01ChallengeCertDomain`
`../../../rsc.io/letsencrypt/lets.go:716: undefined: acme.TLSSNI01ChallengeCertDomain`

In an attempt to isolate the issue I created a project using only the readme and get the same result.

All my research and work has led me to the interaction with this package.  Is this a dependency that may have changed?

Thanks

xenolf:
Hey and thanks for submitting the issue :)

What version of go do you have installed? The package rsc/letsencrypt uses vendoring to pin it's dependencies to a certain version and it seems like your installation pulled the latest source from here and used that instead of the vendored version.
The latest master in this repository had the symbol you are referring to renamed.

sn0bored:
Wow thanks for the quick response

`go version go1.5.1 darwin/amd64`

xenolf:
IIRC for go 1.5 you need to set the environment variable `GO15VENDOREXPERIMENT=1` before running the go command. As of go 1.6 you no longer need to do that.

sn0bored:
You rock!  Updated to 1.6.2 and that fixed it.

Danke!

xenolf:
Glad you got it working! :)

PR #227, opened 2016-06-10
Change TLS SNI Challenge Cert function to return domain

dmcgowan:
Used by rsc.io/letsencrypt to get the challenge domain. Originally committed under rsc.io/letsencrypt/vendor and maintained original authorship as from https://github.com/rsc/letsencrypt/commit/2ba1dda8fd06c47af5847588616112a9bf93c346.

Addresses issues with vendoring rsc.io/letsencrypt, especially with godep, see
- #225 
- https://github.com/rsc/letsencrypt/issues/9
- https://github.com/docker/distribution/pull/1779

### DIFF
--- a/acme/tls_sni_challenge.go
+++ b/acme/tls_sni_challenge.go
@@ -41,7 +41,7 @@ func (t *tlsSNIChallenge) Solve(chlng challenge, domain string) error {
 }
 
 // TLSSNI01ChallengeCert returns a certificate and target domain for the `tls-sni-01` challenge
-func TLSSNI01ChallengeCertDomain(keyAuth string) (tls.Certificate, string, error) {
+func TLSSNI01ChallengeCert(keyAuth string) (tls.Certificate, string, error) {
 	// generate a new RSA key for the certificates
 	tempPrivKey, err := generatePrivateKey(RSA2048)
 	if err != nil {
@@ -64,10 +64,4 @@ func TLSSNI01ChallengeCertDomain(keyAuth string) (tls.Certificate, string, error
 	}
 
 	return certificate, domain, nil
-}
-
-// TLSSNI01ChallengeCert returns a certificate for the `tls-sni-01` challenge
-func TLSSNI01ChallengeCert(keyAuth string) (tls.Certificate, error) {
-	cert, _, err := TLSSNI01ChallengeCertDomain(keyAuth)
-	return cert, err
 }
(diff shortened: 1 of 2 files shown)


acme/tls_sni_challenge.go
```diff
@@ -40,12 +40,12 @@ func (t *tlsSNIChallenge) Solve(chlng challenge, domain string) error {
 	return t.validate(t.jws, domain, chlng.URI, challenge{Resource: "challenge", Type: chlng.Type, Token: chlng.Token, KeyAuthorization: keyAuth})
 }
 
-// TLSSNI01ChallengeCert returns a certificate for the `tls-sni-01` challenge
-func TLSSNI01ChallengeCert(keyAuth string) (tls.Certificate, error) {
+// TLSSNI01ChallengeCert returns a certificate and target domain for the `tls-sni-01` challenge
+func TLSSNI01ChallengeCertDomain(keyAuth string) (tls.Certificate, string, error) {
```
mholt: I don't think we need to change the name of the function here.

acme/tls_sni_challenge.go
```diff
@@ -55,13 +55,19 @@ func TLSSNI01ChallengeCert(keyAuth string) (tls.Certificate, error) {
...
 	}
 
-	return certificate, nil
+	return certificate, domain, nil
+}
+
+// TLSSNI01ChallengeCert returns a certificate for the `tls-sni-01` challenge
+func TLSSNI01ChallengeCert(keyAuth string) (tls.Certificate, error) {
+	cert, _, err := TLSSNI01ChallengeCertDomain(keyAuth)
+	return cert, err
```
mholt: I don't think this function adds any value. There's no extra computation saved by not returning the domain; might as well just have one function.

acme/tls_sni_challenge.go
dmcgowan: To be clear, you are suggesting just updating the original function to now return the domain? I am OK doing that but was unsure the impact it would have on importers since it is already an exported function.

mholt: Yeah, that's what I think would be good. I can see the value in it; it's up
to @xenolf what he wants to do about compatibility. Personally I would
rather break it now than later.

xenolf: I'd rather deprecate it for one release. but seeing that go has no way to do that I agree with @mholt.

dmcgowan:
@mholt updated with your suggestion

mholt:
I like this now. :+1:

Issue #229, opened 2016-06-12
account.json has executable bit set


doherty:
This file should be chmod 600, not 700.

PR #230, opened 2016-06-12
Write account.json without executable bit set

doherty:
Fixes gh-229


xenolf:
Thanks for catching and fixing it! :)

PR #231, opened 2016-06-15
providers/dns/route53: Adjust DNS challenge TTL to 10 seconds

vancluever:
While more than likely never to come up in a real-world situation, during renewal integration testing a value of 120 seconds has proven to be too high (the old challenge record has not expired by the time the new one is created).

### DIFF
--- a/providers/dns/route53/route53.go
+++ b/providers/dns/route53/route53.go
@@ -18,6 +18,7 @@ import (
 
 const (
 	maxRetries = 5
+	route53TTL = 10
 )
 
 // DNSProvider implements the acme.ChallengeProvider interface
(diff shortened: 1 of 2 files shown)


xenolf:
I'm not sure if changing the default is the right thing to do there as I'm pretty sure that many of our supported DNS providers don't even support TTL values that low.
Which DNS provider are you using and getting this problem with?

vancluever:
Hey @xenolf, I've been doing the testing in the R53 provider.

Fair enough on the provider side, looks like [DNSSimple only supports a minimum of 1 minute](https://support.dnsimple.com/articles/what-minimum-time-to-live/), for example. Route 53 recommends 60 seconds "or less" for health check records - [docs link](http://docs.aws.amazon.com/Route53/latest/APIReference/API_ChangeResourceRecordSets_Requests.html).

I'll check with their support and see what the actual minimum is.

vancluever:
Looks like the Route 53 recommended minimum is 10 seconds (it can go all the way to zero, but they don't necessarily recommend it. I'll fix this up so it's just the Route 53 provider.

vancluever:
@xenolf - now done with an integration test as well. :)

xenolf:
This is looking good, thank you!
I'm sorry I fell behind on this... I had a pretty rough time :)

vancluever:
No worries @xenolf and thanks for the merge!

PR #234, opened 2016-06-19
Add optional support for .pem output (.crt + .key)

jboelter:
Using this with the default configuration of lighthttpd on an EdgeOS router from Ubiquiti that expects a .pem.  Adding a --pem option saves a few lines of script and directly generates the concatenated crt + key.

I could move the --pem option to under the run and renew commands.  Wasn't clear if there's a preference here.


xenolf:
Thank you for taking the time to implement this! Much appreciated.

PR #237, opened 2016-06-28
parallelize solving challenges

jmhodges:
This change parallelizes the solving of challenges and greatly speeds it up. My larger certificates completed around an order of magnitude faster.

Tested with the Google Cloud DNS provider. There may be other races that need to be fixed.

The DNS timeout increase just made the DNS challenges more reliable and since most Let's Encrypt certs have 2 or more domains in them, it shouldn't slow people down.

The global fqdnToZone could probably stand to be refactored into a real locked object.

The jws struct needed some refactoring to make its use of its nonce slice thread-safe.

### DIFF
--- a/acme/client.go
+++ b/acme/client.go
@@ -507,22 +507,36 @@ func (c *Client) RenewCertificate(cert CertificateResource, bundle bool) (Certif
 func (c *Client) solveChallenges(challenges []authorizationResource) map[string]error {
 	// loop through the resources, basically through the domains.
 	failures := make(map[string]error)
+	ch := make(chan domainError, len(challenges))
 	for _, authz := range challenges {
-		// no solvers - no solving
-		if solvers := c.chooseSolvers(authz.Body, authz.Domain); solvers != nil {
-			for i, solver := range solvers {
-				// TODO: do not immediately fail if one domain fails to validate.
-				err := solver.Solve(authz.Body.Challenges[i], authz.Domain)
-				if err != nil {
-					failures[authz.Domain] = err
-				}
-			}
-		} else {
-			failures[authz.Domain] = fmt.Errorf("[%s] acme: Could not determine solvers", authz.Domain)
+		go func(authz authorizationResource) {
+			err := c.solveChallenge(authz)
+			ch <- domainError{Domain: authz.Domain, Error: err}
+		}(authz)
+	}
+	for i := 0; i < len(challenges); i++ {
+		de := <-ch
+		if de.Error != nil {
+			failures[de.Domain] = de.Error
 		}
 	}
-
 	return failures
+}
+
+func (c *Client) solveChallenge(authz authorizationResource) error {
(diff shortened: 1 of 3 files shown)


jmhodges:
I noticed that lego has never been closing the bodies of the httpPost responses. Since I'm all over the relevant code already, I've included a fix for those calls in this change as well.

jmhodges:
Oops, I take that last commit back. The httpPost was fine (it returned the response) and the other was an httpHead. My bad.

jmhodges:
One minor complication: the manual DNS provider requires you to hit enter for each domain. With this patch, it could be a little harder to know how many times to hit enter.

One option: use this patch as-is and be okay with that
Second option: push the parallelization into the providers by making them take more than one domain at a time. This has the benefit of allowing providers that can batch operations to do so.

I'm open to hearing other options.

xenolf:
Hey there! This is awesome, thank you!

I'd argue that it's better to push it into the providers as they know better how to handle it versus the challenge. Plus this would allow us to have the other challenges handle that themselves as well.

mholt:
Although I can see the value in allowing providers to batch operations, I also like the simplicity and consistency of this approach: "it just works."

As for the manual DNS challenge, parallelism won't really help here anyway, so how about we simply don't perform the challenges in parallel if there is a manual DNS solver that is used?

jmhodges:
To special case the manual DNS provider ("solver" in some of this code), we would need to move the chooseSolvers call up to its own for-loop a head of the current one, figure out if any of the solvers that could be used are the manual DNS solver, and then de-parallelize when we loop back through the authzs if so.

On the other hand, what backwards compatibility constraints are we under for the solvers? Can Present be made to take take a slice of structs representing what it currently takes or is that too much of an API break?

jmhodges:
Poking around a bit more: for the manual DNS challenge check, that means we'd also have to change ObtainCertificateForCSR and ObtainCertificate because solveChallenge's input would have to include a slice of solvers.

jasonwbarnett:
fixes #173

lenovouser:
Any chance of getting this merged?

xi2:
Hi. One thing I feel might be wise is to make the parallel solving logic something that is not initially applied to all dns providers. It could maybe be controlled by consulting a map after obtaining the provider name through a type check. Then, as the author/maintainer of each dns provider has a chance to check things work, they can "flip the switch" and turn it on through a pull request.

I know that the gandi provider (that I authored) may require some very minor changes, but it depends on how parallel solving is eventually implemented. I assume the same may be true for some of the other dns providers.

ldez:
Since 2016, the code has changed a lot :wink: 

The parallelization of challenges (concurrent access to the DNS API) produces a major problem with wildcards or subdomains: the DNS will have to modify the same domain which will create an inconsistent entry (one of the entries can be overwritten by another).

For the majority of the DNS providers (by example GCloud) it will not work.

The solution #607 is more suitable for reducing the time to obtain certificates.

Thanks for the contribution.

PR #238, opened 2016-06-28
Add DNSProvider for DNSMadeEasy (http://dnsmadeeasy.com/)

moomerman:
👋 Here's a DNSProvider implementation for DNSMadeEasy (http://dnsmadeeasy.com/).


xenolf:
Hey there, I'm sorry it took me so long to respond!
This is looking good though, especially props for only using the stdlib!

PR #243, opened 2016-07-10
Support setting DNS parameters in a file with godotenv

mmonaco:
This uses godotenv.Load() which means that environment variables already
set take precedent, as opposed to godotenv.Overload(). No need for
errors to be fatal, so log and move on.


voltagex:
Ping @xenolf - this would be especially helpful when running `lego` from cron or similar.

@mmonaco, does this require any documentation changes?

ldez:
Closed in favor of #535

PR #245, opened 2016-07-13
codegangsta is now called urfave.

bsiegert:
Update import paths for cli package.


xenolf:
Thank you! Sorry it took me so long to check it out :/

PR #251, opened 2016-07-21
Filter Google Cloud DNS List() calls

bshi:
Make a more efficient API call to only fetch the name of interest.


xenolf:
Thank you for this! :)

PR #254, opened 2016-07-29
Correctly determine the zone for CNAME domains pointing to another zone

janeczku:
Fix for https://github.com/xenolf/lego/issues/248


xenolf:
Thank you for the fix! ;)

PR #261, opened 2016-08-10
add an alternative parameter "zones" while NewDNSProvider (googlecloud)

lbeeon:
Original google cloud dns provider use the input domain as the zone while dns challenge, and it will cause some issues. See below.

Ex. 1 (Success)
 challenge domain: test.com
 google zone: test.com

Ex. 2 (Failed)
 challenge domain: www.test.com
 google zone (X:  www.test.com 
 google zone (O:  test.com  

To solve this, I provide a parameter zones, while new provider. ("NewDNSProvider(zones map[string]string)".
By this to maintain the relation between challenge domains and zones. 

> Original (error: No matching GoogleCloud domain found for domain www.test.com 

`p, _ := NewDNSProvider()`
`certificates, failures := client.ObtainCertificate([]string{"test.com", "www.test.com"} , true, nil)`

> After ( if there is no mapping zone provided, use the domain as default zone

`zones := map[string]string{ "www.test.com": "test.com"}`
`p, _ := NewDNSProvider(zones)`
`certificates, failures := client.ObtainCertificate([]string{"test.com", "www.test.com"} , true, nil)`

### DIFF
--- a/providers/dns/googlecloud/googlecloud.go
+++ b/providers/dns/googlecloud/googlecloud.go
@@ -20,18 +20,19 @@ import (
 type DNSProvider struct {
 	project string
 	client  *dns.Service
+	zones   map[string]string
 }
 
 // NewDNSProvider returns a DNSProvider instance configured for Google Cloud
 // DNS. Credentials must be passed in the environment variable: GCE_PROJECT.
-func NewDNSProvider() (*DNSProvider, error) {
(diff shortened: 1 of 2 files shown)


xenolf:
Hello and thanks for the PR! I opted to take #262 over this as it does not need an additional flag and solves the same problem.

PR #262, opened 2016-08-11
Fix GoogleCloud DNS challenge to allow subdomains

### LINKED ISSUE #257: Proposal: Add --zone flag for DNS challenge
I ran into an issue trying to a cert with multiple SANs on Google using lego. The issue is when trying to get a cert for a subdomain(i.e. `foo.example.com`) when my hosted zone is `example.com`. This doesn't work (as least for Google DNS provider) due to the assumption that the `domain` being passed is also the `zone` being hosted: https://github.com/xenolf/lego/blob/master/providers/dns/googlecloud/googlecloud.go#L127

This is simple enough to fix when the hosted zone is just the `domain.tld` and I have a working fix here: https://github.com/doublerr/lego/commit/99b62912aee47ae917417b1caa0d54893948525b

**BUT** this would break for anyone that want a cert for a fqdn that is multiple sub domains deep (which is me), `a.foo.example.com`.

Since we can't reasonably assume what the hosted zone is, I'd like to propose adding a `--zone` flag for the DNS challenge. This would allow an user to do something like:

```bash
GCE_PROJECT=example-123 lego --dns gcloud --zone=example.com \
--domains=foo.exmaple.com \
--domains=a.foo.example.com \
--domains=bar.example.com \
--email=me@example.com --accept-tos run
```

Since this would change the interface I think we can use a reasonable default that doesn't really change current behavior. Some possible defaults:

1. Assume the zone is `domain.tld`of the supplied fqdn.
1. If 1 fqdn is supplied, assume the zone is that domain. ( This is how things are today)
1. If 2 fqdns are supplied, assume the zone the common portion.

I'm happy to work on this, if this seems like a feature you want.

doublerr:
Fixes #257

Previously the google cloud provider assumed the domain being provided
was also the authoritative zone. This fix uses an acme function to
recursively lookup the authoritative zone for a provided domain.

This solves the same problem as https://github.com/xenolf/lego/pull/261 but doesn't require the extra flag.

### DIFF
--- a/providers/dns/googlecloud/googlecloud.go
+++ b/providers/dns/googlecloud/googlecloud.go
@@ -5,7 +5,6 @@ package googlecloud
 import (
 	"fmt"
 	"os"
-	"strings"
 	"time"
 
 	"github.com/xenolf/lego/acme"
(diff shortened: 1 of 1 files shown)


xenolf:
LGTM
Thank you for your time @doublerr

PR #268, opened 2016-08-16
Skip challenges when authz is already valid

mholt:
Let me know what you think of this fix. It seems straightforward enough, but I may have missed something. This change was tested and I've confirmed it fixes the issue for me. Want to double-check it for correctness? I'd love to get this out for the next Caddy release, like, tomorrow. :smile:


mholt:
CI checks are passing for Go 1.6+ (vultr dependency requires it).

PR #271, opened 2016-08-18
Add sync.Mutex to lock and unlock j.nonces

katejefferson:
This PR adds sync.Mutex to jws, and locks/unlocks methods where j.nonce is edited. This fixes the [data race](https://golang.org/doc/articles/race_detector.html) during tests for ObtainCertificate.


xenolf:
Thank you for your time and for implementing this!

PR #280, opened 2016-08-29
Add support for Linode as a DNS provider.

timewasted:
As the subject says, I've added support for Linode as a DNS provider.  Confirmed it works on at least subdomains (www.example.com) by acquiring a certificate for one of my domains.  Bare domains (example.com) SHOULD also work as well, but are currently untested.

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -195,6 +195,7 @@ Here is an example bash command using the CloudFlare DNS provider:
 	fmt.Fprintln(w, "\tdnsmadeeasy:\tDNSMADEEASY_API_KEY, DNSMADEEASY_API_SECRET")
 	fmt.Fprintln(w, "\tgandi:\tGANDI_API_KEY")
 	fmt.Fprintln(w, "\tgcloud:\tGCE_PROJECT")
+	fmt.Fprintln(w, "\tlinode:\tLINODE_API_KEY")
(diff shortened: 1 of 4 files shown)


xenolf:
Hello there and thank you for providing this! I'm sorry it took me so long to get back to you but the last couple of weeks were really exhausting :)

Seeing that you are also the maintainer of the used package for the linode api interaction. Do you intend to maintan that further?

timewasted:
And I'm sorry it took me so long to get back to you!  I'm apparently bad at realizing that I have something that I should respond to.

I wrote the Linode API package pretty much for the sole purpose of integrating with Lego.  In order to do that, I obviously needed to implement parts of their DNS API.  It didn't make sense to implement ONLY the parts of the DNS API that are required for a DNS provider, so I decided to implement it all.  So to that extent, yes, I intend to maintain it.

As for adding in support for other aspects of the Linode API... that I'm not as sure about.  I'm not strictly opposed to it if:
A) there is a demand for it and
B) I actually have the time to do so.

Lately, B seems to be the harder condition to work around.

RobertDeRose:
@xenolf This would be cool if this was merged

RobertDeRose:
🎉 👍

PR #293, opened 2016-09-29
Dns from resolv.conf

stuart-warren:
if an /etc/resolv.conf file exists, then get the DNS servers from there

if not, uses the original defaults

handles IPv6 addresses in resolv.conf and adds port 53 if not specified

Todo:
- [x] add tests


xenolf:
Hey there!
Thank you for taking the time to implement this feature. 
The change looks fine but I think we should include a log line to tell the user which resolver is being used.
What are your thoughts about this?

stuart-warren:
Sure, any better?

Though as `RecursiveNameservers` is a public variable, it could be modified by something and then the wrong details will have been logged.

xenolf:
You are right. I guess L102 is a good spot for it. Just add the information to the existing log statement and use the value of `RecursiveNameservers`. As this is run right when we are about to run the queries we should get the right values.

xenolf:
@stuart-warren any update on this?

stuart-warren:
I assume this is what you meant?
```
2016/11/02 16:06:55 [INFO][example.com] Checking DNS record propagation using [10.248.8.249:53 10.248.8.250:53 10.249.0.63:53 10.252.15.11:53]
```

xenolf:
Exactly, thank you for the updated PR.

PR #295, opened 2016-10-09
Add NS1 DNS provider.

calavera:
Hi!

I'd like to be able to use this package with [NS1](https://nsone.com) as DNS provider.
I added integration with their REST API uning their official Go client: 

https://github.com/ns1/ns1-go

Please, let me know if there any anything else I need to do to see this Pull Request merged.

Thanks for putting all this code together! :raised_hands:

Signed-off-by: David Calavera <david.calavera@gmail.com>

### DIFF
--- a/cli_handlers.go
+++ b/cli_handlers.go
@@ -24,6 +24,7 @@ import (
 	"github.com/xenolf/lego/providers/dns/googlecloud"
 	"github.com/xenolf/lego/providers/dns/linode"
 	"github.com/xenolf/lego/providers/dns/namecheap"
+	"github.com/xenolf/lego/providers/dns/ns1"
 	"github.com/xenolf/lego/providers/dns/ovh"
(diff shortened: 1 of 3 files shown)


xenolf:
Hey there!

Thank you for your time and effort!
Would you be able to support this provider in case bugs or other problems arise with it?
Could you also add your provider to the CLI [here](https://github.com/xenolf/lego/blob/master/cli_handlers.go#L116) so other people are able to use your provider from the CLI?

Thank you!

calavera:
> Would you be able to support this provider in case bugs or other problems arise with it?

Yes, just ping me.

> Could you also add your provider to the CLI here so other people are able to use your provider from the CLI?

Absolutely. Do you have any documentation that needs to be changed to? I want to make sure I don't miss anything.

xenolf:
The changes you made are fine. :) Thank you!

PR #296, opened 2016-10-11
add memcached provider

clinta:
This adds a memcached provider for use with nginx's memcached module.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,10 @@ go:
 - 1.6.3
 - 1.7
 - tip
+services:
+  - memcached
+env:
+  - MEMCACHED_HOSTS=localhost:11211
 install:
 - go get -t ./...
 script:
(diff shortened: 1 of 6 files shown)


xenolf:
Hello there!

Thank you for providing this!
Would it be possible to add unit tests to the provider?

clinta:
Would it be okay to edit the travis config so that memcached is running in travis where the tests will run? Or is there an expectation that the tests can be run without any external dependencies? If so I'll have to find some good way to mock a memcached instance.

xenolf:
I don't have any objections against adding memcached to travis if that's possible.
The only requirement I have is that `go test ./...` does also function without the dependecy. Basically as most of the DNS providers are doing based on if an ENV variable is set or not. Doing it via an ENV variable is not required though, that's up to you.

clinta:
Rebased and added tests.

PR #299, opened 2016-10-15
Add AuroraDNS support

edeckers:
This commit adds support for Aurora DNS (https://www.pcextreme.com/aurora/dns)


xenolf:
Hello there!

Thank you for implementing this provider!
Before merging it in, may I ask you if you are able to continue to support it in the case of bugs or other problems?

edeckers:
Yes, I will support it, no problem.

thedanielgray:
This adds a dependency towards https://github.com/edeckers/auroradnsclient, which is GPL 3.0 licensed. Wouldn't this force the license of this repo (github.com/xenolf/lego) to also be GPL 3.0? (due to the virality of the GPL license).

edeckers:
You're right about that. Changing the licensing of my library to LGPLv3.0 would fix this issue, correct?

thedanielgray:
As far as I know, LGPL disallows static linking which is Go's normal way of packaging (maybe I'm wrong), so in that case, even LGPL would not work for all use cases.

edeckers:
@thedanielgray Seems you're right once again. I looked into a few other licenses and it seems MPL-2.0 has about everything I want, so I'll go with that, unless you (or someone else) tells me that this license, too, doesn't suffice.

thedanielgray:
From what I can see on the Mozilla website (https://www.mozilla.org/en-US/MPL/2.0/FAQ/) this seems to allow static linking. Thank you!

PR #311, opened 2016-11-02
Add Exoscale DNS provider

trecloux:
Add support for Exoscale DNS provider


xenolf:
Hello there!

Thank you for implementing this provider!
Before merging it in, may I ask you if you are able to continue to support it in the case of bugs or other problems?

trecloux:
@xenolf Sure, no problem to support it.
I can't guarantee to be very quick but not problem

PR #312, opened 2016-11-02
Add dns provider for dnspod

decker502:
Add dns provider for dnspod.
Dnspod provides domain name services in China.


decker502:
yes, I do.

PR #313, opened 2016-11-07
Extract from CLI the name -> DNS provider mapping

trecloux:
This avoids duplication of this mapping in lib clients 
Ex : https://github.com/containous/traefik/pull/738#issuecomment-258810469


tyrken:
Agreed - exposing this has got to make referencing code simpler.

trecloux:
@xenolf : I resolved the conflict.

No specific reason to use exoscale, I wanted to test one provider.
Do you want to extend the coverage for each provider ?

PR #317, opened 2016-11-14
Export DomainError and ChallengeError

fstehle:
in order to allow type assertion on specific errors, eg.:
```
certificate, failures := s.client.ObtainCertificate(domains, true, nil)
for domain, failure := range failures {
        switch f := failure.(type) {
        case acme.ChallengeError:
                // Handle acme.ChallengeError
        case acme.RemoteError:
                // Handle acme.RemoteError
        default:
                // Handle other errors
        }
}
```


fstehle:
Thanks a lot for reviewing the PR. However we don't need the error handling.

PR #326, opened 2016-12-01
Support reading GCE credential from service account file.

TamalSaha:


ManuelGysin:
I run today into trouble with this.  Your changes seems good, but we can keep it simple and just run:

`export GOOGLE_APPLICATION_CREDENTIALS=/conf/gcloud.json`

The rest is handled by https://godoc.org/golang.org/x/oauth2/google#FindDefaultCredentials
So there is no need to this within lego itself. Maybe adding some information about it in the README?

TamalSaha:
@ManuelGysin , we are using this library inside a Kubernetes pod and we read this data from a Kubernetes secret. So, we don't have the data on any file that we can set as ENV variable. So, we need an option to pass JWT key bytes to construct the client.

ManuelGysin:
@TamalSaha I also use kubernetes. Is there a reason not to mount the secret? I create the secrets with --from-file and just mounting it at a specific path and setting GOOGLE_APPLICATION_CREDENTIALS there. Just curious why you are reading the secret as byte.

TamalSaha:
@ManuelGysin I am using this library from a controllers that runs as static pods. Static pods can't access mount secrets. Also, I read the secrets from annotations applied to Ingress. So, the controller pods can't mount all secrets ahead of time.

TamalSaha:
@ManuelGysin I have changed the code to use GOOGLE_APPLICATION_CREDENTIALS env var to check for a file. So, this keeps the original behavior unchanged and does not introduce any new env var.

PR #331, opened 2016-12-11
Added multiple DNS provider support

CPlusPlus17:


acme/client.go
```diff
@@ -104,14 +104,17 @@ func (c *Client) SetChallengeProvider(challenge Challenge, p ChallengeProvider)
...
 		c.solvers[challenge] = &tlsSNIChallenge{jws: c.jws, validate: validate, provider: p}
-	case DNS01:
-		c.solvers[challenge] = &dnsChallenge{jws: c.jws, validate: validate, provider: p}
 	default:
 		return fmt.Errorf("Unknown challenge %v", challenge)
 	}
 	return nil
 }
 
+func (c *Client) SetChallengeProviderDNS(challenge Challenge, p []ChallengeProvider) error {
```
xenolf: Please add a comment to exported functions. Also use plural as we're possibly setting multiple providers?

acme/dns_challenge.go
```diff
@@ -69,13 +69,13 @@ func DNS01Record(domain, keyAuth string) (fqdn string, value string, ttl int) {
...
 	validate validateFunc
-	provider ChallengeProvider
+	providers []ChallengeProvider
 }
 
 func (s *dnsChallenge) Solve(chlng challenge, domain string) error {
 	logf("[INFO][%s] acme: Trying to solve DNS-01", domain)
 
-	if s.provider == nil {
+	if s.providers == nil {
```
xenolf: Don't check for nil here, a slice will not be nil but have a len() of zero.

acme/dns_challenge.go
```diff
@@ -85,14 +85,22 @@ func (s *dnsChallenge) Solve(chlng challenge, domain string) error {
 		return err
 	}
 
-	err = s.provider.Present(domain, chlng.Token, keyAuth)
-	if err != nil {
-		return fmt.Errorf("Error presenting token: %s", err)
+	for _, v := range s.providers {
+		err = v.Present(domain, chlng.Token, keyAuth)
```
xenolf: This would try to `Present()` every domain on every provider. Shouldn't we have a way to tell which provider a certain domain uses?

ManuelGysin: Ah I didn't thought about this use case. But normally you have one or more DNS provider per Domain. I designed the function that way. This means if you want to validate something.url.com, and you have two DNS providers, you can add both to set the right TXT record. With the old system, it was not possible to know which DNS provider will serve the answer and validation could be invalid. I'm not sure if we want to provide a mapping per domain and DNS provider. Actually I don't see such a use case.

xenolf: @ManuelGysin I see where you are coming from. But I also see the use case where people want to add domains to a SAN certificate which are hosted by different DNS providers.

ldez:
Closed in favor of #605.

Thanks for the contribution.

PR #336, opened 2016-12-27
Try to detect Zone from fqdn using PublicSuffix package.

### LINKED ISSUE #330: googlecloud.getHostedZone doesn't work correctly for CNAMEs that cross zones
In my Google cloud project, I have two domains and two zones:
* example.com, with google cloud dns zone `example-com-zone`
* example.net, with google cloud dns zone `example-net-zone`

Subdomains on example.com and example.net are synonyms of each other, so for simplicity, the A records are in `example-com-zone`, and `example-net-zone` just has CNAMEs to the equivalent domain in `example-com-zone`. So for example:
* `foo.example.com. A 1.2.3.4`
* `foo.example.net. CNAME foo.example.com.`

Unfortunately, when `googlecloud.getHostedZone` does a lookup of the `SOA` record (using `acme.FindZoneByFqdn`) for `foo.example.net`, the reported authority is `example.com`. I've confirmed this with `dig -t soa`.

As a result, we try to add the `txt` record in `example-com-zone` instead of `example-net-zone`, and the call fails.

TamalSaha:
Fixes #330

### DIFF
--- a/acme/dns_challenge.go
+++ b/acme/dns_challenge.go
@@ -262,9 +262,15 @@ func FindZoneByFqdn(fqdn string, nameservers []string) (string, error) {
 		if in.Rcode == dns.RcodeSuccess {
 			for _, ans := range in.Answer {
 				if soa, ok := ans.(*dns.SOA); ok {
-					zone := soa.Hdr.Name
-					fqdnToZone[fqdn] = zone
-					return zone, nil
+					if zone, err := publicsuffix.EffectiveTLDPlusOne(UnFqdn(fqdn)); err == nil {
+						zone = ToFqdn(zone)
+						fqdnToZone[fqdn] = zone
+						return zone, nil
+					} else {
+						zone = soa.Hdr.Name
+						fqdnToZone[fqdn] = zone
+						return zone, nil
+					}
 				}
 			}
 		}


acme/dns_challenge.go
```diff
@@ -262,9 +262,15 @@ func FindZoneByFqdn(fqdn string, nameservers []string) (string, error) {
 		if in.Rcode == dns.RcodeSuccess {
 			for _, ans := range in.Answer {
 				if soa, ok := ans.(*dns.SOA); ok {
-					zone := soa.Hdr.Name
-					fqdnToZone[fqdn] = zone
-					return zone, nil
+					if zone, err := publicsuffix.EffectiveTLDPlusOne(UnFqdn(fqdn)); err == nil {
```
xenolf: Wouldn't that lead to most lookups using `FindZoneByFqdn()` to use this branch instead of the original one - effectively bypassing the lookup?

TamalSaha: With the caveat that I am not a DNS protocol expect, this seems to be the best that can be done. First checking SOA records to ensure, this is a valid domain. Then if this is a known public suffix (known to the publicsuffix lib), then use that to determine zone. If this is unknown to public suffix lib (presumably a new TLD), then use the result from DNS server (but in that the result might be incorrect if CNAME is used).

PR #340, opened 2017-01-11
Reduce nonce locking

fxposter:
Do not lock on http request, lock only on datastructure changes. @xenolf WDYT?

### DIFF
--- a/acme/client_test.go
+++ b/acme/client_test.go
@@ -10,6 +10,7 @@ import (
 	"net/http/httptest"
 	"strings"
 	"testing"
+	"time"
 )
 
 func TestNewClient(t *testing.T) {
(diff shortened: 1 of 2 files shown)


xenolf:
Hello and thank you for the pull request!

Does this solve a concrete problem for you?

fxposter:
@xenolf We didn't have such problems yet, but it's definitely not a good thing to hold a lock while making network IO :)

I've added a test, which tries to show the problem...

fxposter:
@xenolf I actually introduced this behavior in https://github.com/xenolf/lego/pull/319, which definitely fixed the problem that we there (panics on accessing wrong indexes), but introduced locking during http request, which I missed at that point. :(

fxposter:
@xenolf And in "nonceManager" (the name, that I don't like :)) actually more clearly separates that "concurrent nonce management" from "signing" part that jws does.

PR #341, opened 2017-01-14
Close response body in error case and close first one

mholt:
I noticed that the response body is not closed in the error case of waiting for a certificate to be issued: https://sourcegraph.com/github.com/xenolf/lego@ce8fb060cb8361a9ff8b5fb7c2347fa907b6fcac/-/blob/acme/client.go#L673-674

AFAIK the response body needs to be closed even if there isn't one or it is not read. We close it above if the response code is 201 or 202, but not 200 or anything else. I think this will fix that problem.

### DIFF
--- a/acme/client.go
+++ b/acme/client.go
@@ -670,6 +670,7 @@ func (c *Client) requestCertificateForCsr(authz []authorizationResource, bundle 
 
 			break
 		default:
+			defer resp.Body.Close()
 			return CertificateResource{}, handleHTTPError(resp)
 		}
 
--- a/acme/client.go
+++ b/acme/client.go
@@ -23,6 +23,9 @@ var (
 	Logger *log.Logger
 )
 
+// maxBodySize is the maximum size of body that we will read.
+const maxBodySize = 1024 * 1024
+
 // logf writes a log entry. It uses Logger if not
 // nil, otherwise it uses the default log.Logger.
 func logf(format string, args ...interface{}) {
@@ -783,6 +786,3 @@ func validate(j *jws, domain, uri string, chlng challenge) error {
 		}
 	}
 }
-
-// maxBodySize is the maximum size of body that we will read.
-const maxBodySize = 1024 * 1024
--- a/acme/client.go
+++ b/acme/client.go
@@ -619,9 +619,6 @@ func (c *Client) requestCertificateForCsr(authz []authorizationResource, bundle 
 
 	maxChecks := 1000
 	for i := 0; i < maxChecks; i++ {
-		if i == maxChecks-1 {
-			return CertificateResource{}, fmt.Errorf("polled for certificate %d times; giving up", maxChecks)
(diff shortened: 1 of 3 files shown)


mholt:
Incidentally, I think there may be one more subtle leak, but I want someone else to verify.

On [this line](https://sourcegraph.com/github.com/xenolf/lego@ce8fb060cb8361a9ff8b5fb7c2347fa907b6fcac/-/blob/acme/client.go#L608-609) we initially get an *http.Response, which has a body that must be closed. I think the body for that very first response gets closed _if and only if_ it comes back with a 201 or 202 status code. If it doesn't, we replace `resp` with a new one [here](https://sourcegraph.com/github.com/xenolf/lego@ce8fb060cb8361a9ff8b5fb7c2347fa907b6fcac/-/blob/acme/client.go#L676-677) but we never closed the body of the first one.

I actually caught this thanks to Sourcegraph's code browser; when I clicked the `resp` variable, it took me up to that first line where I saw that we were re-using an earlier variable, not creating a new one.

Am I missing something or is that also a bug?

mholt:
Actually, I'm not sure if my initial PR is entirely correct. There are some places where handleHTTPError is called after a `defer resp.Body.Close()`, meaning in those cases, the body will be closed twice. I just pushed an amendment that should defer the Close() in the right place without duplicating/overlapping Close() calls.

acme/client.go
```diff
@@ -670,6 +670,7 @@ func (c *Client) requestCertificateForCsr(authz []authorizationResource, bundle
 
 			break
 		default:
+			defer resp.Body.Close()
```
miekg: Why not put this defer directly after the post, i.e. line 612? Then the `Close` in the other switch can be removed as well.

mholt: @miekg Because it might be read in line 621. If that's never the case, then sure, but I don't want to break things. I need feedback from @xenolf.

mholt:
@xenolf @miekg I've pushed a commit that ensures the body for both responses is closed. I had to pull the select out into its own function in order for it to be reasonably readable.

I also took `1024*1024` into a constant, and set a cap for the number of iterations (polling) we're allowed to do in checking for a certificate. There's no reason it should take 1000 polls to get a single cert, and it will be one extra layer of protection against some buggy protocol implementation, perhaps.

Tested and it works :) Please review ASAP. (Will go forward including this in new Caddy releases either way, though.)

acme/client.go
```diff
@@ -760,3 +783,6 @@ func validate(j *jws, domain, uri string, chlng challenge) error {
 		}
 	}
 }
+
+// maxBodySize is the maximum size of body that we will read.
+const maxBodySize = 1024 * 1024
```
xenolf: Could we move that up to the top? :)

mholt: Sure! Will do in a little bit.

PR #345, opened 2017-01-30
dns/route53: Allow specifying hosted zone ID

jen20:
This commit adds support for specifying hosted zone ID via the environment variable `AWS_HOSTED_ZONE_ID`. If this is not specified, the previous discovery process is used.

This is useful in environments where multiple hosted zones for the same domain name are present in an account.


xer0x:
+1 this would be nice. Sometimes I think lego might be grabbing the wrong zone id.

gianrubio:
I realise that lego doesn't work appropriately with delegated zones. BTW I merged this this PR and it fixed the issue.

xenolf:
Greetings! Thank you for this PR!

I realize that I'm a bit late to the party and I apologize for that.

PR #357, opened 2017-02-27
Add sleeps for new-authz rate limits

etiennetremel:
When creating bulk certificates, we often hit the acme rate limiting which is defined to 20 requests per second against the new-authz endpoint.

This PR add a workaround.

Note: letsencrypt set the limit to 20rps, but after some testing it seems that using a value of 18 work better.

### DIFF
--- a/acme/client.go
+++ b/acme/client.go
@@ -23,8 +23,15 @@ var (
 	Logger *log.Logger
 )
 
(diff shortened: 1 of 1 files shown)


xenolf:
Thanks for this! Sorry it took me a bit to look at it :)

PR #367, opened 2017-03-21
Move nonce retry from jws to http

manvalls:
The error raised by an "invalid nonce" response never appeared inside `jws.go`, but instead it was handled at `http.go`, so it makes sense to move the retry logic to that file. The previous code from `jws.go` had no effect and did not solve issues related to invalid nonces.

### DIFF
--- a/acme/http.go
+++ b/acme/http.go
@@ -107,22 +107,25 @@ func postJSON(j *jws, uri string, reqBody, respBody interface{}) (http.Header, e
 		switch err.(type) {
 
 		case NonceError:
-			resp, err := j.post(uri, jsonBytes)
+
+			// Retry once if the nonce was invalidated
+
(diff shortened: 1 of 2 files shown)


acme/http.go
```diff
@@ -97,10 +97,38 @@ func postJSON(j *jws, uri string, reqBody, respBody interface{}) (http.Header, e
...
 
 	if resp.StatusCode >= http.StatusBadRequest {
-		return resp.Header, handleHTTPError(resp)
+
+		err := handleHTTPError(resp)
+
+		switch err.(type) {
+
+		case NonceError:
+			resp, err := j.post(uri, jsonBytes)
```
mholt: To avoid confusion, rename resp to something else that isn't used higher in the scope.

manvalls: What about `retryResp`?

PR #363, opened 2017-03-17
Use DNSimple official Go client

jodosha:
Hi I work at DNSimple. We're helping OSS projects to migrate to our newer API clients.

I noticed that you are using `github.com/weppos/dnsimple-go`, which points to our deprecated API v1. This PR introduces `github.com/dnsimple/dnsimple-go`, our official client that points to our new API v2.

Read more at: https://blog.dnsimple.com/2016/12/api-v2-stable/

Thanks ✌️

PR #373, opened 2017-04-03
Update error messages for DNSimple provider

jodosha:
Hi again! Since last PR (#363), we recognised that error messages needed to be tweaked for the sake of usability.

Would you mind to merge this PR with these small adjustments? Thank you.


jodosha:
@xenolf Thank you!

PR #374, opened 2017-04-05
Update DNSimple instructions to use new env key

Maikon:
The help documentation still points to the old environment variable (`DNSIMPLE_API_KEY`) so attempts that use that fail with: `DNSimple OAuth token is missing`. 

This updates the help message with the correct key (`DNSIMPLE_OAUTH_TOKEN`).


mholt:
Thanks! I've updated Caddy's docs too.

cli.go
```diff
@@ -205,7 +205,7 @@ Here is an example bash command using the CloudFlare DNS provider:
 	fmt.Fprintln(w, "\tauroradns:\tAURORA_USER_ID, AURORA_KEY, AURORA_ENDPOINT")
 	fmt.Fprintln(w, "\tcloudflare:\tCLOUDFLARE_EMAIL, CLOUDFLARE_API_KEY")
 	fmt.Fprintln(w, "\tdigitalocean:\tDO_AUTH_TOKEN")
-	fmt.Fprintln(w, "\tdnsimple:\tDNSIMPLE_EMAIL, DNSIMPLE_API_KEY")
```
nanliu: Based on #363 should this be DNSIMPLE_BASE_URL instead of DNSIMPLE_EMAIL?

PR #376, opened 2017-04-09
Update Dockerfile to use alpine:latest

jauderho:
Add musl-dev to list of packages as it is now needed to build image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,9 @@
-FROM alpine:3.4
+FROM alpine:latest
 
 ENV GOPATH /go
+ENV SYSROOT /go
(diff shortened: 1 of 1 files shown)


xenolf:
Hey there! Thanks for the pull request. Why the change from latest to 3.5 if I may ask?

jauderho:
Per
https://developers.redhat.com/blog/2016/02/24/10-things-to-avoid-in-docker-containers/

Says to avoid using the latest tag. It works either way so I'm fine if you
prefer to use the latest tag instead.

xenolf:
Closed by merge of #409

PR #380, opened 2017-04-18
optional authoritative precheck

hightoxicity:
Hi guy,
Here at Blablacar, servers in charge of cert management can make http/https outgoing calls only though proxies (exaproxy), concerning DNS resolution, it only can speak udp with recursive internal dns, not with outside authoritative dns directly.
I think we are not alone to have such restrictions, in another hand, if my recursive DNS are up to date it means that they got refreshed by the fact authoritatives one are updated.
So I propose to make the authoritative checking optional.

Thx,
Tony.

### DIFF
--- a/acme/dns_challenge.go
+++ b/acme/dns_challenge.go
@@ -34,6 +34,8 @@ var RecursiveNameservers = getNameservers(defaultResolvConf, defaultNameservers)
 // DNSTimeout is used to override the default DNS timeout of 10 seconds.
 var DNSTimeout = 10 * time.Second
 
+var DNSNoAuthoritCheck = false
+
 // getNameservers attempts to get systems nameservers before falling back to the defaults
 func getNameservers(path string, defaults []string) []string {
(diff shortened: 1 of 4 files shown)


kfiles:
We are in the same boat, with split-DNS providing different internal and external results for DNS01 challenges. I have been able to work around it by forcing external authoritative servers at the cli with --dns-resolvers option, but that's an ugly hack. I would welcome an option to disable the pre-check. I have opened issue #481 to address.

EugenMayer:
what @kfiles proposes as hack is e.g. not even possible if LEGO is used as a lib by e.g. Traefik - a better fix for now is using "typetransparent" as the zone type, which will resolve all entries which are not in the private dns server, using recursion instead of right away telling, there is no local entry, so there is none. This fixes SOA record lookups, see https://github.com/xenolf/lego/issues/481#issuecomment-428545211

Any chance you reroll the patch @hightoxicity or did you just move away from LEGO and have no interest?

ldez:
Closed in favor of #700

PR #382, opened 2017-04-20
Expose challenge-related methods and types.

### LINKED ISSUE #378: Expose private client methods
I'd like to get a set of challenges, persist them someplace, and then later (i.e., in a separate request) solve the challenges and request a certificate. Because the relevant methods and types are all private, this isn't possible without copying and pasting a bunch of code from client.go. Would you be open to making `Client.getChallenges`, `Client.solveChallenges`, and `Client.requestCertificate` public methods, and `authorizationResource` a public type? I'd be happy to submit a patch if so.

cc @cnelson

jmcarp:
Resolves #378.

### DIFF
--- a/acme/client.go
+++ b/acme/client.go
@@ -327,13 +327,13 @@ DNSNames:
 		logf("[INFO][%s] acme: Obtaining SAN certificate given a CSR", strings.Join(domains, ", "))
 	}
 
-	challenges, failures := c.getChallenges(domains)
+	challenges, failures := c.GetChallenges(domains)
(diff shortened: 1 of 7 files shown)


jmcarp:
WDYT @xenolf ? I'll resolve merge conflicts if you're open to making these methods public.

jmcarp:
Ping @xenolf

xenolf:
Hello!

Sorry for leaving you hanging that long. I'm not opposed to this, but this is a breaking change and the newly exposed methods would need documentation.

ldez:
I closed because it's outdated since v2.

PR #385, opened 2017-04-26
Add description for RFC2136 env vars

jraby:
I just spent an hour reading the RFCs trying to remember what was the difference between `TSIG_KEY` and `TSIG_SECRET`... turns out the KEY is actually the key *name* as set in the nameserver config.

So here's a description of the variables so I don't have to look that up again next time.


mholt:
Thanks for going through that pain for the rest of us. :smile: We accept!

PR #386, opened 2017-04-27
RFC2136_TIMEOUT: tuneable DNS propagation timeout

jraby:
Useful for slower DNS environments.

Time string is parsed with time.ParseDuration, so units are mandatory
(eg. RFC2136_TIMEOUT=10m)


jraby:
this was useful to let this provider work with he.net which seems to initiates zone transfers within ~10m, but it is very inconsistent.

I didn't want to hardcode it to some big value so that it can fail fast in environment where slow DNS propagation would indicate problems.

PR #395, opened 2017-05-29
Switch dnspod-go

mukaiu:
Switch github.com/decker502/dnspod-go to github.com/mukaiu/dnspod-go

Dnspod API changes lead to dns set the exception, @decker502 failed to deal with the update pr, now switch dnspod-go.


mholt:
Excellent, thank you for taking initiative on this! Glad the original repo has updated.

PR #399, opened 2017-06-12
Allow for using header X-Forwarded-Host

bkrukowski:
Using `X-Forwarded-Host` header would be useful for proxies

### DIFF
--- a/acme/client.go
+++ b/acme/client.go
@@ -128,14 +128,14 @@ func (c *Client) SetChallengeProvider(challenge Challenge, p ChallengeProvider) 
 //
 // NOTE: This REPLACES any custom HTTP provider previously set by calling
 // c.SetChallengeProvider with the default HTTP challenge provider.
-func (c *Client) SetHTTPAddress(iface string) error {
+func (c *Client) SetHTTPAddress(iface string, allowXForwardedHost bool) error {
 	host, port, err := net.SplitHostPort(iface)
 	if err != nil {
 		return err
 	}
 
 	if chlng, ok := c.solvers[HTTP01]; ok {
-		chlng.(*httpChallenge).provider = NewHTTPProviderServer(host, port)
+		chlng.(*httpChallenge).provider = NewHTTPProviderServer(host, port, allowXForwardedHost)
 	}
 
 	return nil
--- a/acme/http_challenge_server.go
+++ b/acme/http_challenge_server.go
@@ -11,17 +11,18 @@ import (
 // It may be instantiated without using the NewHTTPProviderServer function if
 // you want only to use the default values.
 type HTTPProviderServer struct {
-	iface    string
-	port     string
-	done     chan bool
-	listener net.Listener
+	iface               string
+	port                string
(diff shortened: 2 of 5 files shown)


acme/client.go
```diff
@@ -128,14 +128,14 @@ func (c *Client) SetChallengeProvider(challenge Challenge, p ChallengeProvider)
 //
 // NOTE: This REPLACES any custom HTTP provider previously set by calling
 // c.SetChallengeProvider with the default HTTP challenge provider.
-func (c *Client) SetHTTPAddress(iface string) error {
+func (c *Client) SetHTTPAddress(iface string, allowXForwardedHost bool) error {
```
mholt: :-1: on changing the function signature.

acme/http_challenge_server.go
```diff
@@ -11,17 +11,18 @@ import (
...
+	listener            net.Listener
+	allowXForwardedHost bool
 }
 
 // NewHTTPProviderServer creates a new HTTPProviderServer on the selected interface and port.
 // Setting iface and / or port to an empty string will make the server fall back to
 // the "any" interface and port 80 respectively.
-func NewHTTPProviderServer(iface, port string) *HTTPProviderServer {
-	return &HTTPProviderServer{iface: iface, port: port}
+func NewHTTPProviderServer(iface, port string, allowXForwardedHost bool) *HTTPProviderServer {
```
mholt: Again, let's not change the function signature. :-1:

acme/http_challenge_server.go
```diff
@@ -58,7 +59,15 @@ func (s *HTTPProviderServer) serve(domain, token, keyAuth string) {
...
 	mux := http.NewServeMux()
 	mux.HandleFunc(path, func(w http.ResponseWriter, r *http.Request) {
-		if strings.HasPrefix(r.Host, domain) && r.Method == "GET" {
+		requestHost := r.Host
+		if s.allowXForwardedHost {
+			xForwardedHost := r.Header.Get("x-forwarded-host")
+			if xForwardedHost != "" {
+				requestHost = xForwardedHost
+			}
+		}
```
mholt: Why can't this be done in pre-processing? I'd much rather this kind of thing live outside the acme lib, it seems like kind of a hack.

xenolf:
Hey there! Do I understand this right that this allows for the challenge handler to receive the incoming domain boulder is validating if there is a proxy in between?

bkrukowski:
@xenolf yes, exactly

xenolf:
> @xenolf yes, exactly

Okay. So why not just use the header as a fallback if the `Host` header does not match any expected domain? This would get rid of all the other changes needed and still achieve the goal.
@mholt what do you think?

mholt:
Is there any more risk here of an attacker changing or setting this header value as opposed to changing the Host header?

ldez:
It's highly insecure because `X-Forwarded-Host` can be spoof.

ldez:
It's time to close and decline this PR.

Thanks.

PR #408, opened 2017-07-11
RFC: providers/dns: add Service Account authorization option to gcloud

danrl:
Using a Service Account file allows us to operate easier in an automated environment. Service Account files can be obtained from GCP IAM. If the GCE_SERVICE_ACCOUNT_FILE env var is set, it is used for authentication.

We may discuss a fallback to NewDNSProviderCredentials if NewDNSProviderServiceAccount fails.

### DIFF
--- a/providers/dns/googlecloud/googlecloud.go
+++ b/providers/dns/googlecloud/googlecloud.go
@@ -4,12 +4,14 @@ package googlecloud
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
 	"time"
 
 	"github.com/xenolf/lego/acme"
 
 	"golang.org/x/net/context"
+	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/google"
 
 	"google.golang.org/api/dns/v1"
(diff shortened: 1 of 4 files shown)


danrl:
Updated to branch to reflect the latest master.

Any comments on the PR?

providers/dns/googlecloud/googlecloud.go
```diff
@@ -22,9 +24,15 @@ type DNSProvider struct {
...
 
 // NewDNSProvider returns a DNSProvider instance configured for Google Cloud
-// DNS. Credentials must be passed in the environment variable: GCE_PROJECT.
+// DNS. Project name must be passed in the environment variable: GCE_PROJECT.
+// A Service Account file can be passed in the environment variable:
+// GCE_SERVICE_ACCOUNT_FILE
 func NewDNSProvider() (*DNSProvider, error) {
 	project := os.Getenv("GCE_PROJECT")
+	saFile := os.Getenv("GCE_SERVICE_ACCOUNT_FILE")
+	if len(saFile) > 0 {
```
xenolf: I would replace this whole check with a call to `os.LookupEnv` as the value is only used within the if block.

xenolf:
Thank you for your time working on this PR!

> We may discuss a fallback to NewDNSProviderCredentials if NewDNSProviderServiceAccount fails.

Lego defaults to failing with a helpful error message if a configuration parameter was faulty, I think we should keep it like that.

providers/dns/googlecloud/googlecloud.go
danrl: Thanks for the review! I changed it as requested.

xenolf: Thanks for the quick change! One more nitpick 😅 
`if saFile, ok := os.LookupEnv("GCE_SERVICE_ACCOUNT_FILE"); ok {
}`

danrl: Done via fixup and rebase :)

danrl:
Thx! Awesome!

PR #409, opened 2017-07-14
Dockerfile broken with old alpine version with old go

olegstepura:
alpine 3.4 no longer works with error
```
package context: unrecognized import path "context" (import path does not begin with hostname)
```

alpine 3.5 with no changes did not work with error
```
# runtime/cgo
/tmp/go-build671992352/runtime/cgo/_obj/_cgo_export.c:2:20: fatal error: stdlib.h: No such file or directory
 #include <stdlib.h>
                    ^
compilation terminated.
```

alpine 3.6 with no changes did not work with error:
```
# github.com/xenolf/lego
/usr/lib/go/pkg/tool/linux_amd64/link: running gcc failed: exit status 1
/usr/lib/gcc/x86_64-alpine-linux-musl/6.3.0/../../../../x86_64-alpine-linux-musl/bin/ld: cannot find Scrt1.o: No such file or directory
/usr/lib/gcc/x86_64-alpine-linux-musl/6.3.0/../../../../x86_64-alpine-linux-musl/bin/ld: cannot find crti.o: No such file or directory
/usr/lib/gcc/x86_64-alpine-linux-musl/6.3.0/../../../../x86_64-alpine-linux-musl/bin/ld: cannot find -lpthread
/usr/lib/gcc/x86_64-alpine-linux-musl/6.3.0/../../../../x86_64-alpine-linux-musl/bin/ld: cannot find -lssp_nonshared
collect2: error: ld returned 1 exit status
```


xenolf:
Hey there! Thanks for your contribution!

> alpine 3.6 with no changes did not work with error:

So does the error happen with 3.6 as well? I'm a bit confused as you propose to change the version to 3.6

olegstepura:
Hi! Nope, there is no error now, after I added `musl-dev` to 3.6

olegstepura:
I suggest you to fix stable version with some fixed commit in git. Otherwise the docker file will become broken in some time again later.

olegstepura:
This was tested in my environment and it works.

Dockerfile
```diff
@@ -1,13 +1,14 @@
-FROM alpine:3.4
+FROM alpine:3.6
 
 ENV GOPATH /go
 
-RUN apk update && apk add ca-certificates go git && \
+RUN apk update && apk add ca-certificates go git musl-dev && \
```
sfrique: I would change to `apk add --no-cache`  and remove the `rm -rf /var/cache/apk/*` lines (there are two of them)

Dockerfile
```diff
@@ -1,13 +1,14 @@
...
     rm -rf /var/cache/apk/* && \
     go get -u github.com/xenolf/lego && \
     cd /go/src/github.com/xenolf/lego && \
+    git checkout tags/v0.4.0 && \
```
sfrique: I would also add to beginning 
`LEGO_VERSION tags/v0.4.0`

and then change to:
`git checkout ${LEGO_VERSION}`

sfrique:
It worked ta my env as well, used dns at gcloud.

dschaper:
Didn't want to open a new PR since this one tracks the changes, but I've integrated `apk --no-cache` and virtual packages to this one. (It's a small Dockerfile, so I'll just inline the changes I made here, taking in to account the ENV)

```
FROM alpine:3.6

ENV GOPATH /go
ENV LEGO_VERSION tags/v0.4.0

RUN apk add --no-cache --virtual run-dependencies ca-certificates && \
    apk add --no-cache --virtual build-dependencies go git musl-dev && \
    go get -u github.com/xenolf/lego && \
    cd /go/src/github.com/xenolf/lego && \
    git checkout ${LEGO_VERSION} && \
    go build -o /usr/bin/lego . && \
    apk del build-dependencies && \
    rm -rf /go

ENTRYPOINT [ "/usr/bin/lego" ]
```

sfrique:
@dschaper great, i've comment instead of creating a new PR as well!

Hopefully it will get updated!

Dockerfile
olegstepura: Good point!

Dockerfile
```diff
@@ -1,15 +1,15 @@
...
 
 ENV GOPATH /go
+ENV LEGO_VERSION tags/v0.4.0
 
-RUN apk update && apk add ca-certificates go git musl-dev && \
-    rm -rf /var/cache/apk/* && \
+RUN apk update && apk add --no-cache --virtual run-dependencies ca-certificates && \
+    apk add --no-cache --virtual build-dependencies go git musl-dev && \
     go get -u github.com/xenolf/lego && \
-    cd /go/src/github.com/xenolf/lego && \
```
olegstepura: I also changed `/go` with `${GOPATH}`

Dockerfile
```diff
@@ -1,14 +1,15 @@
...
+    cd ${GOPATH}/src/github.com/xenolf/lego && \
+    git checkout ${LEGO_VERSION} && \
     go build -o /usr/bin/lego . && \
-    apk del go git && \
-    rm -rf /var/cache/apk/* && \
+    apk del build-dependencies && \
     rm -rf /go
```
sfrique: Hey, forgot the `/go` here!

olegstepura: Yep, thanks!

dschaper:
Now that Docker has released Build Stages to the stable release, would anyone be interested in a multi-stage build that would clean this up even further?

dschaper:
Implemented build stages in https://github.com/dschaper/lego/tree/docker-stages 

Original docker image: 27.3MB
Staged build: 18.9MB

Could be optimized further in the `go build` but this is a good compromise. I can open a new PR for this if needed.

sfrique:
@dschaper I like the idea of build stage, but most of my docker installs are not yet with the supported docker version. Kubernetes is algo not there, just to point it out.

The image ins't that big, so I guess it is fine without build stage. 
But feel free to open a new PR if you think it is better!

dschaper:
@sfrique Thanks for the tip on Kubernetes, I've closed the pull as that would be a breaking change.

sfrique:
@dschaper I may have miss spoke and some people may not understand what I meant. 
It's not a `kubernetes` problem, It is the GKE Container-Optimized OS, that one still uses docker 1.13 as stable. You can check here:
https://cloud.google.com/container-optimized-os/docs/release-notes
Also It will only break if the container is built inside the GKE cluster, running those images.

But like I said, still quite new, so not all base system have update it yet.

Sorry if I wasn't clear!

sfrique:
This will probably fail because of this:
49b9503635ff26b2b725667de187f05d5960acd0

With this error:
```
HEAD is now at 192334c... Lego version 0.4.0
providers/dns/exoscale/exoscale.go:10:2: cannot find package "github.com/pyr/egoscale/src/egoscale" in any of:
	/usr/lib/go/src/github.com/pyr/egoscale/src/egoscale (from $GOROOT)
	/go/src/github.com/pyr/egoscale/src/egoscale (from $GOPATH)
```

Dockerfile
```diff
@@ -1,14 +1,15 @@
...
+ENV LEGO_VERSION tags/v0.4.0
 
-RUN apk update && apk add ca-certificates go git && \
-    rm -rf /var/cache/apk/* && \
+RUN apk update && apk add --no-cache --virtual run-dependencies ca-certificates && \
+    apk add --no-cache --virtual build-dependencies go git musl-dev && \
     go get -u github.com/xenolf/lego && \
-    cd /go/src/github.com/xenolf/lego && \
+    cd ${GOPATH}/src/github.com/xenolf/lego && \
+    git checkout ${LEGO_VERSION} && \
```
sfrique: Because of recent path changes you need to add this ugly hack to fix it:
`git cherry-pick --no-commit 49b9503635ff26b2b725667de187f05d5960acd0`
just after the checkout.

olegstepura:
@sfrique Will this happen despite we use fixed version of lego (`git checkout tags/v0.4.0`) ?

sfrique:
@olegstepura yes

`go get -u` get the dependencies and it fails.

olegstepura:
Ah, I got it! Yep, seems like new tag is required with adjusting it in this PR as well.

xenolf:
I will release 0.4.1 today which should fix that. Could you try running it against current master to see if it runs okay?

olegstepura:
@xenolf Image build was successful

xenolf:
I pushed the v0.4.1 release tag to github. Please change the PR to use this and I will merge it in.

olegstepura:
I updated lego version tag

xenolf:
Merged. Thank you for your work! :)

PR #415, opened 2017-08-06
Add DNS Provider for CloudXNS

giuem:
Add DNS Provider for [CloudXNS](https://www.cloudxns.net/Index/index.html?l=en-us).

[API Document](https://www.cloudxns.net/Support/detail/id/1361.html) (Chinese version only)

### DIFF
--- a/providers/dns/cloudxns/cloudxns.go
+++ b/providers/dns/cloudxns/cloudxns.go
@@ -183,8 +183,7 @@ func (c *DNSProvider) makeRequest(method, uri string, body []byte) (json.RawMess
 	req.Header.Set("API-HMAC", c.hmac(url, requestDate, string(body)))
 	req.Header.Set("API-FORMAT", "json")
 
-	client := http.Client{Timeout: 60 * time.Second}
-	resp, err := client.Do(req)
+	resp, err := acme.HTTPClient.Do(req)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


providers/dns/cloudxns/cloudxns.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+
+	requestDate := time.Now().Format(time.RFC1123Z)
+
+	req.Header.Set("API-KEY", c.apiKey)
+	req.Header.Set("API-REQUEST-DATE", requestDate)
+	req.Header.Set("API-HMAC", c.hmac(url, requestDate, string(body)))
+	req.Header.Set("API-FORMAT", "json")
+
+	client := http.Client{Timeout: 60 * time.Second}
```
mholt: I think the client should be reused; in fact, acme.HTTPClient might be the right thing to use here. - @xenolf correct me if I'm wrong.

providers/dns/cloudxns/cloudxns.go
```diff
@@ -0,0 +1,215 @@
...
+	_, err := c.makeReauest("DELETE", fmt.Sprintf("record/%s/%s", recordID, zoneID), nil)
+	return err
+}
+
+func (c *DNSProvider) hmac(url, date, body string) string {
+	sum := md5.Sum([]byte(c.apiKey + url + body + date + c.secretKey))
+	return hex.EncodeToString(sum[:])
+}
+
+func (c *DNSProvider) makeReauest(method, uri string, body []byte) (json.RawMessage, error) {
```
mholt: This should be spelled `makeRequest` -- I understand how with a Chinese keyboard this could be difficult. :) But we would appreciate it!

PR #417, opened 2017-08-08
[RFC] Script DNS provider

iavael:
Prototype of integration with any DNS systems via external script/binary

### DIFF
--- a/providers/dns/exec/exec.go
+++ b/providers/dns/exec/exec.go
@@ -1,11 +1,11 @@
-// Package script implements a DNS provider for solving the DNS-01 challenge using
-// a user-provided script.
-package script
+// Package exec implements a DNS provider for solving the DNS-01 challenge using
+// a user-provided executable.
+package exec
 
 import (
 	"fmt"
 	"os"
-	"os/exec"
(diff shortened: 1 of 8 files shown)


mholt:
This is an interesting idea; so it looks like the script must conform to a certain CLI (with subcommands like present and cleanup). Is that a reasonable expectation in general?

iavael:
@mholt

>  Is that a reasonable expectation in general?

In general - yes. These two "providers" are for custom integration (e.g. with internal dns management system), so it should be expected to write a short script.


I hasn't documented script and pipe interfaces yet, because I'm still working on it. I gonna put README files in respected directories.

providers/dns/exec/exec.go
```diff
@@ -0,0 +1,53 @@
+// Package exec implements a DNS provider for solving the DNS-01 challenge using
+// a user-provided executable.
+package exec
```
mholt: Can the package be named something less "stdliby"? Like `dnsexec` maybe? Then we can ditch the renaming of the os/exec package.

mholt:
Closing in favor of #508

PR #423, opened 2017-08-24
Dockerfile stage build

dschaper:


dschaper:
#409 pre-empts, will re-open if there is interest.

PR #425, opened 2017-09-05
Fix gcloud naming inconsistency; parse projectID

danrl:
This commit fixes the naming inconsistency of Google Cloud DNS being
named `googlecloud` in the source and `gcloud` in the user interface.
Furthermore, improving my previous commit, I realized that the project ID
is stored in the service account file. We can save one environment variable
when using service account file (e.g. in automation applications).


xenolf:
Hello!

This will get merged into the next version bump to 0.5.0 to avoid breaking current implementations.

Issue #429, opened 2017-09-12
exoscale path change


carlwgeorge:
While trying to build caddy with all the dns providers enabled, I noticed the error `cannot find package "github.com/pyr/egoscale/src/egoscale"`.  I think I chased it down to the [correct place](https://github.com/xenolf/lego/blob/e370f2a4c5c211bcfa09ca2474a0a1ad64d20705/providers/dns/exoscale/exoscale.go#L10), but I'm fairly new to go so I was hoping you could confirm.  Checking the egoscale repo, I see that they recently [removed the src/egoscale path](https://github.com/exoscale/egoscale/commit/325740036187ddae3a5b74be00fbbc70011c4d96).

carlwgeorge:
Fixed in #430.  Thanks @sean-brandt!

PR #430, opened 2017-09-13
handle move of egoscale to exoscale

sean-brandt:
Fix for issue: #429 

@carlwgeorge - Ran into this at about the same time. :)


mholt:
Thank you!

PR #435, opened 2017-09-21
Add NewDNSProviderCredentials for AWS

tamalsaha:
- Removed region for tests as Route53 api is gobal.
- Supports AWS IAM Roles with MFA. https://aws.amazon.com/blogs/developer/assume-aws-iam-roles-with-mfa-using-the-aws-sdk-for-go/


tamalsaha:
We use lego as a library. We would appreciate a way to pass credentials in a factory method.

fd:
What's the status on this?

PR #436, opened 2017-09-24
Users of an effective top-level domain can use the DNS challenge

### LINKED ISSUE #434: DNS Challenge for PSL domains
Hi,

I'm not sure if this is an issue about lego or my DNS config, but I guess this is the first one.

I own a domain (hosted by Rackspace) published on the [PSL](https://github.com/publicsuffix/list) and try to create certs for subdomains with a DNS challenge but I get the following error:
`Error presenting token: Could not find the start of authority`

To be more specific I have my domain "example.com" in the PSL. I create a subdomain "cert.example.com". SOA record is published at the example.com level. 

Looking at `FindZoneByFqdn` give us clue about this behaviour, but I'm not sure how we can handle this.

Should we create an option to specify the zone and skip this function?

LeSuisse:
They will not get anymore an error message saying
"Could not find the start of authority".

Finding the zone cut of a FQDN now only rely on the presence
of a SOA record. Indeed, in the context of an eTLD the
authority will be the eTLD itself so you need to continue
to recurse until you get an answer instead of cutting the search
when you find the public suffix of a domain.

Fixes #434

### DIFF
--- a/acme/dns_challenge.go
+++ b/acme/dns_challenge.go
@@ -11,7 +11,6 @@ import (
 	"time"
 
 	"github.com/miekg/dns"
-	"golang.org/x/net/publicsuffix"
 )
 
 type preCheckDNSFunc func(fqdn, value string) (bool, error)
@@ -242,10 +241,6 @@ func FindZoneByFqdn(fqdn string, nameservers []string) (string, error) {
 	labelIndexes := dns.Split(fqdn)
 	for _, index := range labelIndexes {
 		domain := fqdn[index:]
-		// Give up if we have reached the TLD
-		if isTLD(domain) {
-			break
-		}
 
 		in, err := dnsQuery(domain, dns.TypeSOA, nameservers, true)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


tmm1:
I created another approach in #438 to fix the bug in #434 

However I'm not sure how else you could fix the `com.ac` vs `ac` issue. Both entries are listed in the PSL, so the loop always stops early at `com.ac`. Perhaps it's best to remove the PSL checks altogether as this PR proposes.

FYI the PSL dependency was first added in https://github.com/xenolf/lego/commit/4945919c

cc @middelink

tcottier-enalean:
Bypassing isTLD the way @LeSuisse did works for me.

LeSuisse:
Hello @tmm1,

Yep, I agree that's why I ended up removing the check. As of today lego tries to determine the zone cut by using both technical data (SOA records) and "administrative" data (the PSL) which does not end well.

Removing the PSL check only add, in the worst case, a few DNS queries before failing which seems reasonable.

xenolf:
Thank you! I believe this is the better solution for this problem for the reasons @LeSuisse outlined in his response above.

PR #438, opened 2017-09-28
Fix DNS challenge on private PSL subdomains

### LINKED ISSUE #434: DNS Challenge for PSL domains

tmm1:
Fixes #434 

@tcottier-enalean @LeSuisse Can you see if this patch works for your use-cases also?

### DIFF
--- a/acme/dns_challenge.go
+++ b/acme/dns_challenge.go
@@ -274,8 +274,8 @@ func FindZoneByFqdn(fqdn string, nameservers []string) (string, error) {
 }
 
 func isTLD(domain string) bool {
-	publicsuffix, _ := publicsuffix.PublicSuffix(UnFqdn(domain))
-	if publicsuffix == UnFqdn(domain) {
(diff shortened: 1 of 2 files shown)


tmm1:
I see that this approach doesn't fix the test added in #436 for `example.co.au.`

xenolf:
Closing this in favor of #436

PR #442, opened 2017-10-21
add concurrency to DNS Challenges

### LINKED ISSUE #173: Improvement: setting dns-01 challenge record for multiple domains in parallel

jasonwbarnett:
Fixes #173 

I don't actually expect this to get merged, but I'm hoping this will start some design decisions.

It's probably better to handle concurrency for DNS within the actual DNS providers due to the performance limitations which are provider specific (e.g. google API limits are different than AWS Route53...). Also, __how__ changes are made between providers is drastically different. For example, Google Cloud DNS you're able to submit an atomic change for all records in a single API call.

I only tested this against Google Cloud DNS and it works pretty well, but still slow because the poor handling of errors from `validate()`.

### DIFF
--- a/acme/dns_challenge.go
+++ b/acme/dns_challenge.go
@@ -8,6 +8,7 @@ import (
 	"log"
 	"net"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/miekg/dns"
@@ -19,8 +20,9 @@ type preCheckDNSFunc func(fqdn, value string) (bool, error)
 var (
 	// PreCheckDNS checks DNS propagation before notifying ACME that
 	// the DNS challenge is ready.
-	PreCheckDNS preCheckDNSFunc = checkDNSPropagation
-	fqdnToZone                  = map[string]string{}
+	PreCheckDNS  preCheckDNSFunc = checkDNSPropagation
+	fqdnToZone                   = map[string]string{}
+	fqdnToZoneMu                 = &sync.Mutex{}
 )
 
 const defaultResolvConf = "/etc/resolv.conf"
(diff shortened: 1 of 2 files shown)


acme/dns_challenge.go
```diff
@@ -108,13 +108,19 @@ func (s *dnsChallenge) Solve(chlng challenge, domain string) error {
...
 	})
 	if err != nil {
 		return err
 	}
 
+	extraWaitTime := 20 * time.Second
+	logf("[INFO][%s] DNS record is live, inserting %s of wait time to reduce the probability of a Let's Encrypt 400 error.", domain, extraWaitTime)
+	time.Sleep(extraWaitTime)
+
 	return s.validate(s.jws, domain, chlng.URI, challenge{Resource: "challenge", Type: chlng.Type, Token: chlng.Token, KeyAuthorization: keyAuth})
```
jasonwbarnett: If might make more sense to put a for loop around this and __then__ return so we can catch the crappy 400 errors from Let's Encrypt.

captncraig:
I've been thinking about this a lot lately. I feel like solving challenges concurrently may cause some problems. Some providers (gandi, namecheap, probably others) will definitely have issues if run concurrently, since they have to fully regenerate the zone for each change, and so could overwrite each other's changes. 

What if, instead of solving **concurrently**, we did a sort of two-phase solving process. First, present each authorization serially, then go back through and check dns, submit the validation, and clean up as a second pass.

I was thinking you could have an optional Solver interface that has a `Prepare` method or something. The driver could detect this in `solveChallenges` and switch to the dual pass mode for those solvers.

ldez:
The parallelization of challenges (concurrent access to the DNS API) produces a major problem with wildcards or subdomains: the DNS will have to modify the same domain which will create an inconsistent entry (one of the entries can be overwritten by another).

For the majority of the DNS providers (by example GCloud) it will not work.

The solution #607 is more suitable for reducing the time to obtain certificates.

Thanks for the contribution.

PR #448, opened 2017-11-04
travis: update go versions

dvrkps:


dvrkps:
@xenolf ping.

xenolf:
Thanks!

Did this change at some point? I was under the impression that omitting the `.x` would do the same.

PR #449, opened 2017-11-13
Fix zone detection for cross-zone cnames

fd:
CNAMEs cannot co-exist with SOA records so responses with
a CNAME should be skipped.

The `cross-zone-example.assets.sh.` is currently hosted by
me (@fd) and will continue to exist for as long as the assets.sh
domain exists. (The assets.sh domain is used as a CDN and is unlikely
to go away.)

See #330

### DIFF
--- a/acme/dns_challenge.go
+++ b/acme/dns_challenge.go
@@ -258,14 +258,7 @@ func FindZoneByFqdn(fqdn string, nameservers []string) (string, error) {
 
 			// CNAME records cannot/should not exist at the root of a zone.
 			// So we skip a domain when a CNAME is found.
-			hasCNAME := false
-			for _, ans := range in.Answer {
-				if _, ok := ans.(*dns.CNAME); ok {
-					hasCNAME = true
-					break
-				}
-			}
-			if hasCNAME {
+			if dnsMsgContainsCNAME(in) {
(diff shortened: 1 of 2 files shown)


fd:
@xenolf 🏓  This patch has been running in our production environment for a while now. (it processed ±1000 domains with DNS hosted by various providers)

acme/dns_challenge.go
```diff
@@ -255,6 +255,20 @@ func FindZoneByFqdn(fqdn string, nameservers []string) (string, error) {
 
 		// Check if we got a SOA RR in the answer section
 		if in.Rcode == dns.RcodeSuccess {
+
+			// CNAME records cannot/should not exist at the root of a zone.
+			// So we skip a domain when a CNAME is found.
+			hasCNAME := false
+			for _, ans := range in.Answer {
+				if _, ok := ans.(*dns.CNAME); ok {
+					hasCNAME = true
```
xenolf: Would this look better if we had a label on the outer loop and would continue right here? Minor thing but what do you think?

fd: It might be even better to extract it into a small helper function.
I'll prepare a patch.

fd:
🎉

PR #451, opened 2017-11-19
Support for DNS Provider: nic.ru

nordicdyno:
Add support for the most popular DNS in Russia)
It tested manually and works fine in our case

### DIFF
--- a/providers/dns/nicru/nicru.go
+++ b/providers/dns/nicru/nicru.go
@@ -1,0 +1,329 @@
+package nicru
+
+import (
+	"bytes"
+	"encoding/json"
+	"encoding/xml"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"os"
+	"strings"
+	"time"
+
+	"github.com/xenolf/lego/acme"
+)
+
+var baseURL = "https://api.nic.ru"
+
+// DNSProvider is an implementation of the acme.ChallengeProvider interface that uses
+// DNSMadeEasy's DNS API to manage TXT records for a domain.
+type DNSProvider struct {
+	baseURL     string
+	apiUsername string
+	apiPassword string
+	apiID       string
+	apiSecret   string
+	serviceName string
+	auth        *authResponse
+	client      http.Client
+}
+
+type authResponse struct {
+	ExpiresIn   string `json:"expires_in"`
+	AccessToken string `json:"access_token"`
+	TokenType   string `json:"token_type"`
+}
+
+type authError struct {
+	Error string `json:"error"`
+}
+
+// Record holds the nic.ru API representation of a DNS Record
+type Record struct {
+	ID       int    `json:"id"`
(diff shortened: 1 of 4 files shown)


providers/dns/nicru/nicru.go
```diff
@@ -0,0 +1,329 @@
...
+	provider := &DNSProvider{
+		baseURL: baseURL,
+
+		apiUsername: nicruUsername,
+		apiPassword: nicruPassword,
+		apiID:       nicruAPIKey,
+		apiSecret:   nicruAPISecret,
+		serviceName: nicruServiceName,
+
+		client: http.Client{Timeout: 30 * time.Second},
```
mholt: Am thinking this should use `acme.HTTPClient` instead

providers/dns/nicru/nicru.go
```diff
@@ -0,0 +1,329 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return provider, nil
+}
+
+// getCredentials uses the supplied credentials to return a
+// DNSProvider instance configured for nic.ru
+func (d *DNSProvider) getAuthRespone() error {
+	// TODO: validate all credential are set
```
mholt: Should we wait to merge this until this is done? I imagine if credentials aren't set, the requests fail anyway.

providers/dns/nicru/nicru.go
```diff
@@ -0,0 +1,329 @@
...
+		body, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return fmt.Errorf("nicru auth failed, code=%v with unreadable body (%s)", resp.StatusCode, err)
+		}
+
+		var errInfo authError
+		err = json.Unmarshal([]byte(body), &errInfo)
+		if err != nil {
+			return fmt.Errorf("nicru auth failed, code=%v without parsable error: %s, body: %s", resp.StatusCode, err, body)
+		}
```
mholt: You can simplify this whole section by using `json.NewDecoder(resp.Body).Decode(&errInfo)` like you do below

providers/dns/nicru/nicru.go
```diff
@@ -0,0 +1,329 @@
...
+	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", d.auth.AccessToken))
+
+	resp, err := d.client.Do(req)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != 200 {
+		body, err := ioutil.ReadAll(resp.Body)
```
mholt: This could be moved outside the `if` statement because you do the same thing below too.

nordicdyno:
@ldez sorry, I don't have access to nic.ru account on my current work, and also no need this solution anymore.

ldez:
I close this PR because it can not be finalized or maintained by the contributor.

Thanks anyway for the contribution.

PR #454, opened 2017-12-15
Manage nonces pushed concurrency bewteen validation goroutines

hightoxicity:
Having certs with many SAN certificates, we encounter a lot of errors with bad nonces.
Analyzing code we see that concurrency is managed on nonce pushed to nonces.
But several goroutines are managing in parallel the challenges and they pop the last pushed occurence of nonce, sometime the pop() happen bewteen first attempt and the retry and another goroutine pushed a new nonce.

To fix this issue, I set a mutex to allow only one goroutine at a time to deal with jws post + nonce stuff.

### DIFF
--- a/acme/http.go
+++ b/acme/http.go
@@ -9,11 +9,14 @@ import (
 	"net/http"
 	"runtime"
 	"strings"
+	"sync"
 	"time"
 )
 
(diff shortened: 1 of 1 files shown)


hightoxicity:
Something even better would consist in pushing nonces under FQDN namespaces.

PR #473, opened 2018-01-20
Slightly improve Dyn DNS provider error reporting.

rbroemeling:
If Dyn responds with a 3xx or 4xx status code, information describing exactly
what went wrong is generally included in the body of the response (as part of
the typical Dyn JSON response). On the other hand, if Dyn responds with a 5xx
status code, we very likely have extremely limited information.

This commit modifies the reporting to display the explanatory messages included
in the body of the Dyn response for 3xx and 4xx status codes. The intent is to
make it much easier to determine what might be going wrong (when something is
going wrong).


mholt:
Thank you!

Issue #474, opened 2018-01-20
[FEATURE] Clarify/Improve DNS Propagation Timeouts


rbroemeling:
Hi,

So I've encountered this error while testing the latest `lego` for use within our infrastructure:

```
Time limit exceeded. Last error: NS ns5.dnsmadeeasy.com. returned NXDOMAIN for <domain>
```

I understand why it occurs, and have looked far enough into it to see how I can adjust the source code to increase the propagation timeout (by modifying the provider's timeout). In my particular case, the timeout being used is for the "Dyn" provider -- and I'm hitting the timeout when the DNS propagation check happens to choose one of our secondary DNS provider's (DNS Made Easy) servers. When it happens to choose a Dyn (primary) server, the 60s allowed by the Dyn provider timeout is enough; when it chooses a DNS Made Easy (secondary) server, the 60s allowed by the Dyn provider timeout is _not_ enough.

I've also browsed through the various open issues around DNS timeouts, and see that this is a really pretty common problem.

That all said, I'm here to argue that fundamentally, timeouts can have per-provider defaults but that default should be overridable by a command-line argument. Providers having differing defaults makes sense (since generally they'll have different propagation latency characteristics), but really each individual's system is a little different (i.e., like mine; where the secondary provider servers aren't guaranteed to receive the propagation of the change within the necessary window). Allowing over-ride of this via a command-line argument in order to allow users to make adjustments for their specific situation makes sense to me.

What I would suggest is that an argument be implemented, and then the code in `dns_challenge.go` be updated to prioritize the argument over all other cases. i.e.,

```go
// Insert code here that uses the timeout from the command-line, if set.
switch provider := s.provider.(type) {
case ChallengeProviderTimeout:
	timeout, interval = provider.Timeout()
default:
	timeout, interval = 60*time.Second, 2*time.Second
}
```

I'm not completely clear on what this command-line argument should be: `--dns-timeout` already exists and is used for something different (and is symmetrical with `--http-timeout`). Perhaps `--dns-propagation-timeout` should be implemented, and then the usage/help text for `--dns-timeout` should be updated as well so that the difference between the two of them is clear?

I'm willing to take this on and provide a PR for the change, but as per the contribution guidelines I wanted to check here first to see whether such a PR would be welcomed or not.

rbroemeling:
The more I looked at it, the more I was sure it wasn't a bunch of work; so I spent a few minutes doing an initial/WIP implementation -- see #475.

[...]

PR #475, opened 2018-01-20
Initial implementation of `--dns-propagation-time`.

rbroemeling:
#474: this is an initial (WIP) implementation of allowing users to override DNS propagation time with a `--dns-propagation-time` command-line argument.

### DIFF
--- a/acme/dns_challenge.go
+++ b/acme/dns_challenge.go
@@ -31,6 +31,9 @@ var defaultNameservers = []string{
 
 var RecursiveNameservers = getNameservers(defaultResolvConf, defaultNameservers)
 
+// DNSPropagationTimeout is used to override the DNS propagation timeout.
+var DNSPropagationTimeout time.Duration
+
 // DNSTimeout is used to override the default DNS timeout of 10 seconds.
 var DNSTimeout = 10 * time.Second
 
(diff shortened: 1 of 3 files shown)


rbroemeling:
@mholt / @xenolf Is there any comment on this PR? I believe it to be clean, but I'm happy to push updates to it if there's a problem that we need to fix. We require this functionality in order to use `lego`, and I'd prefer if we didn't have to maintain our patch in perpetuity.

acme/dns_challenge.go
```diff
@@ -31,6 +31,10 @@ var defaultNameservers = []string{
 
 var RecursiveNameservers = getNameservers(defaultResolvConf, defaultNameservers)
 
+// DNSPropagationTimeout is used to override the DNS propagation timeout (if left unset, this timeout is inherited from
+// the DNS provider if implemented, or defaults to 60 seconds if not).
+var DNSPropagationTimeout = 0 * time.Second
```
mholt: Can just be:

```
var DNSPropagationTimeout time.Duration
```

And simplify the comment; it makes claims that aren't about the variable being defined, and which will easily become wrong if timeout values are changed ~100 lines down.

rbroemeling:
@mholt Your requested changes have been made and the PR has been updated.

yellowmegaman:
Sounds really useful, for no apparent reason letsencrypt staging server fails to acknowledge TXT record in time.

lenovouser:
I think this can be merged?

ldez:
Closed in favor of  #634

PR #484, opened 2018-02-04
Update Dockerfile

pall-valmundsson:
The Docker container cannot currently be built because of exoscale dependency updates, which were fixed in #468.

This updates the Dockerfile to use multi-stage builds and enables a build argument to specify which version or SHA is built.


greut:
or release a new version?

PR #491, opened 2018-02-13
Remove CRLF in CloudXNS

ldez:
As we commit vendor, the CRLF breaks our dependencies validation system.


ldez:
Updated to master.

PR #506, opened 2018-03-21
cname dns challenge

ConradWood:
This patch looks up cname records instead of only a records for _acme-challenge. If no cname is found, or an error is encountered when looking up cname, it will continue to use the same codepath as before. The lookup happens before the challenge is completed and will use the same verificiation code path, except that, if the cname lookup succeeds, it will update the nameserver serving the domain where the cname points to.
Example:
zonefile "singingcat.net":
www                            IN A     46.16.75.24
_acme-challenge.www            IN CNAME     acme6.acme.conradwood.net.

I dynamically update the domain "acme.conradwood.net" to complete the challenge and NOT singingcat.net.


ldez:
Closed by  #791

PR #508, opened 2018-03-21
Add DNS challenge provider 'exec'

### LINKED ISSUE #505: DNS challenge provider: External script?
Hi,

thank you all very much for your work on lego! I'd like to use lego (the CLI) for a project which requires using the DNS challenge and running custom code. So while I could add a custom DNS challenge provider and implement the code needed to setup the DNS record in Go, I'm wondering: Would you be willing to merge a new DNS challenge provider which calls out to an external program or script?

The background is that for this project, we're using infrastructure already in place, and all Go code that I would be writing to use it with lego would be highly specific for our setup and of no use to anybody else. Encapsulating this code in a small script (or separate Go program) feels like a much better idea. I can imagine that this is something which may be helpful to other users as well.

It could be named `external` or `script`, the name of the program/script to call could be configured via an environment variable. The program/script would then be run by `lego` with arguments, e.g. `present fqdn value` and `cleanup fqdn value` or so.

Would that be something you'd be willing to merge? If so, I'd be happy to do a PR.

Thanks again!

fd0:
As discussed in #505, this commits adds a very simple DNS provider which calls out to an external program/script which must then add or remove the DNS record.

I'm not sure how much testing is required for such a simple provider, please advise.

Closes #505, supersedes #417

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -226,6 +226,7 @@ Here is an example bash command using the CloudFlare DNS provider:
 	fmt.Fprintln(w, "\tpdns:\tPDNS_API_KEY, PDNS_API_URL")
 	fmt.Fprintln(w, "\tdnspod:\tDNSPOD_API_KEY")
 	fmt.Fprintln(w, "\totc:\tOTC_USER_NAME, OTC_PASSWORD, OTC_PROJECT_NAME, OTC_DOMAIN_NAME, OTC_IDENTITY_ENDPOINT")
+	fmt.Fprintln(w, "\texec:\tEXEC_PATH")
 	w.Flush()
 
 	fmt.Println(`
--- a/providers/dns/exec/exec.go
+++ b/providers/dns/exec/exec.go
@@ -1,0 +1,72 @@
+// Package exec implements a manual DNS provider which runs a program for
+// adding/removing the DNS record.
+//
+// The file name of the external program is specified in the environment
+// variable EXEC_PATH. When it is run by lego, three command-line parameters
+// are passed to it: The action ("present" or "cleanup"), the fully-qualified domain
+// name, the value for the record and the TTL.
+//
+// For example, requesting a certificate for the domain 'foo.example.com' can
+// be achieved by calling lego as follows:
+//
+//  EXEC_PATH=./update-dns.sh \
+//    lego --dns exec \
+//    --domains foo.example.com \
+//    --email invalid@example.com run
+//
+// It will then call the program './update-dns.sh' with like this:
+//
+//  ./update-dns.sh "present" "_acme-challenge.foo.example.com." "MsijOYZxqyjGnFGwhjrhfg-Xgbl5r68WPda0J9EgqqI" "120"
+//
+// The program then needs to make sure the record is inserted. When it returns
+// an error via a non-zero exit code, lego aborts.
+//
+// When the record is to be removed again, the program is called with the first
+// command-line parameter set to "cleanup" instead of "present".
+package exec
+
+import (
+	"errors"
+	"os"
+	"os/exec"
+	"strconv"
+
+	"github.com/xenolf/lego/acme"
+)
+
+// DNSProvider adds and removes the record for the DNS challenge by calling a
+// program with command-line parameters.
+type DNSProvider struct {
+	program string
+}
+
+// NewDNSProvider returns a new DNS provider which runs the program in the
+// environment variable EXEC_PATH for adding and removing the DNS record.
+func NewDNSProvider() (*DNSProvider, error) {
+	s := os.Getenv("EXEC_PATH")
+	if s == "" {
+		return nil, errors.New("environment variable EXEC_PATH not set")
+	}
+
+	return &DNSProvider{program: s}, nil
+}
+
+// Present creates a TXT record to fulfil the dns-01 challenge.
+func (d *DNSProvider) Present(domain, token, keyAuth string) error {
+	fqdn, value, ttl := acme.DNS01Record(domain, keyAuth)
+	cmd := exec.Command(d.program, "present", fqdn, value, strconv.Itoa(ttl))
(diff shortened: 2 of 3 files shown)


providers/dns/script/script.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	fmt.Printf("init dns provider with script %v", s)
+	return &DNSProvider{script: s}, nil
+}
+
+// Present creates a TXT record to fulfil the dns-01 challenge.
+func (d *DNSProvider) Present(domain, token, keyAuth string) error {
+	fqdn, value, ttl := acme.DNS01Record(domain, keyAuth)
+	cmd := exec.Command(d.script, "add", fqdn, value, strconv.Itoa(ttl))
```
mholt: The argument value for the action should probably be "present" instead of "add"

providers/dns/script/script.go
```diff
@@ -0,0 +1,75 @@
...
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+
+	return cmd.Run()
+}
+
+// CleanUp removes the TXT record matching the specified parameters
+func (d *DNSProvider) CleanUp(domain, token, keyAuth string) error {
+	fqdn, value, ttl := acme.DNS01Record(domain, keyAuth)
+	cmd := exec.Command(d.script, "remove", fqdn, value, strconv.Itoa(ttl))
```
mholt: Same here, "cleanup" instead of "remove" to be consistent with lego and also more descriptive

providers/dns/script/script.go
```diff
@@ -0,0 +1,75 @@
...
+// It will then call the script './update-dns.sh' with like this:
+//
+//  ./update-dns.sh "add" "_acme-challenge.foo.example.com." "MsijOYZxqyjGnFGwhjrhfg-Xgbl5r68WPda0J9EgqqI" "120"
+//
+// The script then needs to make sure the record is inserted. When it returns
+// an error via a non-zero exit code, lego aborts.
+//
+// When the record is to be removed again, the script is called with the first
+// command-line parameter set to "remove" instead of "add".
+package script
```
mholt: Technically this can be any executable file or program, not just a script. Is there a better name we can use?

providers/dns/script/script.go
fd0: Yeah, I thought about that. From the point of view of a user, it's not clear why it's `present`. I had to look it up to. Since the script/program is called to "add" or "modify" a record, I thought it'd be clearer to the script/program author to call it "add".

Happy to change it though ;)

providers/dns/script/script.go
fd0: Here't it's easier, "cleanup" has nearly the same meaning as "remove", so we can just change it if you prefer.

providers/dns/script/script.go
fd0: I thought about this as well, but liked the other names even less:
 * `manual`
 * `ext`
 * `external`
 * `program`

The name `script` is the least ugly I came up with.

mholt: It's not a major thing, just want to get the optimal name the first time. How about something more general like "delegated" or "exec" (maybe not exec for a package name, of course).

If those aren't great then maybe we can just keep it the same, but I kinda like "delegated" or even "external".

fd0: Hm, then I'd prefer `exec`. The package can be called anything, we're just talking about the user interface, right?

providers/dns/script/script.go
fd0: So, let me know what you think.

fd0:
Updated, let me know what you think!

cli.go
```diff
@@ -226,6 +226,7 @@ Here is an example bash command using the CloudFlare DNS provider:
 	fmt.Fprintln(w, "\tpdns:\tPDNS_API_KEY, PDNS_API_URL")
 	fmt.Fprintln(w, "\tdnspod:\tDNSPOD_API_KEY")
 	fmt.Fprintln(w, "\totc:\tOTC_USER_NAME, OTC_PASSWORD, OTC_PROJECT_NAME, OTC_DOMAIN_NAME, OTC_IDENTITY_ENDPOINT")
+	fmt.Fprintln(w, "\tscript:\tSCRIPT_PATH")
```
mholt: Change to "exec" and "EXEC_PATH"

providers/dns/exec/exec.go
```diff
@@ -0,0 +1,72 @@
...
+// It will then call the program './update-dns.sh' with like this:
+//
+//  ./update-dns.sh "present" "_acme-challenge.foo.example.com." "MsijOYZxqyjGnFGwhjrhfg-Xgbl5r68WPda0J9EgqqI" "120"
+//
+// The program then needs to make sure the record is inserted. When it returns
+// an error via a non-zero exit code, lego aborts.
+//
+// When the record is to be removed again, the program is called with the first
+// command-line parameter set to "cleanup" instead of "present".
+package exec
```
mholt: I have to admit, the package name conflicting with one in the standard lib is unfortunate.

mholt:
LGTM except you missed a spot. ;) Then we'll merge and give it a shot.

cli.go
fd0: Oh, good catch!

providers/dns/exec/exec.go
fd0: I'm open to suggestions :)

I'm also okay with naming the package different to what we expose for the users, but that may introduce more confusion for everybody who looks at the code. But after all, the package is just used in exactly one place, in `dns_providers.go`, so the chance for a package name collision is very low.

fd0:
Fixed, thanks!

fd0:
Tried it in real life, works for the use case described in #505

fd0:
Awesome, thanks for taking the time to review this!

PR #512, opened 2018-03-26
Fix Domain does not exist error

dereknex:


dereknex:
@mholt I have a mistake when try merge from origin master. then I have clean it.

ldez:
@dereknex This message is related to `AllocateStaticIp` method, and this method is not used in Lego

https://github.com/aws/aws-sdk-go/blob/b2e866cc3d25372c6bde1ce18dbbf11862e954e9/service/lightsail/api.go#L96

dereknex:
@ldez  `CreateDomainEntry`  method seems to only available in the N. Virginia (us-east-1) Region.

https://github.com/aws/aws-sdk-go/blob/master/service/lightsail/api.go#L1167

ldez:
@dereknex I think this warning is out of date, but I cannot find any clear information about it.

Maybe you can create an env var to override this value and set the default value to `us-east-1`.

ldez:
Without any real source of trust, I restore the constraint on the region.

PR #515, opened 2018-03-27
Fixed CN regression in ACMEv2.

### LINKED ISSUE #514: ACMEv2: Common Name regression
When requesting a certificate for a domain which has subdomains using ACMEv2, domain order is not preserved. This potentially results in certificates having an inconsistent CN.

Currently, these domains are sorted alphabetically, but this is not guaranteed to be the case.

See https://github.com/xenolf/lego/issues/457#issuecomment-373420366 for more context.

mikepulaski:
This fixes #514. I didn't add tests because it doesn't seem that tests for ACMEv2 are in a working state yet.


xenolf:
Thanks a lot!

PR #518, opened 2018-03-27
ResolveAccountByKey no longer returns an EOF error on success.

### LINKED ISSUE #517: ACMEv2: ResolveAccountByKey returns EOF error
When calling `client.ResolveAccountByKey()`, an `io.EOF` error is returned. This is due to the fact that an ACME server does not necessarily return a response when an account is found, but rather a `Location` header, according to [section 7.3.1](https://ietf-wg-acme.github.io/acme/draft-ietf-acme-acme.html#rfc.section.7.3.1). 

It seems as though the client is attempting to read an empty response body.

See https://github.com/xenolf/lego/issues/457#issuecomment-373072789 for more details.

mikepulaski:
Fixes #517.

### DIFF
--- a/acmev2/client.go
+++ b/acmev2/client.go
@@ -189,7 +189,7 @@ func (c *Client) ResolveAccountByKey() (*RegistrationResource, error) {
 	logf("[INFO] acme: Trying to resolve account by key")
 
 	acc := accountMessage{OnlyReturnExisting: true}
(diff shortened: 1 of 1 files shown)


fd:
I can confirm that this works.

xenolf:
Thanks for catching this!
The tests should now be fixed. Could you add one please? :)

mikepulaski:
@xenolf Ok, done!

PR #523, opened 2018-04-02
Update dnsproviders in acmev2 branch

KnicKnic:
Desire is to update https://github.com/containous/traefik which is using acmev2 branch to support recently added dnsproviders.

Merging in master to acmev2 which only contains updates to the dns providers (and associated cli.go).


KnicKnic:
traefik forked lego.

PR #525, opened 2018-04-06
Digital Ocean: use TTL when creating DNS entry

ulrichSchreiner:
hi,

when creating a DNS entry with the digitalocean API the default TTL used by digitalocean will be 1800secs. Although letsencrypt will use a shorter TTL when query'ing the enttry it is longer than the default value of 120secs.

This PR appends a TTL to the digitalocean API when creating the TXT record.

### DIFF
--- a/providers/dns/digitalocean/digitalocean.go
+++ b/providers/dns/digitalocean/digitalocean.go
@@ -49,6 +49,7 @@ func (d *DNSProvider) Present(domain, token, keyAuth string) error {
 		RecordType string `json:"type"`
 		Name       string `json:"name"`
 		Data       string `json:"data"`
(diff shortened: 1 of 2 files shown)


nmengin:
Hello @ulrichSchreiner .

The feature seems to exist since the merge described [here](https://github.com/xenolf/lego/pull/555/files#diff-1ceac10ffc82d5f7e299fc1df47074c2).

The TTL has been fixed to `30 seconds` in the way to allow users to generate certificates for both a wildcard domain and its root domain (for example *.foo.com as CN and foo.com as SAN) is the allowed time.

ulrichSchreiner:
yes, you're right. i'll close the PR.

PR #529, opened 2018-04-15
Fix dnsimple api

dajenet:
Fixes:

> /go/src/github.com/xenolf/lego/providers/dns/dnsimple/dnsimple.go:179:49: cannot use whoamiResponse.Data.Account.ID (type int64) as type int in argument to strconv.Itoa


greut:
fyi, https://github.com/dnsimple/dnsimple-go/commit/83e0b9d4bd45f290bffb46486145b7693747302c

PR #535, opened 2018-04-22
Add support for reading DNS provider setup from files

### LINKED ISSUE #509: Add support for secrets in files
For applications deployed in Docker containers using Docker Swarm - secrets _must_ be stored in files instead of the environment. For reasons largely amounting to the Docker Swarm team's threat model, it seems they will never permit secrets to appear in the environment.

Many Docker images have been implementing the ability to suffix the names of environment variables with `_FILE` to cause the Docker container to read a file to get the value of a secret. For example, the WordPress Docker Image permits `WORDPRESS_DB_PASSWORD` if you wish to put the password in the environment directly, or `WORDPRESS_DB_PASSWORD_FILE` if you wish to have the Docker container read a file to determine the secret.

Lego is used by Traefik and currently my Cloudflare API keys are the only component that I can't populate from a Docker Swarm secret. Given that Traefik is often used as an ingress controller, I think that having support for a `CLOUDFLARE_API_KEY_FILE` et al for the various providers would be a significant value add.

Happy to open a pull request if there's interest in this.

farmdawgnation:
This PR adds the ability for lego to read DNS provider settings from files in addition to environment variables. To facilitate this I've continued the pattern that many Docker images follow wherein you can provide the credential using the regular environment variable name, like `CLOUDFLARE_API_KEY`, or you can provide the credential as a file and provide the path to that file by suffixing the normal env var with `_FILE`, for example `CLOUDFLARE_API_KEY_FILE`.

If you're interested, you can see the MySQL docker image doing that [here](https://github.com/docker-library/mysql/blob/2e4f2224f31e101d07c1e6a7c0da8245fde725c9/8.0/docker-entrypoint.sh#L21-L41).

In this implementation, the presence of the _regular_ environment variable will take precedence. This should mean that we minimize the overhead of running this method for folks who are just using the environment variables - which is the only thing that lego currently supports.

I've also standardized on using this for _all_ environment variables for DNS providers so as to not make any assumptions or start any debates about what folks should or should not treat as secrets - with the exception of Google Cloud which was already checking for a file _only_ because that's what Google actually gives you by way of credentials.

closes #509

### DIFF
--- a/acme/utils.go
+++ b/acme/utils.go
@@ -2,8 +2,6 @@ package acme
 
 import (
 	"fmt"
-	"io/ioutil"
-	"os"
 	"time"
 
 	"github.com/xenolf/lego/log"
@@ -32,31 +30,4 @@ func WaitFor(timeout, interval time.Duration, f func() (bool, error)) error {
 
 		time.Sleep(interval)
 	}
-}
-
-// Attempts to resolve 'key' as an environment variable. Failing that, it will
-// check to see if '$key_FILE' exists. If so, it will attempt to read from the
-// referenced file to populate a value.
-func GetenvOrFile(envVar string) string {
-	envVarValue := os.Getenv(envVar)
-
-	if envVarValue != "" {
-		return envVarValue
-	}
-
-	fileVar := envVar + "_FILE"
-	fileVarValue := os.Getenv(fileVar)
-
-	if fileVarValue == "" {
-		return envVarValue
-	}
-
-	fileContents, err := ioutil.ReadFile(fileVarValue)
-
-	if err != nil {
-		fmt.Printf("Error reading the file %s (defined by env var %s): %s\n", fileVarValue, fileVar, err)
-		return ""
-	}
-
-	return string(fileContents)
 }
--- a/acme/utils.go
+++ b/acme/utils.go
@@ -2,6 +2,8 @@ package acme
 
 import (
 	"fmt"
+	"io/ioutil"
+	"os"
 	"time"
 
 	"github.com/xenolf/lego/log"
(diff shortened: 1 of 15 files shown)


acme/utils.go
```diff
@@ -50,7 +50,7 @@ func GetenvOrFile(envVar string) string {
 	fileContents, err := ioutil.ReadFile(fileVarValue)
 
 	if err != nil {
-		fmt.Errorf("Error reading the file at " + fileVarValue + " (defined by " + fileVar + ")", err)
+		fmt.Printf("Error reading the file at " + fileVarValue + " (defined by " + fileVar + ")", err)
```
greut: The `Printf` will by lost with the extra arg. Something along that line maybe: `"Error reading %s (by %s). %s", fileVarValue ,fileVar, err`

farmdawgnation:
Ah good catch.

farmdawgnation:
This PR has been updated per review comments and is ready for further consideration!

farmdawgnation:
I'm not sure what's up with Travis CI... it... never reported CI results?

ldez:
Seems really similar to #243

farmdawgnation:
It does. I'm happy to continue work on this if there's interest but the lack of comments from @xenolf makes me think this won't get merged anytime soon. :(

farmdawgnation:
Not sure who else would be a good person to ask if this is wanted? @ldez perhaps? They have had some recent commits.

Postremus:
@farmdawgnation I am interested in this feature. :)
Any way I can help with this PR?

farmdawgnation:
I don't think so? I'm happy to push it over the finish line, but I do want some sense of interest from the maintainers or it's not really worth my time. :/

ldez:
After long thinking, to determine which design to take between the #243 and this PR, I chose to take your PR.
I don't like to have a lot of files but I think it's the best solution.

So, that great, your PR will be merged.

farmdawgnation:
Huzzah! Cheers!

endersonmaia:
any idea in what release this will be ?

PR #537, opened 2018-04-24
Adding output of which envvars are missing in Cloudflare and Azure

dalanmiller:
Just had a pretty rough time trying to determine which envvars were being surfaced to Caddy and which weren't when using systemd.

Thought that something like this might forgo someone the pain if they go through the same headache.

### DIFF
--- a/providers/dns/azure/azure.go
+++ b/providers/dns/azure/azure.go
@@ -28,7 +28,7 @@ type DNSProvider struct {
 	tenantId       string
 	resourceGroup  string
 
-	context        context.Context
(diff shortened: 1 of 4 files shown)


dalanmiller:
Not sure if you get test pass updates for every branch, looks like we're good to go here now! :+1: 

I got started fixing up some of the other ones and finished up azure, but given that the logic is going to pretty much be the same I'm going to do this in a different PR but have added azure here.

mholt:
Sure, thanks!

PR #540, opened 2018-05-04
Compilation fixes for acmev2 branch

scottjg:
I was able to use `lego` to fetch a multi-domain wildcard cert using the acmev2 branch. While it mostly worked, there were a few compilation fixes needed. Perhaps needed from changes to upstream deps?


scottjg:
merging in master does make the full diff a little tougher to review... but i think i've addressed your feedback.

mholt:
Hmm, I'd rather we just merge acmev2 into master. Will wait for @xenolf to decide on that though, since I wasn't sure if he's quite done with acmev2 yet.

PR #561, opened 2018-05-31
Added 'processing' status as valid challenge status

nelsonkram:
In ACME draft 10+, section 8 lists the valid challenge status' as:

`status (required, string):  The status of this challenge.  Possible
      values are: "pending", "processing", "valid", and "invalid".`

In acme/client.go the "processing" status is missing as a possible return value.

### DIFF
--- a/acme/client.go
+++ b/acme/client.go
@@ -822,6 +822,7 @@ func validate(j *jws, domain, uri string, c challenge) error {
 			log.Printf("[INFO][%s] The server validated our request", domain)
 			return nil
 		case "pending":
+		case "processing":
 		case "invalid":
 			return handleChallengeError(chlng)
 		default:


xenolf:
Hey, thanks for the PR!

The `processing` status was left out intentionally so it can be handled by the `default` branch of the switch as it does not indicate an error but rather that the challenge is still being validated by the server. 
Adding it to the switch like this would break the logic as it would end the loop and report an error, which would not be true at all.

nelsonkram:
Hi xenolf-

Sorry for the confusion on my part - I don't completely understand.  If the server replies with a challenge status of `processing`, the default switch case does not wait and throws the error `the server returned an unexpected state` - and processing stops.  Should this be the case?

xenolf:
@nelsonkram Sorry, I completely misread the code on mobile. You are right and the status was missing! :)

PR #565, opened 2018-06-03
Review DNS providers.

ldez:
- manage more finely the missing environment variables
- use Testify
- Fix bluecat `httpClient` argument
- use `defer` in tests to keep clean the environment.

### DIFF
--- a/providers/dns/auroradns/auroradns.go
+++ b/providers/dns/auroradns/auroradns.go
@@ -9,7 +9,7 @@ import (
 	"github.com/edeckers/auroradnsclient/records"
 	"github.com/edeckers/auroradnsclient/zones"
 	"github.com/xenolf/lego/acme"
-	"github.com/xenolf/lego/providers/dns/envvar"
+	"github.com/xenolf/lego/platform/config/env"
 )
 
 // DNSProvider describes a provider for AuroraDNS
@@ -23,7 +23,7 @@ type DNSProvider struct {
 // Credentials must be passed in the environment variables: AURORA_USER_ID
 // and AURORA_KEY.
 func NewDNSProvider() (*DNSProvider, error) {
-	values, err := envvar.Get("AURORA_USER_ID", "AURORA_KEY")
+	values, err := env.Get("AURORA_USER_ID", "AURORA_KEY")
 	if err != nil {
 		return nil, fmt.Errorf("AuroraDNS: %v", err)
 	}
--- a/providers/dns/auroradns/auroradns.go
+++ b/providers/dns/auroradns/auroradns.go
@@ -9,6 +9,7 @@ import (
 	"github.com/edeckers/auroradnsclient/records"
 	"github.com/edeckers/auroradnsclient/zones"
 	"github.com/xenolf/lego/acme"
+	"github.com/xenolf/lego/providers/dns/envvar"
 )
 
 // DNSProvider describes a provider for AuroraDNS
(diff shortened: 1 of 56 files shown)


providers/dns/auroradns/auroradns.go
```diff
@@ -9,6 +9,7 @@ import (
 	"github.com/edeckers/auroradnsclient/records"
 	"github.com/edeckers/auroradnsclient/zones"
 	"github.com/xenolf/lego/acme"
+	"github.com/xenolf/lego/providers/dns/envvar"
```
wyattjoh: Doesn't make sense to me to have this included in the `providers/dns/envvar` package I'd almost say instead to create a `platform/config/env` might be a better fit that would allow better extension. (To perhaps add a dynamic/database backed configuration package)

providers/dns/auroradns/auroradns.go
```diff
@@ -22,20 +23,23 @@ type DNSProvider struct {
 // Credentials must be passed in the environment variables: AURORA_USER_ID
 // and AURORA_KEY.
 func NewDNSProvider() (*DNSProvider, error) {
-	userID := os.Getenv("AURORA_USER_ID")
-	key := os.Getenv("AURORA_KEY")
+	values, err := envvar.Get("AURORA_USER_ID", "AURORA_KEY")
+	if err != nil {
+		return nil, fmt.Errorf("AuroraDNS: %v", err)
```
wyattjoh: Maybe a good idea to consider adding something like https://github.com/pkg/errors to replace our `fmt.Errorf` calls?

mholt: What's the problem that solves, exactly? (i.e. what's wrong with fmt.Errorf?)

wyattjoh: Using `fmt.Errorf` you lose the context of the original underlying error. This use case here it doesn't matter much, but when we have more complex error propagation through the application, it may make sense to actually do something different based on a higher error that occurred. Typical use case I see is when a web service needs to respond with the right error code in response to a database specific error, like `sql.ErrNoRows`, where you may want to return a `http.StatusNotFound`.

In general, `errors.Wrap` does exactly the same thing here, but gives us a future API that we can use to extend.

mholt: Hmm, what do you mean by context though?

```
return fmt.Errorf("while doing this thing: %v", err)
```

is what I do in all my programs and it works great, it easily makes a trail of followable error messages.

As far as I know, wrapping error values like `sql.ErrNoRows` still makes a type assertion necessary or makes a direct comparison impossible at least, whether that's with `fmt.Errorf` or some errors package.

I guess I'm just not personally convinced that we need to introduce another dependency for this, especially not in this PR...

wyattjoh: So completely agree, this particular issue shouldn't be resolved here, that's what I left it as a comment review rather than requesting changes 😉 

What I meant by context comes back to the specific lib I linked for errors. `pkg/errors` provides a [`errors.Cause(err error) error`](https://github.com/pkg/errors#retrieving-the-cause-of-an-error) function that will return the underlying actual error that was wrapped:

```go
package main

import "github.com/pkg/errors"

// MyError is some error type

func funcA() error {
  // some underlying error
  return &MyError{}
}

func funcB() error {
  if err := funcA(); err != nil {
    return errors.Wrap(err, "can't do A")
  }

  return nil
}

func main() {
  if err := funcB(); err != nil {
    switch err := errors.Cause(err).(type) {
    case *MyError:
            // handle specifically
    default:
            // unknown error
    }
  }
}
```

PR #571, opened 2018-06-07
ACME HTTP: Allow customizing HTTP client x509.CertPool

### LINKED ISSUE #569: Allow non-standard CA for validating ACME server HTTPS in development
Hi folks :wave:,

It would be wonderful if Lego was able to easily interact with a [Pebble](https://github.com/letsencrypt/pebble) instance. This is my own preferred way to debug issues with an ACME client and it may prove useful for integration testing now that Lego has the required ACME v2 support. 

The only stumbling block that I'm aware of is that Lego doesn't facilitate interacting with an ACME server that has a HTTPS certificate from a CA that isn't part of the system trust store. This _can_ be worked around by adding the Pebble CA certificate to the system's CA trust store but we [specifically recommend against this](https://github.com/letsencrypt/pebble#avoiding-client-https-errors) because it can easily compromise the security of a developer's system if done carelessly or forgotten about.

Would you be interested in a feature to specify a non-standard CA certificate via command line flag or an environment variable? For instance, Certbot can be made to trust the Pebble CA by using the `REQUESTS_CA_BUNDLE` env var.

It seems like the challenge is primarily in piping a `*x509.CertPool` populated with the provided non-standard CA to the `*http.Transport` instance that is used by the Lego `*http.Client` here:

https://github.com/xenolf/lego/blob/7fedfd1388f016c7ca7ed92a7f2024d06a7e20d8/acme/http.go#L18-L30

I'd be willing to submit a PR to do the work required but wanted to touch base first to see if the feature was interesting to you and whether there were recommendations on how you would like to see the `*x509.CertPool` propagated forward.

Thanks!

cpu:
This PR updates `acme/http.go` to allow customizing the `*x509.CertPool` used by the `HTTPClient` by specifying the filepath of one or more custom CA certificates via the `LEGO_CA_CERTIFICATES` environment variable.

This allows developers to easily trust a non-standard CA when interacting with an ACME test server (e.g. [Pebble](https://github.com/letsencrypt/pebble)):

```
LEGO_CA_CERTIFICATES=~/go/src/github.com/letsencrypt/pebble/test/certs/pebble.minica.pem \
lego \
  --server https://localhost:14000/dir \
  --email foo@bar.com \
  -d example.com \
  run
```

Resolves #569

### DIFF
--- a/acme/http.go
+++ b/acme/http.go
@@ -23,7 +23,7 @@ var (
 	// HTTPClient is an HTTP client with a reasonable timeout value and
 	// potentially a custom *x509.CertPool based on the caCertificateEnvVar
 	// environment variable (see the `initCertPool` function)
-	HTTPClient http.Client = http.Client{
+	HTTPClient = http.Client{
 		Transport: &http.Transport{
 			Proxy: http.ProxyFromEnvironment,
 			Dial: (&net.Dialer{
--- a/acme/http.go
+++ b/acme/http.go
@@ -23,7 +23,21 @@ var (
 	// HTTPClient is an HTTP client with a reasonable timeout value and
 	// potentially a custom *x509.CertPool based on the caCertificateEnvVar
 	// environment variable (see the `initCertPool` function)
-	HTTPClient http.Client
+	HTTPClient http.Client = http.Client{
+		Transport: &http.Transport{
+			Proxy: http.ProxyFromEnvironment,
+			Dial: (&net.Dialer{
+				Timeout:   30 * time.Second,
+				KeepAlive: 30 * time.Second,
+			}).Dial,
+			TLSHandshakeTimeout:   15 * time.Second,
+			ResponseHeaderTimeout: 15 * time.Second,
+			ExpectContinueTimeout: 1 * time.Second,
+			TLSClientConfig: &tls.Config{
+				RootCAs: initCertPool(),
+			},
+		},
+	}
 )
 
 const (
@@ -40,27 +54,6 @@ const (
 	// the system-wide trusted root list.
 	caCertificateEnvVar = "CA_CERTIFICATE"
 )
-
-// init populates the default HTTPClient instance. The httpClient's Transport's
-// TLSClientConfig is populated with initCertPool to allow developers to
-// override the system trust store for interacting with test ACME servers.
-func init() {
-	HTTPClient = http.Client{
-		Transport: &http.Transport{
-			Proxy: http.ProxyFromEnvironment,
-			Dial: (&net.Dialer{
-				Timeout:   30 * time.Second,
-				KeepAlive: 30 * time.Second,
-			}).Dial,
-			TLSHandshakeTimeout:   15 * time.Second,
-			ResponseHeaderTimeout: 15 * time.Second,
-			ExpectContinueTimeout: 1 * time.Second,
-			TLSClientConfig: &tls.Config{
-				RootCAs: initCertPool(),
-			},
-		},
-	}
-}
 
 // initCertPool creates a *x509.CertPool populated with the PEM certificate
 // found in the filepath specified in the caCertificateEnvVar OS environment
--- a/acme/http.go
+++ b/acme/http.go
@@ -64,13 +64,13 @@ func initCertPool() *x509.CertPool {
 	if customCACertsPath := os.Getenv(caCertificatesEnvVar); customCACertsPath != "" {
 		customCAs, err := ioutil.ReadFile(customCACertsPath)
 		if err != nil {
-			panic(fmt.Sprintf("error reading %s=%q: %s",
-				caCertificatesEnvVar, customCACertsPath, err.Error()))
+			panic(fmt.Sprintf("error reading %s=%q: %v",
+				caCertificatesEnvVar, customCACertsPath, err))
 		}
 		certPool := x509.NewCertPool()
 		if ok := certPool.AppendCertsFromPEM(customCAs); !ok {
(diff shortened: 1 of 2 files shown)


acme/http.go
```diff
@@ -36,8 +33,55 @@ const (
...
+		},
+	}
+}
+
+// initCertPool creates a *x509.CertPool populated with the PEM certificate
+// found in the filepath specified in the caCertificateEnvVar OS environment
+// variable. If the caCertificateEnvVar is not set, or there is an error
+// creating a *x509.CertPool from the provided value then nil is returned and
+// the default Golang CertPool is used.
+func initCertPool() *x509.CertPool {
```
cpu: My approach to error handling with this function was to fall-back to the system certpool by returning `nil`. This means providing an invalid env var arg will silently fall back to the system trust store and probably produce an HTTPS validation error on first connect to the ACME server. That seemed like the best way forward but if you would rather see a `panic` or a warning message logged that would be fine with me :-)

acme/http.go
```diff
@@ -1,33 +1,30 @@
...
-	},
-}
+var (
+	// UserAgent (if non-empty) will be tacked onto the User-Agent string in requests.
+	UserAgent string
+
+	// HTTPClient is an HTTP client with a reasonable timeout value and
+	// potentially a custom *x509.CertPool based on the caCertificateEnvVar
+	// environment variable (see the `initCertPool` function)
+	HTTPClient http.Client
```
mholt: Why was the initialization of this moved to `init()`?

cpu: Good question. I think I deluded myself into thinking I couldn't use my own `initCertPool` function from where `HTTPClient` was initialized before. Removed in ee12bff

mholt: Cool, no problem! Just wanted to double-check. :)

acme/http.go
wyattjoh: Considering that the developer specifically provides the environment variable to add a CA, I would honestly rather see it panic/fail when loading/parsing of the designated file fails. The error you'd get would just say "not trusted", but it wouldn't be clear that it was because loading the file failed.

mholt: Yeah, I think an error must always at least be logged. I think it should be returned or logged (or a panic, although I'm less fond about panics due to env var misconfigurations).

cpu: There's a few different options thrown out in this thread. Is there one that everyone would be happy with?

It sounds like @mholt would like to avoid a panic but always have something logged while @wyattjoh , it sounds like you want to see panic/fail. Would using `log.Fatal` in the failure cases of `initCertPool` be acceptable to both of you?

> I think an error must always at least be logged. I think it should be returned or logged

@mholt If `initCertPool` returns an error I think I'll have to undo ee12bff and put the initialization of the `HTTPClient` somewhere where I can unpack the result and potential error from `initCertPool` before assigning the `TLSClientConfig`'s `RootCAs`.

>  (or a panic, although I'm less fond about panics due to env var misconfigurations).

Speaking of env var misconfigurations, it also occurred to me that maybe using `CA_CERTIFICATES` as the env var name is too generic and more likely to collide with other software. I think it would be sensible to change that to `LEGO_CA_CERTIFICATES` while we're iterating on this branch anyway.

mholt: +1 on the more specific env var. In that case, a panic is probably fine, since an environment variable named like that which was set wrong would definitely not be an accident.

cpu: Ok, sounds like a more specific env var and a `panic` would satisfy everyone. :+1:

acme/http.go
```diff
@@ -36,8 +47,36 @@ const (
...
+// found in the filepath specified in the caCertificatesEnvVar OS environment
+// variable. If the caCertificatesEnvVar is not set then initCertPool will
+// return nil. If there is an error creating a *x509.CertPool from the provided
+// caCertificatesEnvVar value then initCertPool will panic.
+func initCertPool() *x509.CertPool {
+	if customCACertsPath := os.Getenv(caCertificatesEnvVar); customCACertsPath != "" {
+		customCAs, err := ioutil.ReadFile(customCACertsPath)
+		if err != nil {
+			panic(fmt.Sprintf("error reading %s=%q: %s",
+				caCertificatesEnvVar, customCACertsPath, err.Error()))
```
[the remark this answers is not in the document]
cpu: @ldez Sure, fixed this and the other `%s` error print instances in 50d897b

acme/http_test.go
```diff
@@ -98,3 +100,110 @@ func TestUserAgent(t *testing.T) {
...
 	}
 }
+
+// TestInitCertPool tests the http.go initCertPool function for customizing the
+// HTTP Client *x509.CertPool with an environment variable.
+func TestInitCertPool(t *testing.T) {
+	// writeTemp creates a temp file with the given contents & prefix and returns
+	// the file path. If an error occurs, t.Fatalf is called to end the test run.
+	writeTemp := func(t *testing.T, contents, prefix string) string {
+		tmpFile, err := ioutil.TempFile("", prefix)
```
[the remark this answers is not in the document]
cpu: @idez Good idea. Added in d12e162, thanks!

cpu:
@ldez Is there anything I can do to help get this merged? Does it need another maintainer +1? It would be nice to free the mental reference I'm carrying :-)

ldez:
I was just waiting in case someone wanted to add something.
Now I think it's time to merge :smile: 

Thank you very much.

cpu:
@ldez Great, thank you! 

@ldez @xenolf @mholt @wyattjoh I don't often get a chance to throw code at the client side of ACME. It was a pleasant experience contributing to Lego. Thanks for the great project/community!

PR #572, opened 2018-06-09
TLS-ALPN-01 Challenge

wyattjoh:
# What does this do?

Adds support for the draft [TLS-ALPN-01](https://tools.ietf.org/html/draft-ietf-acme-tls-alpn-01) challenge.

# How do I test?

1. [Download, install, and run the latest pebble.](https://github.com/letsencrypt/pebble#install)
2. Until #571 is merged, you have to manually patch lego to allow it to interact with pebble's self signed certificates:

```diff
diff --git a/acme/http.go b/acme/http.go
index f45e7f8..ab10f88 100644
--- a/acme/http.go
+++ b/acme/http.go
@@ -1,6 +1,7 @@
 package acme
 
 import (
+	"crypto/tls"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -26,6 +27,9 @@ var HTTPClient = http.Client{
 		TLSHandshakeTimeout:   15 * time.Second,
 		ResponseHeaderTimeout: 15 * time.Second,
 		ExpectContinueTimeout: 1 * time.Second,
+		TLSClientConfig: &tls.Config{
+			InsecureSkipVerify: true,
+		},
 	},
 }
```
3. Add `127.0.0.1 example.com` to your `/etc/hosts` file.
3. Build, test, and run lego with the following:

```bash
# test and build lego
make test build

# attempt to get a certificate via TLS-ALPN-01
./lego -s https://localhost:14000/dir --email=me@example.com --exclude http-01 --exclude dns-01 -d example.com --tls :5001 run
```

**Note**: You may have issues on a Mac if you have the firewall turned on. For the duration of the test, turn these off to ensure that pebble/lego can communicate. (just be sure to turn it back on!)

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,9 +76,6 @@ NAME:
 USAGE:
    lego [global options] command [command options] [arguments...]
 
-VERSION:
-   0.4.1
-
 COMMANDS:
      run      Register an account, then create and install a certificate
      revoke   Revoke a certificate
@@ -92,12 +89,16 @@ GLOBAL OPTIONS:
    --server value, -s value    CA hostname (and optionally :port). The server certificate must be trusted in order to avoid further modifications to the client. (default: "https://acme-v02.api.letsencrypt.org/directory")
    --email value, -m value     Email used for registration and recovery contact.
    --accept-tos, -a            By setting this flag to true you indicate that you accept the current Let's Encrypt terms of service.
+   --eab                       Use External Account Binding for account registration. Requires --kid and --hmac.
+   --kid value                 Key identifier from External CA. Used for External Account Binding.
+   --hmac value                MAC key from External CA. Should be in Base64 URL Encoding without padding format. Used for External Account Binding.
    --key-type value, -k value  Key type to use for private keys. Supported: rsa2048, rsa4096, rsa8192, ec256, ec384 (default: "rsa2048")
-   --path value                Directory to use for storing the data (default: "/.lego")
+   --path value                Directory to use for storing the data (default: "/Users/wyattjohnson/Go/src/github.com/xenolf/lego/.lego")
    --exclude value, -x value   Explicitly disallow solvers by name from being used. Solvers: "http-01", "dns-01", "tls-alpn-01".
    --webroot value             Set the webroot folder to use for HTTP based challenges to write directly in a file in .well-known/acme-challenge
    --memcached-host value      Set the memcached host(s) to use for HTTP based challenges. Challenges will be written to all specified hosts.
    --http value                Set the port and interface to use for HTTP based challenges to listen on. Supported: interface:port or :port
+   --tls value                 Set the port and interface to use for TLS based challenges to listen on. Supported: interface:port or :port
    --dns value                 Solve a DNS challenge using the specified provider. Disables all other challenges. Run 'lego dnshelp' for help on usage.
    --http-timeout value        Set the HTTP timeout value to a specific value in seconds. The default is 10 seconds. (default: 0)
    --dns-timeout value         Set the DNS timeout value to a specific value in seconds. The default is 10 seconds. (default: 0)
--- a/README.md
+++ b/README.md
@@ -56,6 +56,7 @@ Otherwise the release will be tagged with the `dev` version identifier.
 - Robust implementation of all ACME challenges
   - HTTP (http-01)
   - DNS (dns-01)
+  - TLS (tls-alpn-01)
 - SAN certificate support
 - Comes with multiple optional [DNS providers](https://github.com/xenolf/lego/tree/master/providers/dns)
 - [Custom challenge solvers](https://github.com/xenolf/lego/wiki/Writing-a-Challenge-Solver)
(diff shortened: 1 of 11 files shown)


acme/tls_alpn_challenge_server.go
```diff
@@ -0,0 +1,79 @@
...
+	return &TLSALPNProviderServer{iface: iface, port: port}
+}
+
+// Present generates a certificate with a SHA-256 digest of the keyAuth provided
+// as the acmeValidation-v1 extension value to conform to the ACME-TLS-ALPN
+// spec.
+func (t *TLSALPNProviderServer) Present(domain, token, keyAuth string) error {
+	if t.port == "" {
+		// Fallback to port 443 if the port was not provided.
+		t.port = "443"
```
mholt: Define a const, as this is the port required by the spec.

acme/tls_alpn_challenge_server.go
```diff
@@ -0,0 +1,79 @@
...
+	tlsConf.NextProtos = []string{acmeTLS1Protocol}
+
+	// Create the listener with the created tls.Config.
+	t.listener, err = tls.Listen("tcp", net.JoinHostPort(t.iface, t.port), tlsConf)
+	if err != nil {
+		return fmt.Errorf("Could not start HTTPS server for challenge -> %v", err)
+	}
+
+	// Shut the server down when we're finished.
+	t.done = make(chan bool)
```
mholt: If this done channel has a purpose, it is not obvious to me... what is it for, exactly?

acme/tls_alpn_challenge_server.go
```diff
@@ -0,0 +1,79 @@
...
+		http.Serve(t.listener, nil)
+		t.done <- true
+	}()
+
+	return nil
+}
+
+// CleanUp closes the HTTPS server.
+func (t *TLSALPNProviderServer) CleanUp(domain, token, keyAuth string) error {
+	if t.listener == nil {
```
mholt: Excellent, checking for a nil value just in case

acme/tls_alpn_challenge_server.go
xenolf: It's for shutting down the listener after the challenge has been served.

mholt: Hmm, I think I must be missing something then. The channel is never closed, and nothing ever blocks/reads on it, except `CleanUp()` which has already closed the listener by then (see below, line ~77). Since we don't write to this channel after the `Serve()` method returns, I'm not sure how much help the channel is. Am I reading it wrong?

wyattjoh: You aren't! I based 99% of this off of the existing TLS/SNI code, hence why that propagated. As long as `http.Serve` blocks until it's closed, we're great. From my tests, it worked great.

mholt:
Awesome -- I gave my approval, but this shouldn't be merged into master until Let's Encrypt supports tls-alpn in production.

An alternative is that we could take this challenge out of the default rotation so that it has to be explicitly enabled, until Let's Encrypt has deployed it to production. That way we can merge anytime.

README.md
```diff
@@ -91,12 +89,16 @@ GLOBAL OPTIONS:
...
    --email value, -m value     Email used for registration and recovery contact.
    --accept-tos, -a            By setting this flag to true you indicate that you accept the current Let's Encrypt terms of service.
+   --eab                       Use External Account Binding for account registration. Requires --kid and --hmac.
+   --kid value                 Key identifier from External CA. Used for External Account Binding.
+   --hmac value                MAC key from External CA. Should be in Base64 URL Encoding without padding format. Used for External Account Binding.
    --key-type value, -k value  Key type to use for private keys. Supported: rsa2048, rsa4096, rsa8192, ec256, ec384 (default: "rsa2048")
-   --path value                Directory to use for storing the data (default: "/.lego")
-   --exclude value, -x value   Explicitly disallow solvers by name from being used. Solvers: "http-01", "dns-01",.
+   --path value                Directory to use for storing the data (default: "/Users/wyattjohnson/Go/src/github.com/xenolf/lego/.lego")
```
ldez: `default: "/.lego"`

acme/tls_alpn_challenge_test.go
```diff
@@ -0,0 +1,87 @@
...
+	domain := "localhost:23457"
+	privKey, _ := rsa.GenerateKey(rand.Reader, 512)
+	j := &jws{privKey: privKey}
+	clientChallenge := challenge{Type: string(TLSALPN01), Token: "tlsalpn1"}
+	mockValidate := func(_ *jws, _, _ string, chlng challenge) error {
+		conn, err := tls.Dial("tcp", domain, &tls.Config{
+			InsecureSkipVerify: true,
+		})
+		if err != nil {
+			t.Errorf("Expected to connect to challenge server without an error. %s", err.Error())
```
ldez:
```go
t.Errorf("Expected to connect to challenge server without an error. %v", err)

Comment (wyattjoh) [reply acme/tls_alpn_challenge_test.go]: Why is this preferred?

Comment (mholt) [reply acme/tls_alpn_challenge_test.go]: In case the error value is not a Stringer. Like, if it is a regular struct. It's just more robust.

Comment (wyattjoh) [reply acme/tls_alpn_challenge_test.go]: I can make the change, for sure, but `error` is just an interface for:

```go
type error interface {
  Error() string
}
```

So technically, every error, when called with `.Error()` will return a string, else it wouldn't compile.

acme/tls_alpn_challenge.go
```diff
@@ -0,0 +1,90 @@
...
+	zBytes := sha256.Sum256([]byte(keyAuth))
+
+	// Add the keyAuth digest as the acmeValidation-v1 extension (marked as
+	// critical such that it won't be used by non-ACME software). Reference:
+	// https://tools.ietf.org/html/draft-ietf-acme-tls-alpn-01#section-3
+	extensions := []pkix.Extension{
+		{
+			Id:       idPeAcmeIdentifierV1,
+			Critical: true,
+			Value:    zBytes[:sha256.Size],
```
rolandshoemaker: The hash should be wrapped as a ASN.1 OCTET STRING to properly comply with the draft. This can be done by using `asn1.Marshal(zBytes[:sha256.Size])`.

mholt: Ooo, good catch. Thanks for the review Roland!

mholt:
On second thought, the ACME protocol should only choose a challenge type that is mutually supported, so we should be able to merge this in at any time without adverse effects.

acme/tls_alpn_challenge.go
rolandshoemaker: Np, otherwise basically LGTM.

PR #575, opened 2018-06-11
Add Hurricane Electric DNS provider (dns.he.net)

johnnie502:
### Changes
* Added Hurricane Electric DNS provider

### Test
* Test added
* Live test passed


providers/dns/dns_providers.go
```diff
@@ -78,6 +79,8 @@ func NewDNSChallengeProviderByName(name string) (acme.ChallengeProvider, error)
 		provider, err = gcloud.NewDNSProvider()
 	case "godaddy":
 		provider, err = godaddy.NewDNSProvider()
+	case "he":
```
mholt: I know that HE may be a common way to refer to the company, but can we use something a little more verbose for this string _and_ the package name? Short identifiers are often used as variable names. How about `hurricanedns`.

johnnie502: Thank for your suggestion, it makes sense. If we go that route, should we also rename the package name etc?

The CI failed because library `goquery` in missing in dep. I am not very familiar with dep (and this is my first commit for golang project too) Would you please tell me how to fix that?

mholt: >  If we go that route, should we also rename the package name etc?

Yes, call the package by the same name.

> The CI failed because library goquery in missing in dep. I am not very familiar with dep (and this is my first commit for golang project too) Would you please tell me how to fix that?

/cc @ldez @wyattjoh

ldez: @johnnie502 
- Install [dep](https://github.com/golang/dep)
- then do `dep ensure -v`

ldez: could you also rename the package to `hurricanedns`

johnnie502: @idez will do

providers/dns/hurricanedns/hurricanedns.go
```diff
@@ -0,0 +1,221 @@
...
+}
+
+// NewDNSProvider returns a DNSProvider instance configured for Hurricane Electric DNS.
+// Credentials must be passed in the environment variables: HE_DOMAIN_NAME,
+// HE_USERNAME, HE_PASSWORD.
+func NewDNSProvider() (*DNSProvider, error) {
+	domainName := os.Getenv("HE_DOMAIN_NAME")
+	userName := os.Getenv("HE_USERNAME")
+	password := os.Getenv("HE_PASSWORD")
+	return NewDNSProviderCredentials(domainName, userName, password)
```
ldez: Could rewrite like the following example:
https://github.com/xenolf/lego/blob/c4bbb4b819c087d02066fe745d7ca5b60a480d36/providers/dns/cloudflare/cloudflare.go#L32-L39

providers/dns/hurricanedns/hurricanedns.go
```diff
@@ -0,0 +1,221 @@
...
+
+	return err
+}
+
+// Present creates a TXT record using the specified parameters
+func (d *DNSProvider) Present(domain, token, keyAuth string) error {
+	fqdn, value, ttl := acme.DNS01Record(domain, keyAuth)
+
+	if ttl < 300 {
+		ttl = 300 // 300 is hurricanedns minimum value for ttl
```
ldez: Do you have some documentation about this limitation?

providers/dns/hurricanedns/hurricanedns_test.go
```diff
@@ -0,0 +1,60 @@
...
+func TestNewDNSProvider(t *testing.T) {
+	provider, err := NewDNSProvider()
+
+	if !heLiveTest {
+		assert.Error(t, err)
+	} else {
+		assert.NotNil(t, provider)
+		assert.NoError(t, err)
+	}
+}
```
ldez: Could add a test as the following example:

https://github.com/xenolf/lego/blob/c4bbb4b819c087d02066fe745d7ca5b60a480d36/providers/dns/cloudflare/cloudflare_test.go#L51-L58

providers/dns/hurricanedns/hurricanedns.go
```diff
@@ -0,0 +1,221 @@
...
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"os"
+
+	"github.com/PuerkitoBio/goquery"
+	"github.com/xenolf/lego/acme"
+	"net/url"
+	"regexp"
+)
```
ldez: Could you group standard imports, and sort import alphabetically.
```go
import (
	"bytes"
	"fmt"
	"io/ioutil"
	"net/http"
	"net/url"
	"os"
	"regexp"

	"github.com/PuerkitoBio/goquery"
	"github.com/xenolf/lego/acme"
)
```

johnnie502:
@ldez I agree with the method that posting to a form and resolving the result from the web page looks unstable, however this seems the only way to automatically update DNS record for he.net, they don't have an API for this :disappointed:

ldez:
have you looked [here](https://dns.he.net/), there seems to be an API

johnnie502:
Yes, I used it a lot, the only thing looks like an API Doc is how to update DDNS via curl https://dns.he.net/docs.html
Anything else need to be done via the web UI that you just posted.

providers/dns/hurricanedns/hurricanedns.go
johnnie502: There is no official documentation for that, this is based on the assumption from their web ui at https://dns.he.net

johnnie502:
@ldez I have to log out from their web ui to see the text you quoted, which sound promising. The problem is it seems they still haven't released that feature yet.

johnnie502:
And I apologize for those failing builds on the CI, this is my first time to working on a golang project. And I just learned I should run `make` locally before push to the PR.

PR #576, opened 2018-06-13
Fix no-bundle flag

stintel:
The ACME library incorrectly assumes that an "up" link to the issuer
certificate is always provided.

From the ACME protocol specification:
`The server MAY provide one or more link relation header fields {{RFC5988}}
with relation "alternate".`

Because of this, the no-bundle flag doesn't always work. Fix this by
splitting the received certificate chain on the empty newline in case
the "up" link is not provided.

Signed-off-by: Stijn Tintel <stijn@linux-ipv6.be>

### DIFF
--- a/acme/client.go
+++ b/acme/client.go
@@ -2,6 +2,7 @@
 package acme
 
 import (
+	"bytes"
 	"crypto"
 	"crypto/x509"
 	"encoding/base64"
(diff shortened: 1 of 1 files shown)


stintel:
Fix that warning. I'll have to figure out how to add a test.

ldez:
Seems related to #587

stintel:
Fixed by #587

PR #579, opened 2018-06-15
tls-alpn: add a function to return PEM blocks.

ldez:
Due to the difficulty of serializing/deserializing `tls.Certificate`.

Add a function to return PEM blocks.

### DIFF
--- a/acme/tls_alpn_challenge_server.go
+++ b/acme/tls_alpn_challenge_server.go
@@ -8,8 +8,8 @@ import (
 )
 
 const (
-	// AcmeTLS1Protocol is the ALPN Protocol ID for the ACME-TLS/1 Protocol.
-	AcmeTLS1Protocol = "acme-tls/1"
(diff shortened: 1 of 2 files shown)


acme/tls_alpn_challenge_server.go
```diff
@@ -8,8 +8,8 @@ import (
 )
 
 const (
-	// acmeTLS1Protocol is the ALPN Protocol ID for the ACME-TLS/1 Protocol.
-	acmeTLS1Protocol = "acme-tls/1"
+	// AcmeTLS1Protocol is the ALPN Protocol ID for the ACME-TLS/1 Protocol.
+	AcmeTLS1Protocol = "acme-tls/1"
```
mholt: Change case to `ACMETLS1Protocol`

ldez: done

PR #580, opened 2018-06-16
Review DNS providers

ldez:
- create `log.Infof` and `log.Warnf`
- use one `http.Client` by provider instead of one client by request
- use the same receiver name `d` for all `DNSProvider`
- use `http.MethodXXX`


log/logger.go
```diff
@@ -8,52 +8,58 @@ import (
 // Logger is an optional custom logger.
 var Logger *log.Logger
 
+func initDefaultLogger() {
```
wyattjoh: Why not move this into the main body instead? It still gives importing packages the ability to override it, and it stays more consistent to how the base logger for Go actually is implemented.

```go
var Logger = log.New(os.Stderr, "", log.LstdFlags)
```

providers/dns/pdns/pdns.go
```diff
@@ -52,33 +54,39 @@ func NewDNSProviderCredentials(host *url.URL, key string) (*DNSProvider, error)
...
+	d := &DNSProvider{
 		host:   host,
 		apiKey: key,
+		client: &http.Client{Timeout: 30 * time.Second},
 	}
-	provider.getAPIVersion()
 
-	return provider, nil
+	var err error
+	d.apiVersion, err = d.getAPIVersion()
```
wyattjoh: I almost prefer to see something like:

```go
apiVersion, err := d.getAPIVersion()
if err != nil {
  // ...
}

d.apiVersion = apiVersion
```

As it would only "set" the `apiVersion` when there was no error.

PR #586, opened 2018-06-25
Add EXEC_MODE=RAW support to DNS Provider exec

shupp:
# Overview

This change adds `EXEC_MODE=RAW` support to the exec provider.  When this is set, it does not run _acme.DNS01_ first. It instead passes in the raw domain, token, and keyAuth values to the program. This is so that it can be a wrapper to other existing dns provider packages.

# Motivation

At my company, when a certificate request has multiple domains,  we have to support multiple providers for that certificate.  Since by default the exec provider calls _acme.DNS01_, we can't use the input to the program to then call another existing provider.  This change allows for calling other providers.

# Other approaches

Another approach to support multiple providers per cert would be to support an input mapping of domains to providers (and perhaps even profiles, in the case of route53 for example).


shupp:
Build failure appears to be an unrelated and likely intermittent.

ldez:
Seems like an error with the network or with `cloud.google.com/go`:

> unable to deduce repository and source type for "cloud.google.com/go": unable to read metadata: unable to fetch raw metadata: failed HTTP request to URL "http://cloud.google.com/go?go-get=1": Get https://cloud.google.com/go?go-get=1: net/http: TLS handshake timeout

ldez:
Due to large code duplication between `exec` and `execraw`, I think a single exec provider with an option based on an environment variable can be a better solution.
ex: `EXEC_MODE=RAW`

shupp:
Sounds good.  I'll have an updated PR on Monday.

shupp:
@ldez I updated the PR and rebased against current master.  Ready for another review whenever you have a chance.  Thanks!

shupp:
@ldez thanks for adding EXEC_MODE=RAW support to _cleanUp()_ in d3b7e108b919b26ba5a964e6949abfc9c91ee521, which I neglected to do in my change from execraw provider to exec + EXEC_MODE=RAW.  Sorry about that!

PR #588, opened 2018-06-26
Route53: make provider timeouts configurable

Mongey:
👋 I'm looking to [configure the timeout](https://github.com/vancluever/terraform-provider-acme/issues/55) for the Route53 provider, as I'm hitting the error @ 
https://github.com/xenolf/lego/blob/master/acme/utils.go#L15

```
[*********] Time limit exceeded. Last error: NS ns-71.awsdns-10.net. did not return the expected TXT record
```
This provides a way to pass in a different configuration for the provider, rather than using the hardcoded values. This keeps the current defaults.

### DIFF
--- a/providers/dns/route53/route53.go
+++ b/providers/dns/route53/route53.go
@@ -17,15 +17,19 @@ import (
 	"github.com/xenolf/lego/acme"
 )
 
-const (
-	maxRetries = 5
-	route53TTL = 10
-)
+// Config is used to configure the creation of the DNSProvider
+type Config struct {
+	MaxRetries         int
+	Route53TTL         int
+	PropagationTimeout time.Duration
+	PollingInterval    time.Duration
+}
 
 // DNSProvider implements the acme.ChallengeProvider interface
 type DNSProvider struct {
 	client       *route53.Route53
 	hostedZoneID string
+	config       *Config
 }
 
 // customRetryer implements the client.Retryer interface by composing the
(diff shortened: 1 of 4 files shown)


Mongey:
👋 @ldez, is this something you'd consider merging ?

ianks:
I am also experiencing timeouts with gcloud

```
acme: Error 400 - urn:ietf:params:acme:error:dns - DNS problem: NXDOMAIN looking up TXT
```

Would this patch help with that?

ldez:
@ianks it's not related: `NXDOMAIN` means Non eXistant Domain.
Check your DNS and fush your DNS cache.

providers/dns/route53/route53.go
```diff
@@ -64,36 +85,38 @@ func (d customRetryer) RetryRules(r *request.Request) time.Duration {
 // public hosted zone via the FQDN.
 //
 // See also: https://github.com/aws/aws-sdk-go/wiki/configuring-sdk
-func NewDNSProvider() (*DNSProvider, error) {
-	hostedZoneID := os.Getenv("AWS_HOSTED_ZONE_ID")
+func NewDNSProvider(config *Config) (*DNSProvider, error) {
+	if config == nil {
+		config = NewDefaultConfig()
```
wyattjoh: I'd almost feel better if this did not perform a nil check here, and instead required the config to be passed. In `providers/dns/dns_providers.go`, it could just go `route53.NewDNSProvider(route53.NewDefaultConfig())`

PR #589, opened 2018-06-26
Don't trust identifiers order.

ldez:
[ACME draft Section 7.4 "Applying for Certificate Issuance"](https://tools.ietf.org/html/draft-ietf-acme-acme-12#section-7.4)
> Clients SHOULD NOT make any assumptions about the sort order of
> "identifiers" or "authorizations" elements in the returned order
> object.

Pebble respect this assumption and this produce some weird certificate where the CN is the not the CN but a random SAN.

### DIFF
--- a/acme/client.go
+++ b/acme/client.go
@@ -662,9 +662,18 @@ func (c *Client) requestCertificateForOrder(order orderResource, bundle bool, pr
 
 	// determine certificate name(s) based on the authorization resources
 	commonName := order.Domains[0]
-	var san []string
+
+	// ACME draft Section 7.4 "Applying for Certificate Issuance"
+	// https://tools.ietf.org/html/draft-ietf-acme-acme-12#section-7.4
+	// says:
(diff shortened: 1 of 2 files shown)


acme/client.go
```diff
@@ -662,9 +662,18 @@ func (c *Client) requestCertificateForOrder(order orderResource, bundle bool, pr
...
+	// ACME draft Section 7.4 "Applying for Certificate Issuance"
+	// https://tools.ietf.org/html/draft-ietf-acme-acme-12#section-7.4
+	// says:
+	//   Clients SHOULD NOT make any assumptions about the sort order of
+	//   "identifiers" or "authorizations" elements in the returned order
+	//   object.
+	san := []string{commonName}
 	for _, auth := range order.Identifiers {
-		san = append(san, auth.Value)
+		if auth.Value != commonName {
```
wyattjoh: Why are you excluding `commonName`? Could you provide a comment as to why?

ldez: Because I add the common name at before to have the good order.

PR #590, opened 2018-06-27
CA Server Name.

ldez:
Allow to specify a CA server name (`LEGO_CA_SERVER_NAME`).

This a requirement when you are using Pebble with an external DNS server.

Related to https://github.com/letsencrypt/pebble/pull/139


ldez:
This is a developer-only feature like `LEGO_CA_CERTIFICATES`.

Are you sure you want to add this in the readme?

wyattjoh:
I suppose it's not necessary. Just leaving a note that starting a `DEVELOPMENT.md` or something to describe how to develop against/with lego would be a good solution.

PR #605, opened 2018-07-24
New provider to support complicated multi-dns-provider setups

captncraig:
This adds a "multi" provider that dynamically creates and selects dns providers to use on a domain-by-domain basis. I need this for our production setup because:

1. We need a single certificate that spans many domains, including multiple wildcards.
2. Most domains are hosted on multiple dns providers (google cloud and r53, both active all the time).
3. Some domains are hosted on different providers from others.
4. We even have multiple cloudflare accounts with different feature sets that have different domains on them.

Using the "multi" provider, you supply an addition json configuration that specifies which providers are available with their credentials, and which providers should be applied to which domains.

This can be a little complicated to specify, but it seems necessary if your setup is as convoluted as described above.

Godoc documentation should describe pretty well how to use it. 

Tested working in some pretty complicated scenarios for production-level certs.

### DIFF
--- a/providers/dns/dns_providers.go
+++ b/providers/dns/dns_providers.go
@@ -122,7 +122,7 @@ func NewDNSChallengeProviderByName(name string) (acme.ChallengeProvider, error) 
 	case "manual":
 		return acme.NewDNSProviderManual()
 	case "multi":
-		return multi.New()
+		return multi.NewDNSProvider()
 	case "mydnsjp":
 		return mydnsjp.NewDNSProvider()
 	case "namecheap":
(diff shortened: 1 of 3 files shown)


ldez:
@captncraig could you rebase?

Daegalus:
Any updates on this? it is becoming clear that I might need this in our platform too.

captncraig:
Sorry, I got sidetracked on this and we decided to use our existing dns tool for this as well. I think this feature is still an ok idea, but am reluctant to continue work on it as the config can be a bit tricky, and I can't commit to really offering support going forward. The code is there though. If somebody wants to take it up, rebase and get it merge-ready please do.

ldez:
I already start to work on this, but for now I don't found a satisfying approach.

PR #607, opened 2018-07-25
Submit all dns records up front, then validate serially

captncraig:
This implements a simple way to increase the speed and reliability of dns challenges in cases where there are multiple challenges needed. It does not parallelize anything, it just creates all of the txt records in a preliminary pass so that things can propagate while we perform validations on other records. By the time the first record is resolving and validated, hopefully the others will be as well, and the overall time to issue a certificate can be greatly reduced.

The implementation simply involves an additional `presolver` interface implemented only by the dns solver so far. The intent of the `PreSolve` method is to allow preliminary work to happen in the first phase, and then validation can happen in the second `Solve` phase. If a `solver` does not implement `presolver`, the preliminary phase will just not be run.

### DIFF
--- a/acme/client.go
+++ b/acme/client.go
@@ -47,6 +47,11 @@ type presolver interface {
 	PreSolve(challenge challenge, domain string) error
 }
 
+// Interface for challenges like dns, where we can solve all the challenges before to delete them.
+type cleanup interface {
+	CleanUp(challenge challenge, domain string) error
+}
+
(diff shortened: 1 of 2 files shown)


ldez:
@captncraig could you rebase?

nmengin:
Hello @captncraig ,

I guess your PR can solve lot of problems, I just proposed modifications to manage a specific use case : https://github.com/captncraig/lego/pull/1

PR #629, opened 2018-09-08
CloudXNS: client isolation.

### LINKED ISSUE #619: Wrong host record in CloudXNS
When I use [Caddy](https://caddyserver.com/) as an HTTP server in DNS challenge mode, it cannot get certificate of my domain hosted by [CloudXNS](https://www.cloudxns.net/).

It seems that there is something wrong with record name generated by Lego.
The host record is set to `_acme-challenge.example.com`.
qq20180816-175531 2x

But CloudXNS will resolve it to `_acme-challenge.example.com.example.com`.
qq20180816-174143 2x

According to [the document of CloudXNS](https://www.cloudxns.net/Public/Doc/CloudXNS_api2.0_doc_zh-cn.zip) (Chinese version only, sad...), the key `host` **SHOULD NOT** contain the domain again.
For example `_acme-challenge` means `_acme-challenge.example.com` and `_acme-challenge.foo` means `_acme-challenge.foo.example.com`.

Since I don't know GO, I cannot bring a PR. But I think the problem is here:
[https://github.com/xenolf/lego/blob/master/providers/dns/cloudxns/cloudxns.go#L139](https://github.com/xenolf/lego/blob/master/providers/dns/cloudxns/cloudxns.go#L139)

Hoping you can fix it. Thanks a lot.

ldez:
Client isolation to facilitate testing.

Fixes #619

### DIFF
--- a/providers/dns/cloudxns/client.go
+++ b/providers/dns/cloudxns/client.go
@@ -53,19 +53,19 @@ func NewClient(apiKey string, secretKey string) (*Client, error) {
 	}
 
 	return &Client{
-		apiKey:    apiKey,
-		secretKey: secretKey,
-		HttClient: &http.Client{},
-		BaseURL:   defaultBaseURL,
+		apiKey:     apiKey,
+		secretKey:  secretKey,
+		HTTPClient: &http.Client{},
+		BaseURL:    defaultBaseURL,
 	}, nil
 }
 
 // Client CloudXNS client
 type Client struct {
-	apiKey    string
-	secretKey string
-	HttClient *http.Client
-	BaseURL   string
+	apiKey     string
+	secretKey  string
+	HTTPClient *http.Client
+	BaseURL    string
 }
 
 // GetDomainInformation Get domain name information for a FQDN
@@ -156,7 +156,7 @@ func (c *Client) doRequest(method, uri string, body []byte) (json.RawMessage, er
 		return nil, err
 	}
 
-	resp, err := c.HttClient.Do(req)
+	resp, err := c.HTTPClient.Do(req)
 	if err != nil {
 		return nil, fmt.Errorf("CloudXNS: %v", err)
 	}
--- a/providers/dns/cloudxns/client.go
+++ b/providers/dns/cloudxns/client.go
@@ -1,0 +1,208 @@
+package cloudxns
+
+import (
+	"bytes"
+	"crypto/md5"
+	"encoding/hex"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/xenolf/lego/acme"
+)
+
+const defaultBaseURL = "https://www.cloudxns.net/api2/"
+
+type apiResponse struct {
+	Code    int             `json:"code"`
+	Message string          `json:"message"`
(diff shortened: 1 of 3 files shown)


providers/dns/cloudxns/cloudxns.go
```diff
@@ -1,32 +1,45 @@
...
+	APIKey     string
+	SecretKey  string
+	TTL        int
+	HTTPClient *http.Client
+}
+
+// NewDefaultConfig returns a default configuration for the DNSProvider
+func NewDefaultConfig() *Config {
+	client := acme.HTTPClient
+	client.Timeout = time.Second * time.Duration(env.GetOrDefaultInt("CLOUDXNS_HTTP_TIMEOUT", 30))
```
wyattjoh: Is this not modifying the default/package level client?

providers/dns/cloudxns/client.go
```diff
@@ -0,0 +1,208 @@
...
+		HttClient: &http.Client{},
+		BaseURL:   defaultBaseURL,
+	}, nil
+}
+
+// Client CloudXNS client
+type Client struct {
+	apiKey    string
+	secretKey string
+	HttClient *http.Client
```
wyattjoh: Is there a reason this is `HttClient` vs `HTTPClient`?

providers/dns/cloudxns/client.go
```diff
@@ -0,0 +1,208 @@
...
+	req.Header.Set("API-KEY", c.apiKey)
+	req.Header.Set("API-REQUEST-DATE", requestDate)
+	req.Header.Set("API-HMAC", c.hmac(url, requestDate, string(body)))
+	req.Header.Set("API-FORMAT", "json")
+
+	return req, nil
+}
+
+func (c *Client) hmac(url, date, body string) string {
+	sum := md5.Sum([]byte(c.apiKey + url + body + date + c.secretKey))
```
wyattjoh: Could use a comment to explain what's going on here, a bit cryptic.

providers/dns/cloudxns/client_test.go
```diff
@@ -0,0 +1,283 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func handlerMock(method string, response *apiResponse, data interface{}) http.Handler {
+	return http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
+		if req.Method != method {
+			content, err := json.Marshal(apiResponse{
+				Code:    666,
```
wyattjoh: 666? Needs comments as to why this is the code.

providers/dns/cloudxns/client.go
ldez: It's not my code: I just move the code from `DNSProvider`

providers/dns/cloudxns/client.go
ldez: good catch

providers/dns/cloudxns/cloudxns.go
ldez: no because I use a copy of `acme.HTTPClient`

providers/dns/cloudxns/client_test.go
ldez: It's just a dumb code

PR #633, opened 2018-09-11
Use path/filepath instead of path

mattn:
The path is a package to manipulate pseudo paths. use path/filepath for physical paths. For example, backslash and slash is contained both in `--path` usage.


mattn:
Thank you

PR #634, opened 2018-09-12
Allow to configure TTL, interval and timeout

ldez:
Allow to configure TTL, interval or timeout (if they are available) with env var or with a DNSProvider constructor:
- `XXX_TTL`
- `XXX_PROPAGATION_TIMEOUT`
- `XXX_POLLING_INTERVAL`
- `XXX_HTTP_TIMEOUT`

Related to #474, #475

### DIFF
--- a/providers/dns/alidns/alidns.go
+++ b/providers/dns/alidns/alidns.go
@@ -3,10 +3,14 @@
 package alidns
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"strings"
+	"time"
 
+	"github.com/aliyun/alibaba-cloud-sdk-go/sdk"
+	"github.com/aliyun/alibaba-cloud-sdk-go/sdk/auth/credentials"
 	"github.com/aliyun/alibaba-cloud-sdk-go/sdk/requests"
 	"github.com/aliyun/alibaba-cloud-sdk-go/services/alidns"
 	"github.com/xenolf/lego/acme"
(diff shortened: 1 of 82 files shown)


providers/dns/alidns/alidns.go
```diff
@@ -25,48 +51,74 @@ type DNSProvider struct {
...
 	}
 
-	client, err := alidns.NewClientWithAccessKey(regionID, apiKey, secretKey)
+	conf := sdk.NewConfig().WithTimeout(config.HTTPTimeout)
+	credential := credentials.NewAccessKeyCredential(config.APIKey, config.SecretKey)
+
+	client, err := alidns.NewClientWithOptions(config.RegionID, conf, credential)
 	if err != nil {
-		return nil, fmt.Errorf("AliDNS: credentials failed: %v", err)
+		return nil, fmt.Errorf("alicloud: credentials failed: %v", err)
```
wyattjoh: Another vote for replacing these with `github.com/pkg/errors.Wrap`?

providers/dns/azure/azure.go
```diff
@@ -91,59 +121,55 @@ func (d *DNSProvider) Present(domain, token, keyAuth string) error {
...
-// Returns the relative record to the domain
-func toRelativeRecord(domain, zone string) string {
-	return acme.UnFqdn(strings.TrimSuffix(domain, zone))
+	_, err = rsc.CreateOrUpdate(ctx, d.config.ResourceGroup, zone, relative, dns.TXT, rec, "", "")
+	return fmt.Errorf("azure: %v", err)
 }
 
 // CleanUp removes the TXT record matching the specified parameters
 func (d *DNSProvider) CleanUp(domain, token, keyAuth string) error {
+	ctx := context.Background()
```
wyattjoh: It should accept a context rather than using background.

providers/dns/dnsimple/dnsimple_test.go
```diff
@@ -44,6 +46,8 @@ func TestNewDNSProviderValid(t *testing.T) {
 	defer restoreEnv()
 	os.Setenv("DNSIMPLE_OAUTH_TOKEN", "123")
 
+	acme.UserAgent = "lego"
```
wyattjoh: Why are we settings this in a test?

providers/dns/azure/azure.go
ldez: it's not possible because this method is a part of an interface

providers/dns/alidns/alidns.go
ldez: nope, I don't think we need to use `github.com/pkg/error`

providers/dns/dnsimple/dnsimple_test.go
ldez: because it's the goal of this test: 
```go
assert.Equal(t, "lego", provider.client.UserAgent)
```

wyattjoh: Missed that, sorry!

PR #647, opened 2018-09-24
Cloudflare: support for CF_API_KEY and CF_API_EMAIL

### LINKED ISSUE #492: Allow CF_API_KEY and CF_API_EMAIL as alternative environment variables for Cloudflare
From https://github.com/caddyserver/dnsproviders/issues/22

The cloudflare tools use those environment variable names. If lego can switch, or read from both, then that makes compatibility with those tools easier. :)

https://github.com/cloudflare/python-cloudflare
https://github.com/cloudflare/cloudflare-go/tree/master/cmd/flarectl

ldez:
Fixes #492

### DIFF
--- a/platform/config/env/env_test.go
+++ b/platform/config/env/env_test.go
@@ -9,6 +9,96 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func TestGetWithFallback(t *testing.T) {
+	type expected struct {
+		value map[string]string
+		error string
+	}
+
+	testCases := []struct {
+		desc     string
+		envVars  map[string]string
+		groups   [][]string
+		expected expected
+	}{
+		{
+			desc:   "no groups",
+			groups: nil,
+			expected: expected{
+				value: map[string]string{},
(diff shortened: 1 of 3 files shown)


platform/config/env/env_test.go
```diff
@@ -9,6 +9,96 @@ import (
...
+
+	for _, test := range testCases {
+		t.Run(test.desc, func(t *testing.T) {
+			for key, value := range test.envVars {
+				os.Setenv(key, value)
+			}
+
+			defer func() {
+				for key := range test.envVars {
+					os.Unsetenv(key)
```
wyattjoh: This runs into an interesting situation...

Right now, if I ran the tests with:

```
export TEST_LEGO_H=H
go test ./...
```

It would produce different results. Maybe it's possible to get the env vars before you run the test that you'll be redefining, setting them to the desired ones, running the test, then resetting the variables back again as they were before, thereby preserving the same test results independent of the order that the tests are organized in.

PR #658, opened 2018-10-02
cloudflare: use the official go client.

ldez:
Use the official Cloudfalre go client.

https://github.com/cloudflare/cloudflare-go


providers/dns/cloudflare/cloudflare.go
```diff
@@ -61,37 +67,25 @@ func NewDNSProvider() (*DNSProvider, error) {
...
+
+	client, err := cloudflare.New(config.AuthKey, config.AuthEmail, cloudflare.HTTPClient(config.HTTPClient))
 	if err != nil {
 		return nil, err
 	}
 
-	client.HTTPClient = config.HTTPClient
-
 	// TODO: must be remove. keep only for compatibility reason.
 	client.BaseURL = CloudFlareAPIURL
```
wyattjoh: As we're already breaking the API (by removing the `NewDNSProviderCredentials`, could we possibly remove these as well?

ldez: oops, I will restore `NewDNSProviderCredentials`

mcroach:
I'm unable to build caddyserver with the cloudflare dns provider due to an undefined function error for NewDNSProviderCredentials. 8905dab looks like it was meant to restore that function but is missing in master. Can this be fixed?

PR #684, opened 2018-10-20
Update Azure auth to support instance metadata service

kung-foo:
This PR adds the ability for lego to utilize the [instance metadata service](https://docs.microsoft.com/en-us/azure/virtual-machines/windows/instance-metadata-service) to fetch a bearer token rather than rely on environmental variables.

If the env vars are set, then the behavior should be the same as before.

All auth is now handled via https://github.com/Azure/go-autorest/blob/v10.14.0/autorest/azure/auth/auth.go#L38-L42 which means a lot of the test code was removed since it was specifically testing the reading of environmental variables.


providers/dns/azure/azure.go
```diff
@@ -198,3 +214,33 @@ func (d *DNSProvider) newServicePrincipalToken(scope string) (*adal.ServicePrinc
 func toRelativeRecord(domain, zone string) string {
 	return acme.UnFqdn(strings.TrimSuffix(domain, zone))
 }
+
+// Fetches metadata from environment or he instance metadata service
+// borrowed from https://github.com/Microsoft/azureimds/blob/master/imdssample.go
+func (d *DNSProvider) getMetadata(envVar, field string) (string, error) {
+
+	// first check environment
+	if envVar != "" {
```
ldez: this does not seem necessary

kung-foo:
btw, I wasn't sure if there should be an update https://github.com/xenolf/lego/blob/master/cli.go#L208 that indicates somehow that the env vars are _optional_ if using the instance metadata service. I assume the same would apply to the route53 implementation.

Issue #680, opened 2018-10-17
How to serialize dns solver requests


KnicKnic:
I am using lego through traefik. The provider I am using is DuckDns. 
There is a imitation with duckdns that it only has one txt record per account, that txt record will be presented for any lookup request on any subdomain.

So `test1.test.duckdns.org` and `test2.test.duckdns.org` do not have different txt records. Currently I want to be able to request in traefik 
```toml
 [[acme.domains]]
   main = "test1.test.duckdns.org"
 [[acme.domains]]
   main = "test2.test.duckdns.org"
```
I was wondering how I could serialize the requests such that the two requests do not overlap and that `test1` is validated and then `test2` is validated (currently `test2` will succeed and then eventually `test1` will get retried). I am okay with all requests going through DuckDNS are serialized, but if possible it would be nice to limit it only to one per account. I am curious to this interaction across the operation that spans both `Present` and `CleanUp`.

@ldez - due to already paging in context on https://github.com/xenolf/lego/pull/676

Thanks!

ldez:
I saw the problem with DuckDNS tomorrow, and I may have a solution.
For the moment, I have to think about it.

ldez:
Closed by #700

Issue #551, opened 2018-05-29
Parsing /etc/resolv.conf flips DNS server order


EugenMayer:
in my case i have this `/etc/resolv.conf`

```
domain internal.mycompany.com
search internal.mycompany.com
nameserver 8.8.8.8
nameserver 10.1.7.1
```

while 8.8.8.8 can resolf `SOA` for `www.gateway.internal.mycompany.com` .. our internal DNS server cannot (`10.1.7.1`) since it is using a usual domain override ( 10.1.7.1 ) .

In case that is of any interest, we use the reference implementation of unbound in https://opnsense.org/ with a appliance based private domain defintion.

So beside all the issues with https://github.com/xenolf/lego/issues/544 in general, this issue is about:

When resolv.conf gets parsed, the order of nameservers is inverted, thus 10.1.7.1 is checked first for `_acme-challenge.www.gateway.internal.mycompany.com SOA` and this is an empty response. That will sho a SERVFAIL and the whole challenge fails

EugenMayer:
Extending this section, the order is not simply "inversed". When i use

```
domain internal.mycompany.com
search internal.mycompany.com
nameserver 8.8.8.8
nameserver 10.1.7.1
nameserver 8.8.8.8
```

it will still not work

```
domain internal.mycompany.com
search internal.mycompany.com
nameserver 8.8.8.8
nameserver 10.1.7.1
nameserver 8.8.8.8
```

will work though. Could be still inverse since you might just do dedublication

Issue #481, opened 2018-01-24
disable PreCheckDNS for DNS01 challenges


kfiles:
Many organizations which host their own DNS servers use split-DNS resolution: internal recursive resolvers return different results than external resolvers. This means that attempts to pre-check the TXT records for a DNS01 challenge using recursive resolvers from /etc/resolv.conf will fail to return results available on external authoritative servers. It also means that even the identity of the authoritative nameservers for the domain may be different.

As a result, the PreCheckDNS func will fail in these scenarios. What is needed is an option to disable pre-check on DNS01 challenges altogether.

captncraig:
As a workaround, if you are using acme as a library, you can sub in your own implementation of `acme.PreCheckDNS` ([godoc](https://godoc.org/github.com/xenolf/lego/acme#pkg-variables)) that does nothing. Not optimal, and not useful for cli, but it may work for you.

mholt:
I never did like a pre check at all because of this. Remind me again why we don't simply let the acme challenge itself tell us whether it was successful?

captncraig:
The problem is almost all cloud dns providers have a delay between when their api accepts a change, and when the nameservers actually serve the correct records. This can be a few seconds, or potentially more if their infrastructure is "less optimized".

The safe thing is to pre-verify the nameservers are correct before asking acme validate, because the acme server will just check once and fail.

EugenMayer:
@captncraig the problem here is, see #544 is that the actualy verification is just not working well enough.

In my case, LEGO does way to many mistakes in guessing the SOA record / primary DNS and in the end, fails in doing so.

During that, both reference implementations certbot/acme.sh are working flawlessly in both environments. So what i have right now is, a LEGO lib broken in several ways in DNS-01 in exactly those cases where you need it most:

 - public.tld`s 
 - in private networks
 - with own DNS servers ( because nothing else make sense )

i guess most of the other people would /can use http without any issue, so no need for a cloud DNS provider or anything else. 

For me LEGO is breaking the core aspects of the DNS-01 need and thats a really a bummer. I would really love to see how we could fix that - maybe just :

Propasal:
Go back to the roots of the reference implementations: Let the user define a static, safe timeout (60 seconds). Then tell LE to "check for the txt record" after that timeout. Since LE is not using any private DNS servers, they will never fail to publicly determine the SOA record and thus the whole process will never fail ( and does not for certbot/acme ).

Downside is, waiting for a longer time then maybe actually needed, due to the "safe timeout".

[...]

PR #700, opened 2018-11-04
Refactor the core of the lib

### LINKED ISSUE #263: "Progress" messages for DNS record propagation checks
I'm assuming this check can take 10-20 minutes, especially with multiple domains, but the last message I got was 12 minutes ago.

I understand it'd be pretty difficult to implement any kind of progress bar due to the way the propagation and test works, but maybe a log message when the record is checked again/retried? As it currently stands, I have no idea what the application is doing.

`2016/08/14 16:08:19 [INFO][subdomain.voltagex.org] Checking DNS record propagation...`

### LINKED ISSUE #361: dns_challenge.go does not account for delegated auth zones
dns_challenge.go uses lookupNameservers to find the authoritative nameserver for a given FQDN. That's all well and good. The problem I've run into occurs when that authoritative nameserver is delegating authority to a second nameserver. 
Specifically, in the first nameserver, there is an NS entry in the first nameserver for the given FQDN, pointing to the second nameserver. I'm finding that DNS challenges are failing in this situation as the checkAuthoritativeNss function call only searches for TXT records, it does not handle the case where it encounters an NS entry. The better behavior would be to search the second nameserver for the TXT record.

Information on where to follow the delegation is found in the dnsQuery's response.Ns

### LINKED ISSUE #487: Add omitempty to type and detail for RemoteError
In analyzing payloads for challenge POSTs from xenolf-acme, I noticed that it sends an "error" struct when POSTing a challenge, even though that struct is always empty. It would be nice to either set omitempty on the type and detail fields, or (slightly better but more invasive) change the `Error` field to a `*RemoteError` so it can be omitted when it is nil.

### LINKED ISSUE #493: Poll authz URL instead of challenge URL
In further investigating https://github.com/xenolf/lego/issues/486, I've noticed that at least some of the issues are caused by challenges whose status is "pending" while their parent authz's status is "invalid." 

This is obviously a [Boulder bug](https://github.com/letsencrypt/boulder/issues/3482), and we'll work on fixing it, but one way xenolf-acme can become more robust to it is by polling the authz URL instead of the challenge URL.
We added separate statuses for challenges early on in ACME's evolution, when it seemed like authorizations were likely to have multiple challenges that you would have to fulfill separately.

The latest spec (and Boulder's current assumptions) say that you will never need to fulfill more than one challenge on an authz, so the separate statuses are redundant.
Boulder considers the authz status canonical when calculating whether a client is authorized to issue, so polling the authz URL is somewhat more robust.

### LINKED ISSUE #577: CLI support for renew with CSR?
I'm hoping someone can confirm that the `renew` CLI command doesn't support the use of a CSR without a private key. I'm not a golang dev so I could be misunderstanding something. 

Assuming that's accurate, is it intentional? Simply not implemented yet?

Is it worthwhile to add to the CLI help indicating `--csr` is limited to `run`? I can take a try at writing something if so.

Relevant pull requests:

- #122
- #224

### LINKED ISSUE #582: Pending authorizations not disabled after acmev2 change
I see this part of the code has been commented out - https://github.com/xenolf/lego/blob/d457f70ae0a39ea3e62ddad62caa7c29d33a01e2/acme/client.go#L353-L356

Is there a reason this was done? Shouldn't we continue to disable?

### LINKED ISSUE #651: Where is list command?
I can not find the certificate list command like `$ certbot list`.
Thanks.

### LINKED ISSUE #652: Add a SAN on renew?
The first time I used lego, I added only the domain.com and forgot the www.domain.com. I've tried renewing with the extra --domains command, but the new cert doesn't have the SAN. Is this a limitation of lego, or letsencrypt, or am I just doing something wrong?

Barring that, what's the proper procedure for deleting my old cert and starting from scratch? Is it as simple as deleting the certs and running again with the 'run' command?

Thanks!

### LINKED ISSUE #656: Name.com DNS-provider acme: Error -> One or more domains had a problem:
The error-message continues as follows:
```
time limit exceeded: last error: NS ns2lns.name.com. did not return the expected TXT record
```
This happens to me all the time.
On my local box as well as on my server.
What is interesting is that lego creates a dedicated domain for the acmen-challenge.
_acme-challenge.mydomain.me 
I would expect the TXT-record to be under mydomain.me as I call it for the entire domain.

I call lego in the following way:

```
lego -d mydomain.me -m em@a.il --path $HOME/.lego  --filename $HOME/tmp/certificates/mydomain.crt --pem --dns namedotcom   run
```

As far as propagations are concerned:
locally I use a caching dns-proxy, but I emptied the cache before the lego-run and I run 

```
dig -t TXT _acme-challenge.mydomain.me
```
while I wait for lego to fail.

I see the TXT-record being present.

P.s.: On an unrelated note:
I also noticed that lego does not clean up after itself in the event of a failure. While this is good for debugging, it might not be wanted for the majority of cases a failure happens

### LINKED ISSUE #694: ACME: Support POST-as-GET
[GET and POST-as-GET Requests](https://tools.ietf.org/html/draft-ietf-acme-acme-16#section-6.3)

> On November 1st, 2019 we will remove support for unauthenticated GETs from the V2 API, requiring client support for POST-as-GET. This will have no impact on the legacy V1 API which will remain unchanged.

| Action                | Request                                  | Response       |
|:----------------------|:-----------------------------------------|:---------------|
| Get directory         | GET  directory                           | 200            |
| Get nonce             | HEAD newNonce                            | 200            |
| Create account        | POST newAccount                          | 201 -> account |
| Submit order          | POST newOrder                            | 201 -> order   |
| Fetch challenges      | POST-as-GET order's authorization urls   | 200            |
| Respond to challenges | POST-as-GET authorization challenge urls | 200            |
| Poll for status       | POST-as-GET order                        | 200            |
| Finalize order        | POST order's finalize url                | 200            |
| Poll for status       | POST-as-GET order                        | 200            |
| Download certificate  | POST-as-GET order's certificate url      | 200            |

- [ACME breaking change: Most GETs become POSTs](https://community.letsencrypt.org/t/acme-breaking-change-most-gets-become-posts/71025)
- [Any update on GETS -> POSTs conversation](https://community.letsencrypt.org/t/any-update-on-gets-posts-conversation/73139)
- [ACME v2 - Scheduled deprecation of unauthenticated resource GETs](https://community.letsencrypt.org/t/acme-v2-scheduled-deprecation-of-unauthenticated-resource-gets/74380)

### LINKED ISSUE #709: DNS Provider: Godaddy - Failed to get *.domain.tld with domain.tld as SAN
I use traefik as a load balancer for Docker Swarm. Traefik uses lego to option SSL certificates. 

By setting the following options on the command line, I request traefik to obtain one SSL certificate: 
* *.sub.domain.tld as the main domain
* sub.domain.tld as first SAN entry

```
      --acme=true 
      --acme.email=email@domain.tld
      --acme.storage="acme.json" 
      --acme.entrypoint=https 
      --acme.acmelogging=true 
      --acme.dnschallenge.provider=godaddy
      --acme.dnschallenge.delayBeforeCheck=90
      --acme.domains='*.sub.domain.tld, sub.domain.tld'
```

Lego fails to obtain the certificate:
```
acme: Error 403 - urn:ietf:params:acme:error:unauthorized - Incorrect TXT record "qNAASJ__redacted__c8J0" found at _acme-challenge.sub.domain.tld
```

Looking at the godaddy domain management console, my *impression* is that there are two verification processes for the sub.domain.tld, battling for the _acme-challenge TXT record, overriding the TXT record in the process. 

Once I remove the sub.domain.tld as SAN, the process works flawlessly. 

A similar issue seems to be treated here: https://github.com/josteink/le-godaddy-dns/commit/06c8d3f9b10787a35c26f5ab4e8fe9c788c9dcaa

### LINKED ISSUE #724: builds break against recent dnsimple
As of dnsimple/dnsimple-go@9f1e8e5, I get the following error when building:

    .../providers/dns/dnsimple/dnsimple.go:74:31: undefined: dnsimple.NewOauthTokenCredentials

having imported `github.com/xenolf/lego/providers/dns`. It looks like they deliberately got rid of their built-in OAuth handling, punting that to `golang.org/x/oauth2`, and not keeping this function around for convenience.

### LINKED ISSUE #725: Incompatibility with github.com/miekg/dns package's new updates
the package doesn't compile with the fresh github.com/miekg/dns package's versions since it is incompatible ( usage of dns.ErrTruncated fails)

ldez:
- Packages
- Isolate code used by the CLI into the package `cmd`
- (experimental) Add e2e tests for HTTP01, TLS-ALPN-01 and DNS-01, use [Pebble](https://github.com/letsencrypt/pebble) and [challtestsrv](https://github.com/letsencrypt/boulder/tree/master/test/challtestsrv) 
- Support non-ascii domain name (punnycode)
- Check all challenges in a predictable order
- No more global exported variables
- Archive revoked certificates
- Fixes revocation for subdomains and non-ascii domains
- Fixes #582: Disable pending authorizations
- Fixes #487: use pointer for RemoteError/ProblemDetails
- Fixes #493: Poll authz URL instead of challenge URL
- Fixes #709, #680: The ability for a DNS provider to solve the challenge sequentially
- Fixes #361, #551: Check all nameservers in a predictable order
- Fixes #361, #380, #481: Option to disable the complete propagation Requirement
- Fixes #656
- Closes #577: CLI, support for renew with CSR
- Closes #652: CLI, add SAN on renew
- Closes #651: Add command to list certificates.
- Closes #263: Logs every iteration of waiting for the propagation
- Closes #694: POST-as-GET
- Closes #724: update DNSimple client
- Closes #725: update github.com/miekg/dns
- Closes #401, #400

Before:
```console
.
├── acme
│   └── testdata
├── log
├── platform
│   ├── config
│   │   └── env
│   └── tester
└── providers
    ├── dns
    │   └── ...
    └── http
        └── ...
```

After:
```console
.
├── acme
│   └── api
│       └── internal
│           ├── nonces
│           ├── secure
│           └── sender
├── certcrypto
├── certificate
├── challenge
│   ├── dns01
│   │   └── fixtures
│   ├── http01
│   ├── resolver
│   └── tlsalpn01
├── cmd
│   └── lego
├── e2e
│   ├── dnschallenge
│   ├── fixtures
│   │   └── ...
│   └── loader
├── lego
├── log
├── platform
│   ├── config
│   │   └── env
│   ├── tester
│   └── wait
├── providers
│   ├── dns
│   │   └── ...
│   └── http
│       └── ...
└── registration
```


README.md
```diff
@@ -174,115 +176,129 @@ lego defaults to communicating with the production Let's Encrypt ACME server. If
...
-// A client facilitates communication with the CA server. This CA URL is
-// configured for a local dev instance of Boulder running in Docker in a VM.
-client, err := acme.NewClient("http://192.168.99.100:4000/directory", &myUser, acme.RSA2048)
-if err != nil {
-  log.Fatal(err)
+func main() {
+
+	// Create a user. New accounts need an email and private key to start.
+	const rsaKeySize = 2048
+	privateKey, err := rsa.GenerateKey(rand.Reader, rsaKeySize)
```
mholt: I think users can use ECDSA keys, right?

```suggestion
	privateKey, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
```

README.md
```diff
@@ -174,115 +176,129 @@ lego defaults to communicating with the production Let's Encrypt ACME server. If
...
+	myUser := MyUser{
+		Email: "you@yours.com",
+		key:   privateKey,
+	}
+
+	// A client facilitates communication with the CA server. This CA URL is
+	// configured for a local dev instance of Boulder running in Docker in a VM.
+
+	config := acme.NewDefaultConfig(&myUser).
+		WithCADirURL("http://192.168.99.100:4000/directory").WithKeyType(certcrypto.RSA2048)
```
mholt: I'm not a huge fan of these drive-thru configurations that read like a DSL. There's no way to return errors here, which makes me a little nervous because surely there's a way to pass in a bad configuration?

After inspecting the value that gets returned, I see some effort to keep certain fields unexported. Maybe they could just be exported instead? It's less weird to just do something like this:

```go
config := acme.NewDefaultConfig(&myUser)
config.CADirURL = "..."
config.KeyType = "..."
```

README.md
```diff
@@ -174,115 +176,129 @@ lego defaults to communicating with the production Let's Encrypt ACME server. If
...
+
+	client, err := acme.NewClient(config)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// We specify an http port of 5002 and an tls port of 5001 on all interfaces
+	// because we aren't running as root and can't bind a listener to port 80 and 443
+	// (used later when we attempt to pass challenges). Keep in mind that we still
```
mholt:
```suggestion
	// (used later when we attempt to pass challenges). Keep in mind that you still
```

README.md
```diff
@@ -174,115 +176,129 @@ lego defaults to communicating with the production Let's Encrypt ACME server. If
...
+	reg, err := client.Registration.Register(true)
+	if err != nil {
+		log.Fatal(err)
+	}
+	myUser.Registration = reg
+
+	// The acme library takes care of completing the challenges to obtain the certificate(s).
+	// The domains must resolve to this machine or you have to use the DNS challenge.
+	bundle := true
+	certificates, err := client.Certificate.Obtain([]string{"mydomain.com"}, bundle, nil, false)
```
mholt: Here, I'd love to just pass in a configuration struct:

```go
client.Certificate.Obtain(certificate.CSR{
    SAN: []string{"mydomain.com"},
    Bundle: true,
    PrivateKey: nil,
    MustStaple: false,
})
```

It's elegant and clearer IMO.

ldez: As we got `Obtain` and `ObtainCSR` methods, the name of the type of the parameter cannot be `CSR`. I propose `ObtainRequest` as a CQRS oriented naming.

README.md
ldez: The fluent syntax or the exported field way have the same problem if you want to manage errors.

I propose another way:
```go
config, err := acme.NewConfig(&myUser, acme.WithCADirURL("http://192.168.99.100:4000/directory"), acme.WithKeyType(certcrypto.RSA2048))
```

README.md
mholt: Sounds good. Or `ObtainParams`. 🤷‍♂️  :+1:

README.md
mholt: Why not? Users need to set those values, they're literally supposed to be user-facing.

I have found that usually trying to keep fields unexported lends itself to more complexity for little or no real benefit.

ldez: The configuration manages the configuration: the `Client` doesn't need to validate the configuration because it's not his responsibility (separation of concerns).

The lib's user must use `NewConfig`: if the fields are exported the user can easily create an invalid configuration because he doesn't need to call `NewConfig`.

The configuration must be used only inside the `client` package.

Thanks to the non-exported fields, we can easily modify the configuration struct without breaking the lib's user.

mholt: I see. Well, those are nice benefits. They are needed from the perspective that an invalid config should not be allowed to be created in the first place. (I disagree personally, because functions should validate their input -- and since `NewClient` gets input, it should validate it. What's the harm in using it outside the `client` package? No other packages will probably use it, and if they do, so be it?)

Anyway, just my $0.02 from my own experience, where unexported fields have come back to bite more than actually help. But feel free proceed.

wyattjoh: I try to emulate the designs of the standard library as much as I can, so that new developers coming to the space (and old ones) know what's going on right away. Take a look at for example, the [httputil.ReverseProxy](https://golang.org/pkg/net/http/httputil/#ReverseProxy), which does quite a lot. It allows lots of configuration through it's exported fields. I feel if we are concerned that an internal component/field shouldn't be modified, or we plan to do so, it should be unexported instead, rather than creating a series of "initializer" style config functions, which I think are just a pretty wrapper around accessing the struct directly.

ldez: I know well `httputil.ReverseProxy` and I'm not sure it can be use as a model for Lego.
However, if I am alone in seeing the benefits, I will not impose my point of view on this section.

acme/api/api.go
```diff
@@ -0,0 +1,151 @@
...
+		return nil, fmt.Errorf("failed to post JWS message -> failed to sign content -> %v", err)
+	}
+
+	signedBody := bytes.NewBuffer([]byte(signedContent.FullSerialize()))
+
+	resp, err := a.do.Post(uri, signedBody, "application/jose+json", response)
+
+	nonce, nonceErr := nonces.GetFromResponse(resp)
+	if nonceErr == nil {
+		a.nonceManager.Push(nonce)
```
wyattjoh: It seems like the error is discarded here, I always like to add a comment when this is the case.

challenge/dns01/nameserver.go
```diff
@@ -0,0 +1,232 @@
...
+	"google-public-dns-b.google.com:53",
+}
+
+// recursiveNameservers are used to pre-check DNS propagation
+var recursiveNameservers = getNameservers(defaultResolvConf, defaultNameservers)
+
+// ClearFqdnCache clears the cache of fqdn to zone mappings. Primarily used in testing.
+func ClearFqdnCache() {
+	muFqdnToZone.Lock()
+	defer muFqdnToZone.Unlock()
```
wyattjoh: defer isn't needed here, unneeded overhead, just move call to next line as the next line can't fail.

challenge/dns01/nameserver.go
```diff
@@ -0,0 +1,232 @@
...
+func FindZoneByFqdnCustom(fqdn string, nameservers []string) (string, error) {
+	muFqdnToZone.Lock()
+	defer muFqdnToZone.Unlock()
+
+	// Do we have it cached?
+	if zone, ok := fqdnToZone[fqdn]; ok {
+		return zone, nil
+	}
+
+	var err error
```
wyattjoh: Unless these variables are used outside the loop, I'd consider it over-optimization to put these variables outside the loop.

providers/dns/otc/client.go
```diff
@@ -66,3 +75,183 @@ type recordSet struct {
...
+	err = json.NewDecoder(resp.Body).Decode(&endpointResp)
+	if err != nil {
+		return err
+	}
+
+	for _, v := range endpointResp.Token.Catalog {
+		if v.Type == "dns" {
+			for _, endpoint := range v.Endpoints {
+				d.baseURL = fmt.Sprintf("%s/v2", endpoint.URL)
+				continue
```
wyattjoh: This doesn't make a lot of sense. It loops over the catalog, then if it finds the dns entries, ends up setting the last endpoint to `d`, did you mean break here?

providers/dns/pdns/client.go
```diff
@@ -0,0 +1,219 @@
...
+	var zones []hostedZone
+	err = json.Unmarshal(result, &zones)
+	if err != nil {
+		return nil, err
+	}
+
+	u = ""
+	for _, zone := range zones {
+		if dns01.UnFqdn(zone.Name) == dns01.UnFqdn(authZone) {
+			u = zone.URL
```
wyattjoh: Does this require a break?

challenge/dns01/nameserver.go
ldez: the error is used outside the loop

providers/dns/otc/client.go
ldez: It's not my code, I just move this section. I will check the algo.

mholt:
Great work, @ldez! Thanks for sticking it through. The changes look great.

PR #706, opened 2018-11-07
Add Go modules support

tdewolff:
This adds support for Go modules, generated by conversion from Gopkg.

Be aware that two packages have a different version from that in Gopkg, due to import path problems in those repos. `github.com/ldez/go-auroradns` is now at the latest commit after the author just pushed a fix for the `go.mod` file. `github.com/linode/linodego` is now one commit further, which has the import paths fixed for dependency `resty`.


go.mod
```diff
@@ -0,0 +1,69 @@
...
+	github.com/google/uuid v1.0.0 // indirect
+	github.com/gopherjs/gopherjs v0.0.0-20181103185306-d547d1d9531e // indirect
+	github.com/h2non/gock v1.0.11 // indirect
+	github.com/iij/doapi v0.0.0-20180911005243-8803795a9b7b
+	github.com/json-iterator/go v1.1.5 // indirect
+	github.com/jtolds/gls v4.2.1+incompatible // indirect
+	github.com/juju/ratelimit v1.0.1 // indirect
+	github.com/kolo/xmlrpc v0.0.0-20181023172212-16bdd962781d // indirect
+	github.com/kr/pretty v0.1.0 // indirect
+	github.com/ldez/go-auroradns v0.0.0-20181107021640-532168a1e313
```
ldez: FYI https://github.com/ldez/go-auroradns/releases/tag/v2.0.1

tdewolff: Aha, I'm afraid that the module name for `go-auroradns` is still invalid. I did not realize it is at `v2` already, so it needs `/v2` at the end of it's path.

displague:
@tdewolff `linodego` can be pinned to `0.6.2` to overcome the goresty dependency problem.

tdewolff:
The Travis and GoReleaser scripts need to be updated as well, @ldez do you want to move away from `dep` in the future? It seems like it doesn't go well together with the fact that `ldez/go-auroradns` is not at `v2` and requires `/v2/` in it's path (dep can't find it now), so I don't think they can work alongside eachother...

PR #708, opened 2018-11-08
Add acme-proxy provider

mdbraber:
In my effort to centralize certificate requests for hosts in my network (and not having to spread out credentials) I've written an acme-proxy provider to be used with http://github.com/mdbraber/acme-proxy. It works fine (see below) and the nice thing is because it plugs in to lego, you can simply use any provider that lego offers.

I'm submitting this PR to get comments on tips for the suggested implementation. I'm not sure what's the best way to do this, as I had to change some imports to work with my local branch that obviously shouldn't be in the PR.

Running the proxy:
```console
$ ACMEPROXY_PROVIDER="transip" TRANSIP_ACCOUNT_NAME="mdbraber" TRANSIP_PRIVATE_KEY_PATH="/Users/mdbraber/transip.key" go run acme-proxy.go
```

Requesting a certificate:

```console
$ ACMEPROXY_URL="http://127.0.0.1:9095/" ./lego -m m@mdbraber.com -a -x http-01 -x tls-alpn-01 --dns acme-proxy --dns-resolvers ns0.transip.nl -s https://acme-staging-v02.api.letsencrypt.org/directory -d mdbraber.net -d *.mdbraber.net run
2018/11/08 10:11:38 [INFO] acme: Registering account for m@mdbraber.com
2018/11/08 10:11:38 !!!! HEADS UP !!!!
2018/11/08 10:11:38
		Your account credentials have been saved in your Let's Encrypt
		configuration directory at "/Users/mdbraber/go/src/github.com/mdbraber/lego/.lego/accounts/acme-staging-v02.api.letsencrypt.org/m@mdbraber.com".
		You should make a secure backup	of this folder now. This
		configuration directory will also contain certificates and
		private keys obtained from Let's Encrypt so making regular
		backups of this folder is ideal.
2018/11/08 10:11:38 [INFO] [mdbraber.net, *.mdbraber.net] acme: Obtaining bundled SAN certificate
2018/11/08 10:11:39 [INFO] [*.mdbraber.net] AuthURL: https://acme-staging-v02.api.letsencrypt.org/acme/authz/lBFxlzA3lbOJ8a7cAmIv-vP-Qe1OU4ZSR_q4tmD5Af4
2018/11/08 10:11:39 [INFO] [mdbraber.net] AuthURL: https://acme-staging-v02.api.letsencrypt.org/acme/authz/RY19RDOYqi2UbTBqmU5FmU1ZVx5FT7kP1xsO5dkodIc
[...]

PR #710, opened 2018-11-08
Add DNS provider for "HTTP request".

ldez:
The server must provide:
- `POST` `/present`
- `POST` `/cleanup`

The URL of the server must be define by `HTTPREQ_ENDPOINT`.

### Mode

There are 2 modes (`HTTPREQ_MODE`):
- default mode:
```json
{
  "fqdn": "_acme-challenge.domain.",
  "value": "LHDhK3oGRvkiefQnx7OOczTY5Tic_xZ6HcMOc_gmtoM"
}
```

- `RAW`
```json
{
  "domain": "domain",
  "token": "token",
  "keyAuth": "key"
}
```

### Authentication

Basic authentication (optional) can be set with some environment variables:
- `HTTPREQ_USERNAME` and `HTTPREQ_PASSWORD`
- both values must be set, otherwise basic authentication is not defined.

### Additional information

Thanks to @mdbraber.
Related to #708


mdbraber:
I've been trying to figure out if I can actually push to an unmerged PR I don't own, but I don't think I can? Anyway: you've got "client" in places where it should say "present" (in the README).

Also, I'd change the fqdn example to say '_acme-challenge.domain.tld.' to make it look less like an unqualified domainname.

(P.S. if you can explain if I could push these changes I'd be happy to do so if you can explain how ;-)

ldez:
you cannot push on PR because you are not a maintainer, but you can do a review.

ayang64:
Checks out.

PR #719, opened 2018-11-14
Add version to xenolf-acme in User-Agent.

jsha:
Also, remove "Go-http-client/1.1". In practice this added detail doesn't
wind up being useful in diagnosing problems, particularly since it can
be deduced from the xenolf-acme version.

PR #728, opened 2018-12-06
Release helper

ldez:
- fix: missing version
- add release helper commands

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,3 +37,17 @@ checks:
 
 fmt:
 	gofmt -s -l -w $(SRCS)
+
+# Release helper
+
+patch:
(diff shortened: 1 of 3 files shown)


mholt:
Have you thought of using `-ldflags` to set variable values instead of having to parse the AST? You might find that easier. We do this in Caddy here: https://github.com/mholt/caddy/blob/8f583dcf3694a55e32a1eb9ae4316c3f99d38c6f/caddy/build.go#L64 (actual code is here: https://github.com/caddyserver/builds/blob/c62e2219460a8828970dad09212c3a4cec40b56c/compilation.go#L95)

ldez:
It must be set in the code not at the compile time, see #719

PR #732, opened 2018-12-12
Add gopkg lock go package versions and fix updates with new packages.

angulito:
- Add gopkg lock to the `acmev2`
- Fix `vendor/github.com/xenolf/lego/acme/dns_challenge.go:225:13: undefined: dns.ErrTruncated` error when using this branch because of an update on the `github.com/miekg/dns` lib.
- Fix also update library errors:
    - `providers/dns/exoscale/exoscale.go`
    - `providers/dns/dnsimple/dnsimple.go`


greut:
@angulito thanks for egoscale. You could pin a version in the Gopkg.toml.

angulito:
@greut done - also please check that the tests are failing because of errors that I don't know - it seems like the libraries were updated to new versions and maybe the tests are not working well

angulito:
Interesting but the previous executions for the pipeline didn't work because of I had locally an older version of the `dep` tool - seems it generates a different gopkg.lock file :)

greut:
_dep_ 0.50 introduced a change in the lock format.

angulito:
yep, I am reading about that, thank you @greut

greut:
acmev2 is now master, https://github.com/xenolf/lego/pull/555

angulito:
interesting - then this PR does not make sense! thank you @greut

greut:
@angulito I wish I discovered the base branch earlier.

PR #735, opened 2018-12-19
Use new challtestsrv.

ldez:
Let's Encrypt move the challtestsrv from boulder to pebble (which is great).

**This change made by LE is breaking all our builds.**

- https://github.com/letsencrypt/boulder/pull/3980
- https://github.com/letsencrypt/pebble/pull/181

ping @cpu


cpu:
> This change made by LE is breaking all our builds.

Apologies, I didn't know you were using the `challtestsrv` binary that was moved

PR #736, opened 2018-12-19
fix: errors management in the prober.

### LINKED ISSUE #734: Errors returned by PreSolve are discarded
Hi,

I'm using lego to issue a wildcard certificate using the DNS challenge and the `exec` dns provider. This setup worked for a while, then something broke two weeks ago and now I'm in the process of finding out what's wrong. In the process I updated to the code in the master branch and noticed that when the script called by the `exec` provider returns an error during `present`, this error is just discarded.

Take the following script:
```
#!/bin/bash

echo "script is exiting now"
exit 1
```

produces the following output when used with lego (b05b54d1f69a31ceed92e2995243c5b17821c9e4):
```
export EXEC_PATH=./update.sh
./lego --domains '*.example.com' --dns-resolvers 8.8.8.8 --dns exec renew --days 30
2018/12/19 11:51:28 [INFO] [*.example.com] acme: Trying renewal with 255 hours remaining
2018/12/19 11:51:28 [INFO] [*.example.com] acme: Obtaining bundled SAN certificate
2018/12/19 11:51:29 [INFO] [*.example.com] AuthURL: https://acme-v02.api.letsencrypt.org/acme/authz/[...]
2018/12/19 11:51:29 [INFO] [*.example.com] acme: use dns-01 solver
2018/12/19 11:51:29 [INFO] [*.example.com] acme: Preparing to solve DNS-01
2018/12/19 11:51:29 script is exiting now

2018/12/19 11:51:29 [INFO] [*.example.com] acme: Trying to solve DNS-01
2018/12/19 11:51:29 [INFO] [*.example.com] acme: Checking DNS record propagation using [8.8.8.8:53]
2018/12/19 11:51:29 [INFO] Wait [timeout: 1m0s, interval: 2s]
```

Then it waits for the DNS record (which was never inserted) to propagate and eventually fails.

I found two places where this might have been caused:
 * The error returned by `Present()` is logged, but not returned in the DNS challenge implementation, here: https://github.com/xenolf/lego/blob/b05b54d1f69a31ceed92e2995243c5b17821c9e4/challenge/dns01/dns_challenge.go#L91-L96
 * In `parallelSolve`, the error is saved into a map under the target domain https://github.com/xenolf/lego/blob/b05b54d1f69a31ceed92e2995243c5b17821c9e4/challenge/resolver/prober.go#L137 but then looks it up using the authz(?) value: https://github.com/xenolf/lego/blob/b05b54d1f69a31ceed92e2995243c5b17821c9e4/challenge/resolver/prober.go#L152

The correction for the first issue is easy, but I'm not sure what to do with the `parallelSolve`. Let me know if there's anything else I can do to help.

ldez:
Due to an invalid naming, some errors are not handled as expected.

Fixes #734

### DIFF
--- a/challenge/resolver/prober.go
+++ b/challenge/resolver/prober.go
@@ -110,7 +110,7 @@ func sequentialSolve(authSolvers []*selectedAuthSolver, failures obtainError) {
 		// Solve challenge
 		err := authSolver.solver.Solve(authSolver.authz)
 		if err != nil {
-			failures[authSolver.authz.Identifier.Value] = err
+			failures[domain] = err
 			cleanUp(authSolver.solver, authSolver.authz)
 			continue
 		}
--- a/challenge/resolver/prober.go
+++ b/challenge/resolver/prober.go
@@ -149,14 +149,15 @@ func parallelSolve(authSolvers []*selectedAuthSolver, failures obtainError) {
 	// Finally solve all challenges for real
 	for _, authSolver := range authSolvers {
 		authz := authSolver.authz
-		if failures[challenge.GetTargetedDomain(authz)] != nil {
+		domain := challenge.GetTargetedDomain(authz)
+		if failures[domain] != nil {
 			// already failed in previous loop
 			continue
 		}
 
 		err := authSolver.solver.Solve(authz)
 		if err != nil {
-			failures[challenge.GetTargetedDomain(authz)] = err
+			failures[domain] = err
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


ldez:
The build will fail until #735 is merged.

challenge/resolver/prober.go
```diff
@@ -149,14 +149,14 @@ func parallelSolve(authSolvers []*selectedAuthSolver, failures obtainError) {
 	// Finally solve all challenges for real
 	for _, authSolver := range authSolvers {
 		authz := authSolver.authz
-		if failures[authz.Identifier.Value] != nil {
+		if failures[challenge.GetTargetedDomain(authz)] != nil {
```
wyattjoh: Why not just pull `domain := challenge.GetTargetedDomain(authz)` at the top of the loop?

ldez: because it's only use one time inside the loops.

PR #739, opened 2018-12-20
alidns: add support to handle more than 20 domains

smartwang:
alicloud dns api returns only 20 domains once by default. I tried to add pagination support to handle more than 20 domains correctly, and it's works.


dgautumn:
Hi, I has found that the condition: response.PageNumber >= response.PageSize on line 153 in the alidns provider file ( providers/dns/alidns/alidns.go ) may not correct. The response.PageSize means how many domains at every page in response, so I think it should be response.PageNumber * response.PageSize >= response.TotalCount, whitch the response.TotalCount means how many domains in a alicloud account.

PR #744, opened 2018-12-31
transip: concurrent access to the API.

ldez:
The Transip API doesn't allow to add a record without update all the zone/domain records. 

In some case, this produce a race condition.

See https://github.com/containous/traefik/issues/4346

**Note:** a wildcard certificate is the best way to handle this case.

### DIFF
--- a/providers/dns/transip/transip.go
+++ b/providers/dns/transip/transip.go
@@ -92,6 +92,7 @@ func (d *DNSProvider) Present(domain, token, keyAuth string) error {
 	// get the subDomain
 	subDomain := strings.TrimSuffix(dns01.UnFqdn(fqdn), "."+domainName)
 
+	// use mutex to prevent race condition from GetInfo until SetDNSEntries
 	d.dnsEntriesMu.Lock()
 	defer d.dnsEntriesMu.Unlock()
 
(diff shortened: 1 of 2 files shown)


providers/dns/transip/transip.go
```diff
@@ -90,6 +92,9 @@ func (d *DNSProvider) Present(domain, token, keyAuth string) error {
 	// get the subDomain
 	subDomain := strings.TrimSuffix(dns01.UnFqdn(fqdn), "."+domainName)
 
+	d.dnsEntriesMu.Lock()
```
ronaldevers:
```suggestion
	// use mutex to prevent race condition from GetInfo until SetDNSEntries
	d.dnsEntriesMu.Lock()
```

providers/dns/transip/transip_test.go
```diff
@@ -135,6 +203,124 @@ func TestNewDNSProviderConfig(t *testing.T) {
...
+
+		time.Sleep(100 * time.Millisecond)
+		err = p.CleanUp(domain1, "", "")
+		require.NoError(t, err)
+	}()
+
+	go func() {
+		defer wg.Done()
+
+		domain2 := "foo.lego.wtf"
```
wyattjoh: Why not move this test code into a function to reduce the duplication of logic here?

providers/dns/transip/transip_test.go
```diff
@@ -135,6 +203,124 @@ func TestNewDNSProviderConfig(t *testing.T) {
...
+	token := ""
+	keyAuth := ""
+
+	var wg sync.WaitGroup
+	wg.Add(2)
+
+	go func() {
+		defer wg.Done()
+		time.Sleep(550 * time.Millisecond)
+		err := p.Present("bar.lego.wtf", token, keyAuth)
```
wyattjoh: Same as above.

PR #774, opened 2019-01-29
fastdns: Do not overwrite existing TXT records

### LINKED ISSUE #763: FastDNS: Do not overwrite existing TXT records
When requesting a certificate for `*.mydomian.com` and `mydomain.com` using dns-01 challenge, and FastDNS provider, the TXT gets overridden and the TXT verify fail most of the times with this error:

```
[*.mydomain.com] time limit exceeded: last error: NS a1-216.akam.net. did not return the expected TXT record...
```
Version: lego version 2.0.1 linux/amd64 

The command that i use:
```bash
lego -s https://acme-staging-v02.api.letsencrypt.org/directory -a --email myemail@email.com --dns fastdns --pem --dns.resolvers a1-216.akam.net -d 'mydomain.com' -d '*.mydomain.com' run
```

This seems the same issue resolved for Azure in https://github.com/xenolf/lego/pull/726. The error is not about timeout, i raise it for the test with no success, is because one of the challenges overrides the other, and one of then never completes.

If i have some time, i'll try to investigate how to solve the issue, this error seems probable to appear in other providers, I'll think in a solution that will help to all providers. The thing thar are around my brain is to clear TXT records, before creating the new ones for the challenge, and of course no override existing TXT records for the same name. This will happen always forn SAN and Wildcard certificate at same time.

NOTE: I changed domain name and emails for safety.

ldez:
Fixes #763


ldez:
@wyattjoh Thanks to the change of our algo I found a bug (a panic) in the client https://github.com/xenolf/lego/issues/763#issuecomment-456783312.

So I opened a PR on the client https://github.com/xenolf/lego/issues/763#issuecomment-456801170

Issue #771, opened 2019-01-25
No access to authorization URI after solver failure


dhduvall:
I mentioned this in #770, but I think this deserves its own issue. If you call `Obtain()` and the solver fails, there's no way for a consumer of the lego API to get the authorization URI in order to deactivate it or retry it at a later date.

The `URL` member of `ProblemDetails` might have this information, but I'm not seeing it in any of the errors I'm getting back, except those for invalid nonces. Even if it did, it's not clear to me how I would be able to tell the difference between different kinds of failure.

Issue #770, opened 2019-01-25
gcloud provider needs workaround for inconsistent nameserver results


dhduvall:
I've filed an issue with Google (https://issuetracker.google.com/issues/123397631) but lego probably needs a workaround for the problem. The summary is that even once all of a domain's nameservers have responded with the correct data (and thus triggering a successful result of lego's pre-check routines), one or more of the nameservers may revert to responding with old data or NXDOMAIN. They eventually settle down after a (potentially unbounded?) amount of time.

I'm not sure what the best way is of adding extra time to the pre-check method. Because `checkDNSPropagation()` isn't exported, I can't simply create a pre-check function that calls it first, then waits (or continues to check for a while). Simply trying again isn't a great option, since there's no way to deactivate the authorization from this side of the API, as (as best I can tell) there's no access to the authorization URI in the error you get back from `Obtain()` and no way to create it from the `ObtainRequest`, and I accidentally ended up maxing out the authorizations rate limit figuring this out (thankfully I kept the logs containing the URIs).

I'm happy to put a fix together, but would appreciate some direction.

tjhiggins:
@dhduvall Are you still having this issue? Curious if you think its related to: https://github.com/go-acme/lego/issues/1087

dhduvall:
It might be: lego could be seeing success after those 50 seconds, based on the individual servers it happens to have gotten to respond to its request, but when LE checks, it ends up with servers that don't have the new data yet, and ends up failing the request.

That said, IIRC, I rarely saw lego think the propagation was complete before the timeout. Maybe GCP's architecture is a bit different now, that this is happening?

I haven't seen it myself since I added the workaround with `WrapPreCheck()`.

And I'm still mystified as to why people don't see this when running `certbot` against GCP. Maybe most people aren't doing the DNS verification.

tjhiggins:
Thanks for the quick response. I commented on the other ticket. Looks like my issue is different, but still unsure how to resolve.

PR #778, opened 2019-02-02
Allow for a delay prior to the DNS precheck

dhduvall:
This requires the challenge objects to be able to group authorizations by the order that spawned them, so that waiting isn't done per authorization, but per order (and only once per order), as all of an order's `PreSolve` steps are done before all of its `Solve` steps.  This leads to an `acme.ExtendedAuthorization` object which has a handle to the `acme.ExtendedOrder`, but then requires threading throughout the code.

This gives me what seems to be a usable workaround for the Google Cloud DNS issue in #770, which is to have a two-minute delay between `PreSolve()` and the precheck. This seems to have a significantly lower failure rate than simply running the precheck repeatedly until it succeeds multiple times in a row (in fact, it has a 0% failure rate in my experiments once nonce errors and presolve-stage errors are taken into account).

The `ExtendedAuthorization` type and its threading throughout isn't ideal, IMO, but something similar would be required regardless. My initial naive implementation simply slept the entire delay duration inside `preCheck.call()`, but that meant that every time through was a long sleep, and the propagation timeout could trigger. Pulling it out of the loop and into `Challenge.Solve()` directly helped, but that slept between solves of different authorizations from the same order, which shouldn't be necessary, since they all were told to propagate at the same time. But a single `solveAt` stashed in the `preCheck` object gets reused for all orders from the same client, so that wouldn't work for more than very simple examples.

Having the `Order` object available might make #771 fixable, too.

### DIFF
--- a/acme/api/authorization.go
+++ b/acme/api/authorization.go
@@ -9,15 +9,15 @@ import (
 type AuthorizationService service
 
 // Get Gets an authorization.
-func (c *AuthorizationService) Get(authzURL string) (acme.Authorization, error) {
+func (c *AuthorizationService) Get(authzURL string) (acme.ExtendedAuthorization, error) {
 	if len(authzURL) == 0 {
-		return acme.Authorization{}, errors.New("authorization[get]: empty URL")
+		return acme.ExtendedAuthorization{}, errors.New("authorization[get]: empty URL")
 	}
 
-	var authz acme.Authorization
+	var authz acme.ExtendedAuthorization
 	_, err := c.core.postAsGet(authzURL, &authz)
 	if err != nil {
-		return acme.Authorization{}, err
+		return acme.ExtendedAuthorization{}, err
 	}
 	return authz, nil
 }
--- a/acme/api/authorization.go
+++ b/acme/api/authorization.go
@@ -9,15 +9,15 @@ import (
 type AuthorizationService service
 
 // Get Gets an authorization.
-func (c *AuthorizationService) Get(authzURL string) (acme.ExtendedAuthorization, error) {
+func (c *AuthorizationService) Get(authzURL string) (acme.Authorization, error) {
 	if len(authzURL) == 0 {
-		return acme.ExtendedAuthorization{}, errors.New("authorization[get]: empty URL")
+		return acme.Authorization{}, errors.New("authorization[get]: empty URL")
 	}
 
-	var authz acme.ExtendedAuthorization
+	var authz acme.Authorization
 	_, err := c.core.postAsGet(authzURL, &authz)
 	if err != nil {
-		return acme.ExtendedAuthorization{}, err
+		return acme.Authorization{}, err
 	}
 	return authz, nil
 }
--- a/acme/commons.go
+++ b/acme/commons.go
@@ -163,14 +163,6 @@ type Order struct {
 	Certificate string `json:"certificate,omitempty"`
 }
 
-// ExtendedAuthorization extends Authorization with extra fields.
-type ExtendedAuthorization struct {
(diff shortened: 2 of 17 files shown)


ldez:
It's a breaking change of the API, I will prefer a less invasive modification.

dhduvall:
I don't see how it is. The token needs to be the same across all authorizations of the same order; `Authorization.Identifier` is unique to that authorization.

The only other thing I've been able to sketch out is to add a way to pass the mapping between `Authorization` and `Order` into the `Prober` object, so that it can delay between `PreSolve()` and `Solve()`. But that would mean either changing the `Prober.Solve()` signature (and thus that of `certificate.resolver.Solve()`) or by introducing a new interface next to `certificate.resolver` that allowed for this. I think the latter would be non-breaking.

ldez:
Any changes of the any exported function signatures are a breaking changes of the API lib.
But we can create a breaking change if needed.

For me the `ExtenddAuthorization` is not needed.
A change of the signature of the method `Solve` is enough (`Solve(order, authz)`).
 
Several remarks:

`solveAt map[string]time.Time`:
- as the `precheck` is a singleton, `solveAt` is a singleton.
- it's a map so it's an issue with concurrent calls.
- as it's a singleton, the entry must be clean after the process.

The algo to apply the precheck delay can be simplify: because the delay need to be apply once we don't need to use a `Time` and we don't need to compute a `time.Until()` for each iteration.

As the the order is an `ExtendedOrder` you can use the `Location` as UID.

FYI:

> ACME draft Section 7.4 "Applying for Certificate Issuance"
> https://tools.ietf.org/html/draft-ietf-acme-acme-12#section-7.4
> says:
>   Clients SHOULD NOT make any assumptions about the sort order of
>   "identifiers" or "authorizations" elements in the returned order
>   object.

ldez:
> The algo to apply the precheck delay can be simplify: because the delay need to be apply once we don't need to use a Time and we don't need to compute a time.Until() for each iteration.

I would moderate this sentence due to `parallelSolve`

ldez:
Could you explain your last changes?
Why have you move the delay from `challenge.Solve` to `parallelSolve`?
Why only `parallelSolve` and `sequentialSolve`?

dhduvall:
As I was working through the problem again, it occurred to me that all I really needed was to have all of the `PreSolve()` calls finish, then do the sleep, then make all the `Solve()` calls—that's the definition of what I need. Since `Prober.Solve` operates on a single order, it allows me to not need the whole `Authorization`/`Order` mapping, and eliminates the concurrency issues with the map. But it means that now the `Challenge` needs to be able to communicate the delay time back to this piece of the code.

All in all, it seemed simpler to me. I did forget to update `sequentialSolve()`, though.

I can put the sleep back in the `Challenge` if you prefer, or I can keep it as is and update `sequentialSolve()`.

dhduvall:
No worries.  I was also playing with another change that added a function to wrap the precheck function, giving me the possibility to add some pre- and post-processing (I wanted to call precheck multiple times to make sure it returned consistent results, but that didn't help the way I thought it would).

Originally I thought I could do the delay with that, then realized I couldn't because it would sleep every time. Now I finally realized if the wrapper closed over `solveAt`, then I can keep all the metadata I need outside of `lego`. I still need something like the `Authorization`/`Order` mapping, but I get by with the domains: I keep a map of all the domains in an order to to a unique token when I create the order, and extract the token using the domain, which I'm now passing to the `call()` method on the `preCheck` object.

I can submit that as a separate pull request, along with sample code, if you'd like to see it.

ldez:
Closed in favor of #783

PR #584, opened 2018-06-21
add CNAME support for creating dns-01 TXT challenge

gurvindersingh:
Domain for which certificate is asked for can be a CNAME, so we should check it.
If domain has a CNAME, create the challange TXT record in the alias domain.

Issue #479, opened 2018-01-22
Validate dns challenge via dedicated second domain


captncraig:
I have an interesting use case that may or may not make sense to integrate into this project. Consider the following situation:

- I have many domains that I want on a single SAN cert.
- They have dns hosted in different places (and some in multiple places)
- Organization may be reluctant to provide dns credentials to some or all of those domains for LE purposes

 I had an idea though when I saw someone mention that the Let's Encrypt validator follows CNAME records:

1. Set up permanent `CNAME _acme-challenge.example.com` to point at `_acme-challenge.example.com.otherdomain.tld`
2. To validate the domain, you just need to set the appropriate TXT record on `otherdomain.tld`, and LE can validate that.

This saves you from having to know how to update DNS on each individual domain, as long as you can edit the dedicated matching domain.

I guess the first question is, "Is this general purpose enough to consider for inclusion in this repo"? 

My [proof of concept](https://github.com/xenolf/lego/compare/master...captncraig:transform?expand=1) works as another provider that acts as a wrapper around the real provider you are using. It is fairly limited in how it can transform the record, because of the current way providers generate their own fqdns. But it does work. 

Positives:

- Lets you issue certs on domains that you cannot directly control, as long as you have previously created the appropriate CNAME. 
- Works fairly seamlessly with other providers, and cli (currently just set `--dns transform` along with `TRANSFORM_PROVIDER` and `TRANSFORM_DOMAIN` env vars)

Negative:

- kinda tricky to understand what exactly is going on. Might be kinda niche.
- subdomain certs require a longer, more verbose CNAME setup. I wish a wildcard CNAME could  cover it, but that isn't feasable. In the future I think wildcard certs should only need one CNAME though per wildcard pattern.
- current code path where each provider calls acme.DNS01Record to get the fqdn prevents me from doing other patterns, like have all the cnames point to a single or simpler name on the validation domain. But whatever.

Any thoughts? Should I submit a PR from what I have?

captncraig:
After a train ride to think about this, I think I've talked myself out of it. It feels like a bit too much yak shaving. I can't explain the process cleanly to my team without a whole lot of "you want me to do what?". 

So disregard this. I'm probably going to levera our [existing dns management platform](https://github.com/StackExchange/dnscontrol) as a challenge provider shim, since that already has a clear view of the layout of our domains and providers. I'll have a simple command that gathers the desired certs and uses lego as a library to issue certs after we initialize the provider shims appropriately on a per-domain basis.

[...]

PR #791, opened 2019-02-08
Resolve CNAME when creating dns-01 challenge

mhoran:
It may be desirable to host the dns-01 challenge in a zone other than
the one where the challenge is presented. For example, when validating
a.example.com, the challenge may need to live on example.org.

This change resolves CNAMEs encountered when determining the FQDN of the
challenge, and replaces them with the alias.

This PR is based on the original work in #584.

Closes #584 #506
Related to #479

Co-authored-by: Gurvinder Singh <gurvinder.singh@uninett.no>

### DIFF
--- a/challenge/dns01/cname.go
+++ b/challenge/dns01/cname.go
@@ -7,10 +7,10 @@ func updateDomainWithCName(r *dns.Msg, fqdn string) string {
 	for _, rr := range r.Answer {
 		if cn, ok := rr.(*dns.CNAME); ok {
 			if cn.Hdr.Name == fqdn {
-				fqdn = cn.Target
-				break
+				return cn.Target
 			}
 		}
 	}
+
 	return fqdn
 }
--- a/challenge/dns01/cname.go
+++ b/challenge/dns01/cname.go
@@ -1,0 +1,16 @@
+package dns01
+
+import "github.com/miekg/dns"
+
+// Update FQDN with CNAME if any
+func updateDomainWithCName(r *dns.Msg, fqdn string) string {
+	for _, rr := range r.Answer {
+		if cn, ok := rr.(*dns.CNAME); ok {
+			if cn.Hdr.Name == fqdn {
+				fqdn = cn.Target
+				break
+			}
+		}
+	}
+	return fqdn
+}
--- a/challenge/dns01/dns_challenge.go
+++ b/challenge/dns01/dns_challenge.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/miekg/dns"
 	"github.com/xenolf/lego/acme"
 	"github.com/xenolf/lego/acme/api"
 	"github.com/xenolf/lego/challenge"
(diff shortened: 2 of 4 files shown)


ldez:
My main concerns about this change are:
- possible regression for other providers
- possible regression for current users who used CNAMEs that were not operated by Lego
- adding undocumented behavior (and how to document it)

mhoran:
Thanks for the quick review, @ldez!

> possible regression for other providers

I tested this only with the route53 provider, so this is definitely a possibility. However, I think the limited scope (discussed below) should have limited impact.

> possible regression for current users who used CNAMEs that were not operated by Lego

It's my understanding that this will only affect those who have set up a CNAME for `_acme-challenge.a.example.com`. While one could possibly have set up `*.a.example.com`, that does seem to me to be improbable (except, perhaps, for `*.example.com`).

> adding undocumented behavior (and how to document it)

Totally. When my first attempt to use lego with this configuration failed, I was a bit surprised, but then understood why. My first pass at solving the problem involved modifying the route53 provider itself, but then found #584, which seemed to be a better solution (since that's the behavior I was expecting at first.)

Perhaps some additional logging during the resolution of `_acme-challenge.a.example.com` would be helpful? Alternatively, perhaps this could be optional behavior, enabled via a flag? Or, conversely, enabled by default, but easily disabled.

Given the initial discussion in #479, as well as #506 and #584, this does seem to be a desirable feature. It also sounds like other ACME clients have this [functionality](https://github.com/xenolf/lego/pull/584#issuecomment-454128811).

Daegalus:
As this is an important feature for me, maybe take a look at how acme.sh does it. They have had this feature for a long while, and are currently the only tool that supports it in a release. Its very simple to use and doesn't cause any extra behavioral issues if you don't pass in the extra flags to get it to follow CNAMEs

ldez:
Seems like acme.sh doesn't support CNAME https://github.com/Neilpang/acme.sh/issues/957 :thinking:

Daegalus:
They call it DNS Alias mode. It works just fine and I use it for 2 projects already, just manually.

https://github.com/Neilpang/acme.sh/wiki/DNS-alias-mode

ldez:
ok, it's not an automatic CNAME support but a manually support.
This PR is trying to do an automatic CNAME support, then there are more constraints.

I'm trying to find a way to "feature-flip" the CNAME support, but it's not easy at all.

ldez:
I chose the simplest and least invasive solution.

This is not my preferred solution but it is the least risky at the moment.

The solution is based on an environment variable: `LEGO_EXPERIMENTAL_CNAME_SUPPORT`

The possible values to enable support are: `true`, `TRUE`, `True`.

mhoran:
Sounds good to me! Just verified that this works for my purposes.

sebschlicht:
Sorry to "reopen" such an old issue but I feel like this is the best place to ask: How can I opt-out of this behavior?

My setup relies on the ACME challenging being solved on the very domain specified to lego and not the CNAME.
The CNAME just redirects to my dynamic DNS provider.

ldez:
Hello, this is a PR about `LEGO_EXPERIMENTAL_CNAME_SUPPORT`, not an issue :wink: 

This env var has been removed in PR #1718.

Currently, to disable CNAME, you should set the env var `LEGO_DISABLE_CNAME_SUPPORT` to `true`.

ldez:
The documentation is here: https://go-acme.github.io/lego/usage/cli/options/index.html#lego_disable_cname_support

Issue #486, opened 2018-02-07
Excessive requests from some xenolf-acme clients


jsha:
Hi,

I've been reviewing Let's Encrypt logs for subscribers that are using excessive resources. Some of the top users of resources are using the xenolf-acme user-agent. I'm sure this is not the maintainers' fault; there are probably a number of ways end-users can misconfigure the client to cause these problems. However, I'm wondering if there are things xenolf-acme could do to make it harder to reach these misconfigurations.

For instance, one mode I'm seeing is a client that will poll a single challenge tens of thousands of times, at a rate of sixteen requests per second. Perhaps xenolf-acme could build in a hardcoded max poll rate of once per second, with a deadline of ten minutes, after which the challenge won't be polled anymore?

Similarly, I see some xenolf-acme clients that fail validations for the same domain repeatedly. For instance, one client retried the same failing domain about 1,100 times in a single day. It would be great if xenolf-acme implemented backoff on failing validations as described in https://letsencrypt.org/docs/integration-guide/.

Note that I can't tell from logs whether these error situations are caused by using lego from the command line (in a tight loop) versus the in-memory client. The user-agent string is "Go-http-client/1.1 (linux; amd64) xenolf-acme". I suspect they are probably using the in-memory client based on the rapid polling of the same challenge id for hours.

Also, it would be useful for xenolf-acme to include a version number in the user-agent string, so that after these issues are improved, I can easily tell whether a given client is using an older version or not.

Thanks!

captncraig:
I've done this via caddy on accident multiple times. My scenario has been:

- caddy in a docker container with `restart always`.
- add a new domain and forget to create the appropriate dns records.

Caddy goes into a restart loop and continually fails challenges. This is apparently a design decision for caddy to exit if it fails to get certs, but it does quickly get me rate limited when I make mistakes like the above. 

Not sure if my scenario can cause the types of traffic to a single challenge like you are seeing, or if repeated attempts will result in new challenges.

jsha:
I believe Caddy has its own user-agent, and I have seen it crash in a tight loop as you describe.

I can't be sure, but I think this case is different, partly because the ratio between number of challenge polling requests and number of new-authz requests is really high. Whereas in the Caddy case I think it's generally 1:1ish.

mholt:
[...]

PR #794, opened 2019-02-09
Uses a jittered exponential backoff

ldez:
- during tests, allow to support more than 90% of bad nonce with a minimum of attempts.
   - before with the simple retry system (5 attempts): ~50%
- partial fix for #486 and #793

### DIFF
--- a/acme/api/api.go
+++ b/acme/api/api.go
@@ -2,12 +2,15 @@ package api
 
 import (
 	"bytes"
+	"context"
 	"crypto"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"net/http"
+	"time"
 
+	"github.com/cenkalti/backoff"
 	"github.com/xenolf/lego/acme"
 	"github.com/xenolf/lego/acme/api/internal/nonces"
 	"github.com/xenolf/lego/acme/api/internal/secure"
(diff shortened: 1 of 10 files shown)


acme/api/api.go
```diff
@@ -64,34 +67,46 @@ func (a *Core) post(uri string, reqBody, response interface{}) (*http.Response,
...
-			log.Infof("nonce error retry: %s", err)
-			resp, err = a.retrievablePost(uri, content, response, retry+1)
-			if err != nil {
-				return resp, err
+func (a *Core) retrievablePost(uri string, content []byte, response interface{}) (*http.Response, error) {
+	// during tests, allow to support ~90% of bad nonce with a minimum of attempts.
+	bo := backoff.NewExponentialBackOff()
+	bo.InitialInterval = 200 * time.Millisecond
+	bo.MaxInterval = 5 * time.Second
+	bo.MaxElapsedTime = 20 * time.Second
```
wyattjoh: Maybe worth exporting these duration's from the package? Then it can be controllable by clients?

acme/api/api.go
```diff
@@ -64,34 +67,46 @@ func (a *Core) post(uri string, reqBody, response interface{}) (*http.Response,
...
+	var resp *http.Response
+	operation := func() error {
+		var err error
+		resp, err = a.signedPost(uri, content, response)
+		if err != nil {
+			switch err.(type) {
+			// Retry if the nonce was invalidated
+			case *acme.NonceError:
+				log.Infof("nonce error retry: %s", err)
+				return err
```
wyattjoh: `ctx` is not cancelled from this exit point, is that intentional?

ldez: yes because I want to retry: if I cancel here i will not have retries.

acme/api/api.go
ldez: It's internal behavior, clients don't need to change these values.
If someone really needs to change that, we can export some values later.

ldez: FYI before, with the system of 5 attempts, client waited around 1-2 seconds

PR #800, opened 2019-02-18
exoscale: update library

greut:
- changing the default endpoint to the api.exoscale.com one (both .ch and .com are working)
- bump the library to 0.14 (which supports ctx)

Do you prefer `context.TODO()` (used here) or `context.Background()`?


ldez:
Hello, in this case `context.Background()` seems better.

greut:
thanks @ldez!

PR #825, opened 2019-03-12
Add 'manual' provider to the output of dnshelp

porjo:
The output of `./lego dnshelp` was missing `manual` as a DNS provider.


ldez:
FYI all files with the prefix `zz_gen_` are generated.
https://github.com/go-acme/lego/blob/5e741000cc15c4951e445bbc8cd89c59d14d560c/cmd/zz_gen_cmd_dnshelp.go#L3-L4

The changes must be applied in the code template.
https://github.com/go-acme/lego/blob/5e741000cc15c4951e445bbc8cd89c59d14d560c/internal/dnsdocs/dns.go.tmpl#L15

I fixed the problem.

PR #837, opened 2019-03-21
fix pdns: wildcard with SANs

tbe:
The current implementation of the DNS challenge does not allow
to set multiple TXT records at once.

As PowerDNS has the concept of record sets, and so all records
for the same type and name must set during one call, we would override
existing records.


ldez:
As I understand the sequential behavior it's not the better solution: during the `Present` you get get the existing records and merge them in the recordset like the gcloud provider.

https://github.com/go-acme/lego/blob/0ce6ba36b95542f1ef1beebb416a03130924837f/providers/dns/gcloud/googlecloud.go#L161

tbe:
> As I understand pdns, the sequential behavior it's not the better solution: during the `Present` you can get the existing records and merge them in the recordset like the gcloud provider.

Of course. But as the godaddy provider also use the sequential behavior, i used the approach, that is already used in some parts of the code.

I will redesign this PR to pdns best practice.

ldez:
the sequential behavior is the last solution: when the provider either doesn't support records set or doesn't works well with records for the same domain.
It's more a workaround than a real solution.

tbe:
I have rewritten the PR, to merge the txt records, instead of using sequential resolves

PR #845, opened 2019-04-01
Adds renew hook

### LINKED ISSUE #427: add certbot like hook script of return codes
I'd like to be able to hook a script like certbot can do when a new certificate is generated / updated

Ideally this would be something like: `--renew-hook /root/certbot-renew-hook.sh` 

If this sounds less fun to implement, alternatively or additionally various exit codes could be used to detect renewl?

Today, when I run `renew --days=30` I get the same exit-code from lego regardless of what happenned...

### LINKED ISSUE #844: when no renewal is done, should return a different exit code
I am running lego using the goacme/lego docker image. 
I am setting a cronjob to renew my certificates every day and if a renewal actually takes place, I need to restart my webserver. I don't want to blindly restart the webserver if no renewal took place (because there was still a good amount of days left until renewal).

For this reason, I think it would be very useful if lego would return a non-zero exit code if it refused to do the renewal.

ldez:
Adds renew hook.

Fixes #427
Fixes #844


cmd/cmd_renew.go
```diff
@@ -194,3 +202,25 @@ func merge(prevDomains []string, nextDomains []string) []string {
...
+func renewHook(ctx *cli.Context) error {
+	hook := ctx.String("renew-hook")
+	if hook == "" {
+		return nil
+	}
+
+	ctxCmd, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+
+	parts := strings.Split(hook, " ")
```
wyattjoh: Maybe best to use this instead?

```suggestion
	parts := strings.Fields(hook)
```

ldez: Yes good suggestion!

PR #850, opened 2019-04-08
sakuracloud: supports concurrent update

yamamoto-febc:
The SakuraCloud API doesn't supports concurrent update for DNS records.

It causes a "lost update" when run in parallel.
(e.g.: when be called from [terraform-provider-acme](https://github.com/terraform-providers/terraform-provider-acme))

This PR solves this problem by using Mutex via latest libsacloud(SakuraCloud API library).


ldez:
Create a client inside client with another client that use `http.Client` seems as a design issue in the SakuraCloud lib: the `http.Client` is used without any configuration (so the timeouts are infinite) and the `http.Client` is neither exposed nor overridable nor configurable.

This creates a lot of complexity and this complexity is only useful for testing.

At least the code inside lego must be rethink with less complexity.

yamamoto-febc:
Hi @ldez, Thank you for reviewing!
  
I agree with what you wrote, it is libsacloud's problem.
I'm going to fix it and update this PR.

Also, I think we should make API call timeout configurable after this problem is fixed.
Should this be implemented with this PR? or should be another PR?

ldez:
Hi @yamamoto-febc you can add the configurable timeout in this PR.

FYI, I'm trying to simplify your tests, but I'll wait for your changes.

yamamoto-febc:
I have just updated this PR. It includes following updates:

- Use libsacloud v1.19(includes configurable http.Client)
- Add SAKURACLOUD_HTTP_TIMEOUT env variable for configure API request timeout
- Run `make generage-dns` for generate documents

PR #854, opened 2019-04-10
Adds 'Since' to DNS providers documentation

ldez:
Adds 'Since' to DNS providers documentation.

Allow to know when a DNS provider has been introduce.


ldez:
> Would breaking changes facilitate an update to the Since date?

I'm not sure to understand: do you speak about a way to create a "last updated date" ?

wyattjoh:
If a provider caused a breaking change, I suppose we could implement a different version with an updated `Since` date is all, but it's a problem for the future :)

PR #856, opened 2019-04-16
cloudns: fix TTL and status validation

ldez:
~~If I understand correctly the "not clear" documentation of ClouDNS, the List records operation returns an array: https://www.cloudns.net/wiki/article/57/~~

related to https://github.com/containous/traefik/issues/4772

ping @tzakrajs

### DIFF
--- a/providers/dns/cloudns/internal/client.go
+++ b/providers/dns/cloudns/internal/client.go
@@ -32,6 +32,8 @@ type TXTRecord struct {
 	Status   int    `json:"status"`
 }
 
+type TXTRecords map[string]TXTRecord
+
 // NewClient creates a ClouDNS client
 func NewClient(authID string, authPassword string) (*Client, error) {
 	if authID == "" {
@@ -117,7 +119,7 @@ func (c *Client) FindTxtRecord(zoneName, fqdn string) (*TXTRecord, error) {
 		return nil, err
 	}
 
-	var records []TXTRecord
+	var records TXTRecords
 	if err = json.Unmarshal(result, &records); err != nil {
 		return nil, fmt.Errorf("ClouDNS: TXT record unmarshaling error: %v: %s", err, string(result))
 	}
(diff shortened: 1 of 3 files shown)


steve-todorov:
@ldez Thank you for looking into this! Here's a JSON response from the API:

```
$ curl -s -H accept:application/json "https://api.cloudns.net/dns/records.json?auth-id=my-id&auth-password=a-password&domain-name=domain.com&type=txt" | jq
{
  "5769228": {
    "id": "5769228",
    "type": "TXT",
    "host": "",
    "record": "v=spf1 +a +mx include:_spf.google.com -all",
    "failover": "0",
    "ttl": "3600",
    "status": 1
  },
  "181805209": {
    "id": "181805209",
    "type": "TXT",
    "host": "_github-challenge",
    "record": "b66b8324b5",
    "failover": "0",
    "ttl": "300",
    "status": 1
  }
}

```

Please let me know if I could help with anything more.

ldez:
@steve-todorov  thank you, so it's not an array but a map.

ldez:
If fact the API returns `[]` when there is no record and an object when there are records.

Very strange API, I updated the behavior.

providers/dns/cloudns/internal/client.go
```diff
@@ -130,7 +135,7 @@ func (c *Client) FindTxtRecord(zoneName, fqdn string) (*TXTRecord, error) {
 		}
 	}
 
-	return nil, fmt.Errorf("ClouDNS: no existing record found for %q", fqdn)
+	return nil, nil
```
wyattjoh: Why aren't you returning an error here anymore?

ldez: I manage the `nil` inside the provider, seems like a better way to manage this part.

ldez: the method `FindTxtRecord` is only use for the clean up.

Issue #859, opened 2019-04-18
Problem to access certificates when run with Docker


edouardtheron:
I launched Lego from Docker. My request was accepted by Lets Encrypt, but I could not access the resulting certificates, as they seem stucked in the container.

I read in the doc that the certificates are stored in a `.lego` directory in the current working dir. However, as there is no `WORKDIR` specified within the `Dockerfile`, the working dir is `/` by default.

Therefore, it is impossible to fetch the content of the `.lego` directory with a volume mapping during `docker run ...`

Am I missing something? 

Thanks for your help.

taoru:
> Therefore, it is impossible to fetch the content of the `.lego` directory with a volume mapping during `docker run ...`

Use `docker run -v /path/to/where/you/want/it:/ssl ... goacme/lego ... --path /ssl run` to store your certificates in `/path/to/where/you/want/it`

edouardtheron:
Thank you @taoru I did not see this optional argument `--path` anywhere, must have missed it.

Zyprexia:
> > Therefore, it is impossible to fetch the content of the `.lego` directory with a volume mapping during `docker run ...`
> 
> Use `docker run -v /path/to/where/you/want/it:/ssl ... goacme/lego ... --path /ssl run` to store your certificates in `/path/to/where/you/want/it`

Could you add this mount point to the docker image/container documentation? This is almost vital information needed to use the docker image/container. Thanks!

PR #860, opened 2019-04-18
Update Dockerfile so we have a mappable `WORKDIR` at runtime

edouardtheron:
Ref  #859

Doc updated accordingly.


ldez:
According to #859 a WORKDIR is not needed.

PR #864, opened 2019-04-18
Check if the generation of DNS documentation has been successfully completed.

ldez:
Check if the generation of DNS documentation has been successfully completed.


ldez:
ping @wyattjoh

Makefile
```diff
@@ -65,8 +65,17 @@ docs-serve: generate-dns
...
-# Generate DNS
-.PHONY: generate-dns
+# DNS Documentation
+.PHONY: generate-dns validate-doc
 
 generate-dns:
 	go generate ./...
+
+validate-doc: generate-dns
+ifneq ($(shell git status --porcelain -- ./docs/ ./cmd/ 2>/dev/null),)
```
wyattjoh: After making a small modification to a file in `./cmd/` folder, and running `make validate-doc`, I get the error:

```
go generate ./...
The documentation must be regenerated, please use `make generate-dns`.
 M cmd/cmd_list.go
make: *** [validate-doc] Error 2
```

But after running `make generate-dns`, and then re-running `make validate-doc`, I still get the same error.

Is this only designed to be run in CI on a clean git state? It's not super clear otherwise.

ldez: Yes this mainly designed for the CI and on a clean git state, I know that it's not perfect.
If you have a better solution, I'll take it.

PR #880, opened 2019-05-05
manual: serial behavior.

### LINKED ISSUE #879: CLI instructions imply serial behavior in manual mode
## Description

I am trying to issue a SAN cert for the zone apex and wildcard on first level subdomains. For example:

	-d "example.com" -d "*.example.com"

The UI instructs me (in sequence) to update DNS TXT with two different validation codes:

	2019/05/05 06:27:25 [INFO] [*.example.com] acme: Preparing to solve DNS-01
	lego: Please create the following TXT record in your example.com. zone:
	_acme-challenge.example.com. 120 IN TXT "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
	lego: Press 'Enter' when you are done

	2019/05/05 04:24:01 [INFO] [example.com] acme: Preparing to solve DNS-01
	lego: Please create the following TXT record in your example.com. zone:
	_acme-challenge.example.com. 120 IN TXT "bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb"
	lego: Press 'Enter' when you are done

However, notice that both records are the same, for `_acme-challenge.example.com.`.

The behavior of `certbot` is to validate these records serially. When I ran through this process with `certbot` I would add "aaaaa" to DNS, verify by pressing Enter, then _replace_ it with "bbbbb", and verify the second one by pressing enter.

Based on https://github.com/go-acme/lego/issues/847#issuecomment-480968281 I learned that `lego` validates these **in parallel** after all the console messages have been printed. However, it does not say this anywhere in the CLI output!

It is currently **not obvious** that the DNS checks are done in parallel at the end, or that both TXT records should be returned simultaneously.  I could imply this from the log output, but only _after_ the failure occurs, and it's subtle.

I think the parallel behavior is great but it is not adequately communicated.

## Expected Behavior

_either_

1. `lego` **should** run the DNS checks serially, just as `certbot` does, and as the current behavior of the UI implies.

2. (preferred) `lego` **should** show both codes at the same time. `lego` **should** direct the user to add all TXT records at the same time. `lego` **should** explicitly indicate the records will be validated simultaneously.

I _think_ this only matters for the `--dns manual` case.

## Steps to repro

1. Try to issue SAN with zone apex and first-level wildcard subdomain, i.e: `lego --dns manual --server=https://acme-staging-v02.api.letsencrypt.org/directory -d "example.com" -d "*.example.com" -a -m mail@example.com run`
2. Update DNS record with first TXT verification code
3. Confirm propagation with `dig`
4. Press Enter to continue
5. Update DNS record with second TXT verfication code (only)
6. Confirm propagation with `dig`
7. Press Enter to continue
8. `lego` finds the second code but not the first code because it did not check
   the first code after step 4.

ldez:
Fixes #879

### DIFF
--- a/challenge/dns01/dns_challenge_manual.go
+++ b/challenge/dns01/dns_challenge_manual.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"fmt"
 	"os"
+	"time"
 )
 
 const (
@@ -49,4 +50,10 @@ func (*DNSProviderManual) CleanUp(domain, token, keyAuth string) error {
 	fmt.Printf(dnsTemplate+"\n", fqdn, DefaultTTL, "...")
 
 	return nil
+}
+
+// Sequential All DNS challenges for this provider will be resolved sequentially.
+// Returns the interval between each iteration.
+func (d *DNSProviderManual) Sequential() time.Duration {
+	return DefaultPropagationTimeout
 }


cbednarski:
@ldez Awesome! Thanks for the amazingly quick turnaround!

Issue #720, opened 2018-11-15
Creating Present/Cleanup methods providing just fqdn, value as arguments ?


mdbraber:
Hi,

I'm writing [acme-proxy](https://github.com/mdbraber/acme-proxy) to relay dns challenges using the httpreq provider and integrating with lego DNS providers. It's easy to write a "RAW" method that calls the `Present()` and `Cleanup()` methods using `domain`,` token` and `keyAuth` arguments.

I'd also like to write a version that integrates the "default" method that calls a method using just the `fqdn` and `value`. But there isn't a constructor that I can call for lego DNS providers with just those two values (and logically it's impossible to reverse `value` to a `keyAuth` as it's sha256. 

I'm wondering if anyone has a good suggestion to leverage the existing DNS providers. A complete "solution" would of course make a separate constructor with the `fqdn` and `value` parameter that is called by the 'outer' constructor that uses the `domain`, `token` and `keyAuth` parameters. That might solve the issue but probably this is considered overkill. OTOH having a constructor that can be called with just `fqdn` and `value` for DNS challenges might be useful for other uses cases too. 

I'm happy to rewrite the DNS providers if this would be interesting, but first would like to see if this would be the best way to go or other solutions might be better (or just not having a "default" mode for acme-proxy might be just best)

/cc @ldez

mdbraber:
@ldez any thoughts on this? I'm trying to figure out if there's a way to use lego if I only have the FQDN and Value (TXT record) parameters?

ldez:
I think this is a edge-case and that there are some design issues.
For example, the `Present / Cleanup` methods are defined by the `Provider` interface.

I think that the `RAW` mode allow to meet acme-proxy needs.

mdbraber:
@ldez thanks for your reply! The `RAW` mode is indeed great for *incoming* requests to (in this case) [acmeproxy](https://github.com/mdbraber/acmeproxy) using `httpreq` that are passed on to the `lego` `Present / CleanUp` methods.

My question is  how to make use of `lego` as a _backend_ when for creating requests which are issued through `acmeproxy`. As an example: when I want to use acmeproxy with [acme.sh](https://github.com/Neilpang/acme.sh) the following takes place:

1. solve a challenge with `acme.sh` using a new httpreq provider
2. send `fqdn, value` to acmeproxy
3. acmeproxy receives `fqdn, value`
4. acmeproxy creates TXT record based on `fqdn, value`

[...]

PR #883, opened 2019-05-07
Add CreateRecord/RemoveRecord methods

mdbraber:
Example of how to implement CreateRecord/RemoveRecord as per discussion in https://github.com/go-acme/lego/issues/720


mholt:
I actually think this would be really useful.

mdbraber:
> I actually think this would be really useful.

Thanks for chiming in. I was back working at this a bit and this works fine for my own provider (TransIP) and is easy to change in the current code. If there's any way I can help to move this along I'd be happy to support!

ldez:
I'm going to close this PR because at the moment I'm still thinking it's not the right abstraction for the challenges.
We can continue to discuss in the related issue.

PR #896, opened 2019-05-24
lego: add method to expose API core

vancluever:
This adds the Core function to the lego client object, allowing
access to the lower-level core API when it's needed.

### DIFF
--- a/lego/client.go
+++ b/lego/client.go
@@ -72,3 +72,8 @@ func (c *Client) GetToSURL() string {
 func (c *Client) GetExternalAccountRequired() bool {
 	return c.core.GetDirectory().Meta.ExternalAccountRequired
 }
+
+// Core exposes the low-level core API for this client.
+func (c *Client) Core() *api.Core {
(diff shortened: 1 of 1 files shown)


ldez:
Could you explain why you need this?

vancluever:
Background: A [recent issue ](https://github.com/terraform-providers/terraform-provider-acme/issues/54)in the Terraform ACME provider has caused local copies of certificates to be deleted. As far as I know, there's currently no function to recover a certificate resource from the API via its certificate URL right now, which will be necessary to repair the states once I address the actual bug.

If it's not optimal to expose the API, maybe a method can be added to Certifier to retrieve a certificate resource from CertURL? It should be pretty easy for me to pivot to that instead.

ldez:
I prefer to not expose the core, then your idea to add a method to the  `Certifier` seems better.

vancluever:
Closing this and will re-submit with the higher-level method on Certifier.

PR #897, opened 2019-05-25
Get an existing certificate by URL

vancluever:
This adds the Get function, which allows the ability for a Certifier
to get an existing certificate by URL.

The returned Resource includes all fields populated except for the
PrivateKey and CSR.


vancluever:
Thanks @ldez!

Issue #905, opened 2019-06-11
dnspod provider compile failures in go1.12.5


hjweddie:
i encounter compile failures when i use dnspod's dns provider and go1.12.5, and error messages are as bellow

```
vendor/github.com/go-acme/lego/providers/dns/dnspod/dnspod.go:140:19: cannot convert 0 (type untyped number) to type "encoding/json".Number
vendor/github.com/go-acme/lego/providers/dns/dnspod/dnspod.go:140:19: invalid operation: hostedZone.ID == 0 (mismatched types "encoding/json".Number and int)
```

when i compile it with go1.12.4, it success with no errors

ldez:
duplicate of #876  #827

Issue #876, opened 2019-05-03
Build failure: cannot convert 0 to type "encoding/json".Number w/ latest dnspod-go


eclipseo:
Golang 1.12.2 on Fedora Rawhide, dnspod-go master https://github.com/decker502/dnspod-go/commit/71fbbdbdf1a7eeac949586de15bf96d416d3dd63

```
# github.com/go-acme/lego/providers/dns/dnspod
../providers/dns/dnspod/dnspod.go:140:19: cannot convert 0 (type untyped number) to type "encoding/json".Number
../providers/dns/dnspod/dnspod.go:140:19: invalid operation: hostedZone.ID == 0 (mismatched types "encoding/json".Number and int)
FAIL	github.com/go-acme/lego/cmd [build failed]
```

Latest version of dnspod-go uses json.Number instead of int for its ID field. Base type of json.Number is string, see https://golang.org/pkg/encoding/json/#Number

ldez:
duplicate of #827

Issue #827, opened 2019-03-14
dnspod provider: mismatched types


wI2L:
https://github.com/go-acme/lego/blob/0c87df143e630a1c50b4c36c8fbdda6cd993ebba/providers/dns/dnspod/dnspod.go#L140

`hostedZone` is a `dnspod.Domain` type which contain a field ID of type `json.Number`, a string.
```
go build: invalid operation: hostedZone.ID == 0 (mismatched types "encoding/json".Number and int)
```

ldez:
Hello, you are probably using a newer version than the one used in lego

wI2L:
Yeah that make sense, I am using go mod to manage my dependencies.
Using `go mod edit` to require the exact revision `83a3ba562b048c9fc88229408e593494b7774684` lego is using with dep fix the problem.

Any plans to move to go mod btw ?

ldez:
There is already a PR on the subject https://github.com/go-acme/lego/pull/706

For the moment I prefer to keep dep as long as the support of the modules are not the default.
Some tooling are still missing in the core GoLang distribution.
I know that community tools exist (I made one myself: [motoko](https://github.com/ldez/motoko).

https://blog.golang.org/modules2019

I'll close this issue, because I think the question is answered.

KingzCheung:
I also encountered this error.
Which version should I use to avoid this problem?

badouralix:
Hey @kingzcheung, in case your question is still relevant, here are my issues as of today:

```text
# github.com/go-acme/lego/providers/dns/designate
$GOPATH/pkg/mod/github.com/go-acme/lego@v2.6.0+incompatible/providers/dns/designate/designate.go:206:3: cannot use record.TTL (type int) as type *int in field value
# github.com/go-acme/lego/providers/dns/dnspod
$GOPATH/pkg/mod/github.com/go-acme/lego@v2.6.0+incompatible/providers/dns/dnspod/dnspod.go:140:19: cannot convert 0 (type untyped number) to type "encoding/json".Number
$GOPATH/pkg/mod/github.com/go-acme/lego@v2.6.0+incompatible/providers/dns/dnspod/dnspod.go:140:19: invalid operation: hostedZone.ID == 0 (mismatched types "encoding/json".Number and int)
```

And the current fix: (manually locking a couple of dependencies)

```bash
go mod edit -require=github.com/decker502/dnspod-go@v0.0.0-20180416134550-83a3ba562b04
[...]

PR #906, opened 2019-06-14
fix: json.Number compare with int of Domain.ID

Secbone:


ldez:
See #905,#876 #827

The problem is already fixed in #877

PR #923, opened 2019-07-07
Handles write errors when displaying DNS help.

ldez:
Handles write errors when displaying DNS help.


ldez:
I understand the point but I want to remove errors management during the writes and `fmt.Fprintln` returns an error.

Issue #882, opened 2019-05-07
Best practice to automatically renew all certificates?


mdbraber:
I've got a host where I have multiple certificates all under the --path /etc/lego. Is there a best practice for renewing those all at once with a single command (similar to certbot-renew)?

The easiest would probably be write a small script to cycle through all the domains in that directory (or use `lego list`?) and the call `certbot renew --path /etc/lego -d <domain>` for each domain. Is there a recommended method to do this?

beppler:
Today I'm using the follwing bash script:

```bash
#!/bin/sh

lego_bin=/opt/lego/lego
lego_path=/etc/lego
lego_certs="$lego_path/certificates"
account_email="gti@mps.com.br"
key_type=rsa4096
renew_days=30

reload=0

for domain in ` "$lego_bin" --path "$lego_path" list | grep "Certificate Name: " | awk '{print $3}' `; do
    original=$(date -r "$lego_certs/$domain.crt")

    "$lego_bin" \
          --path "$lego_path" \
          --email "$account_email" --accept-tos \
          --key-type "$key_type" \
          --http --http.port 127.0.0.1:4080 \
          --pem \
          --domains "$domain" \
          renew --days "$renew_days"

    actual=$(date -r "$lego_certs/$domain.crt")

    if [ "$original" != "$actual" ]; then
        reload=1
    fi
done

if [ $reload -eq 1 ]; then
    service nginx reload
fi

exit $?
```

Note that after renew any certificate my nginx is reloaded to use the renewed certificate.

I'm not using the renew-hook options because I want that nginx is reload once after renew all the certificates.

ldez:
With lego v5, you can use a configuration file to handle all your certificates.

https://go-acme.github.io/lego/references/ref-file/index.html

PR #925, opened 2019-07-12
Adds --name flag to list command.

beppler:
This flag causes the list command to output only the common name of the certificates.

This is useful to make scripts to renew all certificates that is about to expire as requested on #882.


beppler:
I'm using another way to list the certificates, so this pull request is not necessary anymore.

edgan:
Can we revisit this? Having to depend on the exact formatting of list using grep and awk is not ideal. With this PR it would be perfect. I also just tested it against master. It merged cleanly, compiled, and works.

beppler:
Hi, I pushed it again on my fork, but I can not reopen the pull request for some reason.

PR #927, opened 2019-07-16
joker: support username and password.

### LINKED ISSUE #926: Joker does NOT offer an API key
Joker only offers API keys for resellers, so end users are not able to use Joker for doing DNS validation for Lets Encrypt.

ldez:
Fixes #926

### DIFF
--- a/providers/dns/joker/client.go
+++ b/providers/dns/joker/client.go
@@ -70,7 +70,20 @@ func (d *DNSProvider) login() (*response, error) {
 		return nil, nil
 	}
 
-	response, err := d.postRequest("login", url.Values{"api-key": {d.config.APIKey}})
+	var values url.Values
+	switch {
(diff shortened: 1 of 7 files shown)


providers/dns/joker/client.go
```diff
@@ -70,7 +70,20 @@ func (d *DNSProvider) login() (*response, error) {
 		return nil, nil
 	}
 
-	response, err := d.postRequest("login", url.Values{"api-key": {d.config.APIKey}})
+	var values url.Values
+	switch {
+	case d.config.Username != "" && d.config.Password != "":
```
wyattjoh: Personally not a fan of using switch like this rather than `if {} else if {} else {}`, but this isn't a blocker.

wyattjoh:
Review CI errors relating to linting.

PR #932, opened 2019-07-21
vultr: invalid record type.

### LINKED ISSUE #931: Vultr API call failed by Invalid record type in v2.7.0
I got API error when using Vultr DNS in lego v2.7.0. Lego v2.6.0 works without problem.

```
acme: error presenting token: vultr: API call failed: Invalid record type
```

[govultr API](https://github.com/vultr/govultr/blob/2def36bf8b1c11c33a73fd75a0ecc337e5caf619/dns_records.go#L35) takes type as the 3rd argument.

```go
func (d *DNSRecordsServiceHandler) Create(ctx context.Context, domain, recordType, name, data string, ttl, priority int) error {
```

while [the call](https://github.com/go-acme/lego/blob/v2.7.0/providers/dns/vultr/vultr.go#L91) seems to specify type as the 4th argument.

```go
	err = d.client.DNSRecord.Create(context.Background(), zoneDomain, name, "TXT", value, d.config.TTL, 0)
```

ldez:
Fixes #931 

CC @ddymko

### DIFF
--- a/providers/dns/vultr/vultr.go
+++ b/providers/dns/vultr/vultr.go
@@ -79,16 +79,18 @@ func NewDNSProviderConfig(config *Config) (*DNSProvider, error) {
 
 // Present creates a TXT record to fulfill the DNS-01 challenge.
 func (d *DNSProvider) Present(domain, token, keyAuth string) error {
+	ctx := context.Background()
+
 	fqdn, value := dns01.GetRecord(domain, keyAuth)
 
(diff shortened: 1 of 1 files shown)


ddymko:
Thanks for fixing this! (sorry for not noticing it in the original implementation) I just checked the other govultr calls and made sure those are being called correctly and those are all good.

PR #952, opened 2019-08-19
Support LEGO_EXPERIMENTAL_CNAME_SUPPORT validation in duckdns provider

romanlum:
There is a experimental feature called CNAME validation (LEGO_EXPERIMENTAL_CNAME_SUPPORT) where you can create a cname record pointing to another domain where the txt record has to be created.

The duckdns provider is always using the validation domain name to create the record. If you use the provider with cname valiation you have a fixed domain where you have to set the txt record.

I created a new config entry DUCKDNS_DOMAIN that points to the domain where the txt record should be set. The value is only used, when it is not empty.

#1580

### DIFF
--- a/providers/dns/duckdns/duckdns.go
+++ b/providers/dns/duckdns/duckdns.go
@@ -14,6 +14,7 @@ import (
 
 // Config is used to configure the creation of the DNSProvider
 type Config struct {
+	DomainName         string
 	Token              string
 	PropagationTimeout time.Duration
 	PollingInterval    time.Duration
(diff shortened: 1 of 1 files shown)


ldez:
Hello,

Lego is a CLI but also a lib, as `DUCKDNS_DOMAIN` is a unique value it's not possible to manage several domains.

You have to find another approach.

romanlum:
That´s the key of the CNAME validation to use a single domain txt record to validate other domains.
Maybe i can add a check if cname validation is used.

Here is my usecase:
I have a domain registered at df.eu. They do not have a dns api, so i am using duckdns to validate my df.eu domains by adding a cname-record to duckdns.

```
_acme-challenge.test.domain.at | CNAME |   | acme-cert.duckdns.org | 
_acme-challenge.test1.domain.at | CNAME |   | acme-cert.duckdns.org |
```

With my code change and DUCKDNS_DOMAIN set to acme-cert.duckdns.org the txt record is set to the duckdns domain and by using cname-validation let`s encrypt follows the cname record and validates the duckdns txt record.

PR #961, opened 2019-09-06
linodev4: propagation timeout configuration.

### LINKED ISSUE #960: linodev4: Random propagation timeouts
Ubuntu 18.04. Using the "linodev4" DNS authenticator.

When I try to generate a new certificate I see a message like
```
"[INFO] Wait for propagation [timeout: 9m0s, interval: 15s]"
```
Where is that "timeout" number coming from? If I stop (control-c) and restart the the program, the "timeout" seemingly picks up where it left off? Very mysterious.

I need that number to always be > 15 minutes for Linode's DNS servers (very long propagation times).

Command :

```
LINODE_TOKEN_FILE=/etc/lego/linode.key lego --dns="linodev4" --path="/etc/lego" --email="me@example.com" --accept-tos --domains='example.com' --domains='*.example.com' run
```

I've experimented with different  valules for environment variables : `LINODE_HTTP_TIMEOUT` and `LINODE_POLLING_INTERVAL`

Also different values for options : `--http-timeout` and `--dns-timeout`

None seems to have an effect on the "timeout".

ldez:
Allow to configure the propagation timeout for Linode v4.

Fixes #960

### DIFF
--- a/providers/dns/linodev4/linodev4.go
+++ b/providers/dns/linodev4/linodev4.go
@@ -24,18 +24,20 @@ const (
 
 // Config is used to configure the creation of the DNSProvider
 type Config struct {
-	Token           string
-	PollingInterval time.Duration
-	TTL             int
-	HTTPTimeout     time.Duration
+	Token              string
+	PropagationTimeout time.Duration
+	PollingInterval    time.Duration
+	TTL                int
+	HTTPTimeout        time.Duration
 }
 
 // NewDefaultConfig returns a default configuration for the DNSProvider
 func NewDefaultConfig() *Config {
 	return &Config{
-		PollingInterval: env.GetOrDefaultSecond("LINODE_POLLING_INTERVAL", 15*time.Second),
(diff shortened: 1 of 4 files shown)


ldez:
Require #962

providers/dns/linodev4/linodev4.go
```diff
@@ -97,19 +99,22 @@ func NewDNSProviderConfig(config *Config) (*DNSProvider, error) {
...
-	minsRemaining := dnsUpdateFreqMins - (time.Now().Minute() % dnsUpdateFreqMins)
-
-	timeout = (time.Duration(minsRemaining) * time.Minute) +
-		(minTTL * time.Second) +
-		(dnsUpdateFudgeSecs * time.Second)
-	interval = d.config.PollingInterval
-	return
+func (d *DNSProvider) Timeout() (time.Duration, time.Duration) {
+	timeout := d.config.PropagationTimeout
+	if d.config.PropagationTimeout <= 0 {
```
wyattjoh: Why not this instead?

```suggestion
	if timeout <= 0 {
```

PR #965, opened 2019-09-12
feat: expose more SOA fields found by dns01.FindZoneByFqdn

### LINKED ISSUE #964: Feature request: expose SOA record found by dns01.FindZoneByFqdn
I'm using lego as library to get certificates both via DNS Update (RFC2136) and via Cloudflare API. At some point, I need to make the decision, which dns01 challenge provider I should use, for which I need to find the SOA RR's MNAME (i.e. the primary NS).

<details><summary>Background information</summary>

---

I've build a CLI modelled after [acmetool](https://github.com/hlandau/acme), which takes a config file like this:

```yaml
challenge-providers:
- type:    cloudflare
  primary: adi.ns.cloudflare.com
  # ... credentials
- type:    rfc2136
  primary: sns.dns.icann.org
  # ... tsig params
- type:    rfc2136
  primary: ns.company.tld
  # ... tsig params

certificates:
- foo.domain1.tld         # managed by cloudflare
- example.com             # managed by sns.dns.icann.org
- foo.intern.company.tld  # managed by ns.company.tld
```

and creates a certificate bundle.

I have a custom `challenge.Provider`, which inspects the `domain` given to `Present()`/`CleanUp()` and multiplexes to the matching challenge providers found in the config file.

---

</details>

To lookup the zone for a domain, I found [`dns01.FindZoneByFqdn`](https://github.com/go-acme/lego/blob/v3.0.2/challenge/dns01/nameserver.go#L101-L109), which get's me half the way to my goal. I could then query the zone's SOA record and read out the `(*dns.SOA).Ns` attribute. However, at this point lego [did aleady retrieve the SOA record](https://github.com/go-acme/lego/blob/v3.0.2/challenge/dns01/nameserver.go#L149) just before returning the zone.

Not only would I do duplicate work, I'd also need to setup basically the same code found in `nameserver.go` (preparing recursive nameservers, preparing, exchanging and parsing DNS messages, etc.; one notable difference would be around line 150, where I'd return `soa.Ns` instead of `soa.Hrd.Name`).

Hence, I propose/kindly request the following changes (bikeshedding please):

1. change `fqdnToZone` from `map[string]string` to `map[string]*dns.SOA` (or `map[string]struct{ zone, ns string }`
   - maybe rename it (and accompanying mutex) to somthing better suited
   - modify the setter/getter accordingly (ll. 34, 114f, 150f)
   - having a `map[string]*dns.SOA` allows to automatically bust the cache according to the value's Expires field
2. add a `FindPrimaryNsByFqdn` (and `FindPrimaryNsByFqdnCustom)` function, retrieving the cache entry's MNAME.
   - it'll share a lot of code with `FindZoneByFqdnCustom`, which can be moved into a separate function

What do you think? I could prepare a PR to discuss the details "on the living object" (but I'd rather not, if this is far out of scope for lego).

dmke:
Having a domain's primary nameserver is handy in some situations.
This reduces query overhead by caching not only the domain's zone,
but also the primary NS (of the same underlying SOA record).

Additional functions FindPrimaryNsByFqdn and FindPrimaryNsByFqdnCustom
can be used to access the same SOA record cache.

fixes #964


dmke:
Rebased onto master.

dmke:
Oh, oops. I forgot this PR was based on `diginoe:master`... I'll reopen this with a proper branch in a second...

PR #972, opened 2019-09-19
ovh: fix int overflow.

ldez:
On armv7, there is a int overflow during the unmarshall:

```
ERRO[2019-09-19T12:34:04Z] Unable to obtain ACME certificate for domains "mydomain.com,*.mydomain.com" : unable to generate a certificate for the domains [mydomain.com *.mydomain.com]: acme: Error -> One or more domains had a problem:
[*.mydomain.com] [*.mydomain.com] acme: error presenting token: ovh: error when call api to add record (/domain/zone/mydomain.com/record): json: cannot unmarshal number 5025119774 into Go struct field Record.id of type int
[mydomain.com] [mydomain.com] acme: error presenting token: ovh: error when call api to add record (/domain/zone/mydomain.com/record): json: cannot unmarshal number 5025119775 into Go struct field Record.id of type int  providerName=letsencrypt.acme
```

In this case (32bits arch and unmarshall), the max value of a `int` is 2147483647


Diaoul:
Awesome! Looking forward a new release being included in traefik image.
Might take a while though

PR #981, opened 2019-10-04
challenge/dns01: pass challenge domain to the Present/CleanUp methods

tmm1:
When requesting certs for "abc.example.com", "*.abc.example.com" together, two DNS challenges are required with the same domain "_acme-challenge.abc.example.com". The two TXT values must be added together as separate records on the same domain.

Currently the Present/CleanUp callbacks are invoked twice with "abc.example.com" as the first argument in both invocations. This makes it hard to distinguish the two challenges and clean them up correctly.

With this change the second invocation of Present/CleanUp uses "*.abc.example.com" as the first argument as expected.

### DIFF
--- a/challenge/dns01/dns_challenge.go
+++ b/challenge/dns01/dns_challenge.go
@@ -91,7 +91,7 @@ func (c *Challenge) PreSolve(authz acme.Authorization) error {
 		return err
 	}
 
-	err = c.provider.Present(authz.Identifier.Value, chlng.Token, keyAuth)
+	err = c.provider.Present(domain, chlng.Token, keyAuth)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


ldez:
Hello,

You cannot change the signature of `Present` and `Cleanup`

ldez:
> When requesting certs for "abc.example.com", "*.abc.example.com" together, two DNS challenges are required with the same domain "_acme-challenge.abc.example.com". The two TXT values must be added together as separate records on the same domain.

In fact in this case two TXT record must be added for `_acme-challenge.abc.example.com`.

The most of the DNS provider support to set multiple TXT records for one domain.
For the provider that doesn't support multiple TXT records for one domain we use the "sequential" solver.

PR #974, opened 2019-09-24
feat: ease operation behind proxy servers

dmke:
fixes #973

PR #982, opened 2019-10-05
feat: ease operation behind proxy servers

### LINKED ISSUE #973: Running behind a reverse proxy (http-01)
Hi,

when lego runs behind a reverse proxy, [its DNS rebind attack prevention](https://github.com/go-acme/lego/blob/v3.0.2/challenge/http01/http_challenge_server.go#L67) causes the ACME validation to fail:

```console
$ lego ... --http --http.port 127.0.0.1:402
...
[WARN] Received request for domain 127.0.0.1 with method GET but the
  domain did not match any challenge. Please ensure your are passing the
  HOST header properly.
...
```

This is because a typical proxy directive for Apache will pass the [original `Host` header as `X-Forwarded-Host`](https://httpd.apache.org/docs/2.4/mod/mod_proxy.html#x-headers), and change the pass a new `Host` header containing the host of the proxied URL:

```apache
ProxyPass "/.well-known/acme-challenge/" "http://127.0.0.1:402/.well-known/acme-challenge/"
```

(NGINX does the same, other servers might behave similiar).

For Apache, there's a workaround in form of the [`ProxyPreserveHost`](https://httpd.apache.org/docs/2.4/mod/mod_proxy.html#proxypreservehost), but that might confuse some legacy applications.

In this particular setup I'd like to have the option to change the header name used to implement DNS rebind prevention, i.e. sth. similar to:

```console
$ lego ... --http --http.port 127.0.0.1:402 --http.proxy-header X-Forwarded-Host
``` 

To make matters worse (or [better?](https://xkcd.com/927/)), [RFC 7239](https://tools.ietf.org/html/rfc7239) tries to standardize the various proxy/forwarding headers into a single `Forwarded` header, but with different semantics.

I'm currently working on a PR addressing this issue ([WIP here](https://github.com/go-acme/lego/compare/master...digineo:http01-proxy)), but that's not finished/polished yet. I would like to have some early feedback though (in particular, whether I should move the `Fowarded` header parsing into a separate package) :-)

dmke:
fixes #973 

This recreates #974 because of repo issues.

### DIFF
--- a/.golangci.toml
+++ b/.golangci.toml
@@ -62,6 +62,12 @@
     path = "challenge/http01/domain_matcher.go"
     text = "string `Host` has \\d occurrences, make it a constant"
   [[issues.exclude-rules]]
+    path = "challenge/http01/domain_matcher.go"
+    text = "cyclomatic complexity \\d+ of func `parseForwardedHeader` is high"
+  [[issues.exclude-rules]]
+    path = "challenge/http01/domain_matcher.go"
+    text = "Function 'parseForwardedHeader' has too many statements"
+  [[issues.exclude-rules]]
     path = "challenge/tlsalpn01/tls_alpn_challenge.go"
     text = "`idPeAcmeIdentifierV1` is a global variable"
   [[issues.exclude-rules]]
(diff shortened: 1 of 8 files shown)


ldez:
Closed in favor of #974

PR #985, opened 2019-10-07
cloudflare: handle restricted API tokens

### LINKED ISSUE #984: Cloudflare and restricted API tokens
## Context

Cloudflare has two API access modes:

1. API key (i.e. global access to everything related to a user's account), and
2. API token (i.e. scoped access to selected resources).

The latter mode is available since [at least Aug 30, 2019](https://blog.cloudflare.com/api-tokens-general-availability/) promoting the principle of least privilege. Lego has support for both access modes (`CLOUDFLARE_EMAIL` + `CLOUDFLARE_API_KEY` since v0.3.0, and `CLOUDFLARE_API_TOKEN` since v3.1, #937).

To create a API token with minimal scope, the following steps are needed:

- log into their dashboard, go to account settings, create API token
- grant the following permissions:
  - Zone / Zone / Read
  - Zone / DNS / Edit
- select resources (zone names for the domains for which Lego should obtain certificates):
  - Include / Specific zone / *zone name*

## Problem

When selecting the domains, you can alternatively choose "Include / All zones from an account / *account name*". Following the principle of least privilege, I naturally chose *not* to do this. Host A should not be able to change DNS settings for host B.

However, here a problem arises: in order for Lego to `Present()`/`CleanUp()` DNS challenge tokens, it needs to lookup the zone ID. Cloudflare's Go client library has a dedicated method for this job: [`(*cloudflare.API).ZoneIDByName`](https://github.com/cloudflare/cloudflare-go/blob/master/cloudflare.go#L138-L139), so that's what Lego uses.

Internally, `ZoneIDByName()` queries the `/client/v4/zones` endpoint (with `name` parameter) to search *all* zones and filter for the apex name. Given a restricted scope for the API token, this endpoint access fails with the following error message:

> Actor `com.cloudflare.api.token.<token id omitted>` requires permission
> `com.cloudflare.api.account.zone.list` to list zones

## Workaround

If I change the API token scope to "include all zones from account xyz", the listing succeeds, `ZoneIDByName` can lookup the ID, and Lego can proceed to create/delete the DNS records.

This however defeats the purpose of having scoped API tokens in the first place (at least a bit, their API has many more endpoints): Now host A can change the DNS settings for host B (and vice versa).

*(Note: This propably applies only to users with multiple zones and/or domains spread to multiple machines, running more than one Lego instance...)*

## Cloudflare's reaction

I've asked the Cloudflare support whether this issue could be solved in their API (the zone endpoint should allow listing zones for which Zone/Zone/Read permissions were granted), and they responded with the following:

> [...] the permission to access the zone does not equal permission to list it.
> [...] our engineers have come up with a different solution for you to try.
>
>> The Lego client should implement a way to optionally provide a Zone ID, which would skip the need for a lookup. That way, there would be no problem with insufficient permissions. Zone ID is immutable for the life of the zone within an account, so this is something that has no problem being put into a config.

## Implementation details

The proposed solution could be implemented with a whitespace-delimited list of `NAME=ID` pairs as input for an internal `map[string]string`, which takes precedence over `ZoneIDByName`.

Having whitespace (including line breaks) allows the name/ID pairs to be read directly from an `CLOUDFLARE_ZONE_MAP` envionment variable, or to be neatly placed in a file (one entry per line) and accessed through a `CLOUDFLARE_ZONE_MAP_FILE` pointer. I believe, Lego already uses a similar mechanism for credentials, doesn't it?

```console
$ export CLOUDFLARE_API_TOKEN=xxxx CLOUDFLARE_ZONE_MAP="example.com=yyyy example.org=zzzz"
$ lego --dns cloudflare -d example.com -d www.example.com -d example.org -d www.example.org ...

## or

$ echo  >cf.map example.com=yyyy
$ echo >>cf.map example.org=zzzz
$ export CLOUDFLARE_ZONE_MAP_FILE="$(pwd)/cf.map"
$ lego --dns cloudflare -d example.com ...
```

The internal map has an additional benefit: it can act as a cache, even if the user didn't provide a name/ID mapping to begin with (or has chosen to use an API key with wider privileges):

Currently `CleanUp()` calls `ZoneIDByName()` again, after `Present()` did it a few moments earlier. The result would in all likelihood be the same as the previous call (making it redundant), as one would have to destroy and recreate the zone at Cloudflare (and in doing so disturbing the ACME dance, which would cause the validation to fail anyway). Unless I'm not aware of some insane use case which does exactly this, I assume we can memoize the result of the first `ZoneIDByName()` call :-)

## Alternative solution

It is also possible to create a pair of API tokens:

- one with Zone/Zone/Read privileges (for all Zones; no other privileges), shared among multiple Lego instances, and
- one with Zone/DNS/Edit privileges (for a selected list of Zones), private to each Lego instance.

I've found naming (and documenting) this harder, although the implementation is straight forward as well.

@ldez, @mholt: What do you think?

dmke:
Work in progress. Todo:

- [x] make test pass
- [x] add test specific to this change
- [x] figure out, how to live test
- [x] perform live test
- [x] update documentation

See #984 for discussion.

fixes #984


providers/dns/cloudflare/cloudflare.go
```diff
@@ -57,7 +65,8 @@ func NewDNSProvider() (*DNSProvider, error) {
 	if err != nil {
 		var errT error
 		values, errT = env.GetWithFallback(
-			[]string{"CLOUDFLARE_API_TOKEN", "CF_API_TOKEN"},
+			[]string{"CLOUDFLARE_DNS_API_TOKEN", "CF_DNS_API_TOKEN"},
+			[]string{"CLOUDFLARE_ZONE_API_TOKEN", "CF_ZONE_API_TOKEN", "CLOUDFLARE_DNS_API_TOKEN", "CF_DNS_API_TOKEN"},
```
ldez:
```suggestion
			[]string{"CLOUDFLARE_ZONE_API_TOKEN", "CF_ZONE_API_TOKEN"},
			[]string{"CLOUDFLARE_DNS_API_TOKEN", "CF_DNS_API_TOKEN"},
```

dmke: Hm, this makes `CLOUDFLARE_ZONE_API_TOKEN` mandatory, doesn't it?

ldez: yes but your approach will not work.

providers/dns/cloudflare/cloudflare.go
```diff
@@ -57,17 +70,25 @@ func NewDNSProvider() (*DNSProvider, error) {
...
 		if errT != nil {
 			return nil, fmt.Errorf("cloudflare: %v or %v", err, errT)
 		}
+		if z := env.GetOrFile("CLOUDFLARE_ZONE_API_TOKEN"); z != "" {
+			values["CLOUDFLARE_ZONE_API_TOKEN"] = z
+		} else if z = env.GetOrFile("CF_ZONE_API_TOKEN"); z != "" {
+			values["CLOUDFLARE_ZONE_API_TOKEN"] = z
+		} else {
+			values["CLOUDFLARE_ZONE_API_TOKEN"] = values["CLOUDFLARE_DNS_API_TOKEN"]
+		}
```
dmke: @ldez This is more what I had in mind when I previously used the `env.GetWithFallback` cascade:

1. try `CLOUDFLARE_ZONE_API_TOKEN`
2. if that's empty, try its alias `CF_ZONE_API_TOKEN`
3. it that's empty as well, fall back to `CLOUDFLARE_DNS_API_TOKEN`

Is there a better suited method in the `env` package, I just have not seen?

ldez: Seems too specific for now to create a method in the `env` package.

providers/dns/cloudflare/cloudflare.go
```diff
@@ -42,13 +46,22 @@ func NewDefaultConfig() *Config {
...
 // NewDNSProvider returns a DNSProvider instance configured for Cloudflare.
-// Credentials must be passed in the environment variables:
-// CLOUDFLARE_EMAIL, CLOUDFLARE_API_KEY, CLOUDFLARE_API_TOKEN.
+// Credentials must be passed in as environment variables:
+//
+// Either provide CLOUDFLARE_EMAIL and CLOUDFLARE_API_KEY,
+// or a CLOUDFLARE_DNS_API_TOKEN.
+// For a more paranoid setup, provide CLOUDFLARE_DNS_API_TOKEN and CLOUDFLARE_ZONE_API_TOKEN.
+//
+// Please see https://go-acme.github.com/lego/dns/cloudflare/ for a more detailed description.
```
dmke: Is it OK to link to the website, or should I repeat what I wrote there?

ldez: I prefer to not link to the website in the go code.

providers/dns/cloudflare/cloudflare.toml
```diff
@@ -2,7 +2,7 @@ Name = "Cloudflare"
 Description = ''''''
 URL = "https://www.cloudflare.com/dns/"
 Code = "cloudflare"
-Since = "v0.3.0"
+Since = "v0.3.0 (API tokens since v3.1.0)"
```
ldez: Could you move `(API tokens since v3.1.0)` inside the env vars description.

dmke: Done.

providers/dns/cloudflare/cloudflare.go
ldez: but you don't need to repeat the documentation.

dmke: Okay, I'm trying to condense the information a bit more.

dmke:
I have performed a live test like so:

1. Add a domain to Cloudflare (example.com)
1. Create three API tokens:
   1. Name: `lego-all`, permissions: Zone:Read + DNS:Edit, scope: all
   2. Name: `lego-zones`, permissions: Zone:Read, scope: all
   3. Name: `lego-dns`, permissions: DNS:Edit, scope: specific zone (example.com)
1. Change `envTest` in `dns/provider/cloudflare/cloudflare_test.go` to `tester.NewEnvTest("CLOUDFLARE_DNS_API_TOKEN")`.
1. Run, and check that a `_acme-challenge.example.com` TXT record was created:

   ```console
   $ CLOUDFLARE_DNS_API_TOKEN="token lego-all" \
     CLOUDFLARE_DOMAIN=example.com \
     go test -v github.com/go-acme/lego/v3/providers/dns/cloudflare -run TestLivePresent
   ```
1. Run, and check that the `_acme-challenge.example.com` TXT record was deleted:

   ```console
   $ CLOUDFLARE_DNS_API_TOKEN="token lego-all" \
     CLOUDFLARE_DOMAIN=example.com \
     go test -v github.com/go-acme/lego/v3/providers/dns/cloudflare -run TestLiveCleanup
   ```
1. Change `envTest` in `dns/provider/cloudflare/cloudflare_test.go` to `tester.NewEnvTest("CLOUDFLARE_DNS_API_TOKEN", "CLOUDFLARE_ZONE_API_TOKEN")`.
1. Run, and check that a `_acme-challenge.example.com` TXT record was created:

   ```console
   $ CLOUDFLARE_DNS_API_TOKEN="token lego-dns" \
     CLOUDFLARE_ZONE_API_TOKEN="token lego-zones" \
     CLOUDFLARE_DOMAIN=example.com \
     go test -v github.com/go-acme/lego/v3/providers/dns/cloudflare -run TestLivePresent
   ```
1. Run, and check that the `_acme-challenge.example.com` TXT record was deleted:

   ```console
   $ CLOUDFLARE_DNS_API_TOKEN="token for lego-dns" \
     CLOUDFLARE_ZONE_API_TOKEN="token for lego-zones" \
     CLOUDFLARE_DOMAIN=example.com \
     go test -v github.com/go-acme/lego/v3/providers/dns/cloudflare -run TestLiveCleanup
   ```

The tests (4, 5, 7, 8) were executed successfully.

I would have liked to automate this, but that probably needs modifications to `tester.NewEnvTest` (to detect manged-but-empty env variables). I think this is out of scope for this issue.

providers/dns/cloudflare/cloudflare.go
dmke: Done.

PR #998, opened 2019-10-27
Fix Cloudflare Propagation

rouing:
If you attempt to get a certificate for your root domain "example" and your wildcard "*.example.com", they require the same _acme-challenge TXT DNS Record. It just so happens the tasks run in parallel too. You end up with some fun results like this:

`[*.****.me] acme: error: 403 :: urn:ietf:params:acme:error:unauthorized :: Incorrect TXT record "RootDomainKey" found at _acme-challenge.***.me, url:
[****.me] acme: error: 403 :: urn:ietf:params:acme:error:unauthorized :: Incorrect TXT record "WildcardDomainKey" found at _acme-challenge.****.me, url:
`

They are flipped.

Taking the Sequential function from the other providers provided a good fix.


rouing:
In the event this fix gets merged upstream and ALSO INTO TRAEFIK, this will fix the issue with duplicate TXT Records error when setting up wildcard LE DNS Prop.

This is a major issue that prevents Traefik and this tool from using Cloudflare DNS for Wildcards.

rouing:
https://github.com/containous/traefik/issues/5317

ldez:
Hello,

According to my knowledge Cloudflare supports multiple registration for the same domain, so use of `Sequential` is not appropriated.

`Sequential` slow down the challenge when you requested several domains.

So real solution, can be to change the way that records are deleted.

rouing:
You would be correct. I currently do not have the time to make changes necessary. If someone else decides to, +1 to them, else ill get around to it ASAP. Good catch.

PR #1003, opened 2019-11-05
fix: use token as unique ID.

ldez:
Use `token` instead of `fqdn` as unique ID to find record ID for:
- Aurora DNS
- Cloudflare
- Digital Ocean
- OVH

The token allows to distinct wildcard domain (`*.example.com`) and root domain (`example.com`)

Closes #998

### DIFF
--- a/e2e/challenges_test.go
+++ b/e2e/challenges_test.go
@@ -32,6 +32,7 @@ var load = loader.EnvLoader{
 }
 
 func TestMain(m *testing.M) {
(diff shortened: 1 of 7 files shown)


dmke:
LGTM, although not sure what the failing test on Travis is about.

ldez:
@dmke the problem was related to bad use of a private key in the tests.

Related to https://github.com/letsencrypt/pebble/issues/285

PR #1004, opened 2019-11-05
Update ACME RFC links from draft to RFC8555.

ldez:
Update ACME RFC links from draft to RFC8555.


acme/commons.go
```diff
@@ -111,7 +111,7 @@ type ExtendedOrder struct {
 }
 
 // Order the ACME order Object.
-// - https://tools.ietf.org/html/draft-ietf-acme-acme-16#section-7.1.3
```
dmke: Was removing the leading space deliberate?

ldez: no, it was a mistake

Issue #1011, opened 2019-11-15
use http but do not start an own server..


AubreyHewes:
Hi, sorry if this issue has been raised before (I could not find it)...

I would like to run the lego cli with `--http` but also something like `--no-server` and that the expected validation is written to a location that is already served by a running server (mapped well-known location).

We run multiple staged environments within docker containers, the staged env already runs a http server which is on http.. using `--http` fails due to the running server. we can not use another port as the corporate firewall does not allow external access to other ports (requesting opening an extra port for this kind of thing is a thing, changing the edge proxy is also a thing). 

Is it possible to use the cli `--http` to just write the validation to a folder which is already publicly available?  (.well-known)

Or does someone have an idea how I can implement this... or point me in a direction ;-) thanks!

cpu:
> I would like to run the lego cli with --http but also something like --no-server and that the expected validation is written to a location that is already served by a running server (mapped well-known location).

@AubreyHewes I think you want to use `--http.webroot` for this.

From the [cli usage docs](https://go-acme.github.io/lego/usage/cli/):

>    --http.webroot value         Set the webroot folder to use for HTTP based challenges to write directly in a file in .well-known/acme-challenge.

You should be able to run an existing webserver on port :80 and have lego write to `<webroot dir>/.well-known/acme-challenge` by running something like:

    ./lego --accept-tos -m <email> --http --http.webroot <webroot dir> -d <domain> run

cpu:
> Or does someone have an idea how I can implement this... or point me in a direction ;-)

alternatively, you could use a non-standard HTTP port for Lego with `--http.port` and then configure the existing webserver on :80 to redirect requests received to the path `/.well-known/acme-challenge/` to the non-standard Lego port.

>  (requesting opening an extra port for this kind of thing is a thing, changing the edge proxy is also a thing).

You wouldn't need to open the non-standard port to the wide-world for this plan but it would require changing the edge proxy to redirect the validation requests. Not sure if that's a hard sell for your environment or not :-)

AubreyHewes:
@cpu Cool thanks! I totally missed the `--http.webroot` as a replacement for starting an own server.. I just assumed it was the webroot for the started server!

PR #1012, opened 2019-11-15
Update examples to explain http.webroot

AubreyHewes:
The `--http.webroot` option does not properly explain that setting this _ONLY_ writes the token to the directory and _DOES NOT_ start a server. Also does not explain that the directory _MUST_ therefore be publicly available/hosted..

I have therefore tried to add this information to the examples documentation.. If not in a correct form please let me know how I should make it into a more intuitive/accepted description/example..

Edit: Also added a more explicit description to the CLI help output

#1011

### DIFF
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -70,7 +70,7 @@ func CreateFlags(defaultPath string) []cli.Flag {
 		},
 		cli.StringFlag{
 			Name:  "http.webroot",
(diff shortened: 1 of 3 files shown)


AubreyHewes:
If required to squash to one commit let me know!

AubreyHewes:
Sorry am slightly losing myself on creating a good readme for this.. :sheep:

PR #1019, opened 2019-12-11
Add support for freebsd/arm64

dmgk:


ldez:
Hello, could you give a more context?

dmgk:
@ldez Go 1.14 will support freebsd/arm64 and golang.org/x/net@83d349e8ac1a is an earliest commit that brings in necessary changes to allow lego port [1] to build on arm64 (we're currently testing it with go master).

[1] https://www.freshports.org/security/lego/

ldez:
ok thanks

PR #1028, opened 2020-01-04
Auto-detection of the project ID for gcloud.

james-lawrence:
can be autodetected from the environment on glcoud using the metadata
service.


providers/dns/gcloud/googlecloud_test.go
```diff
@@ -48,6 +49,8 @@ func TestNewDNSProvider(t *testing.T) {
 			envVars: map[string]string{
 				"GCE_PROJECT":              "",
 				"GCE_SERVICE_ACCOUNT_FILE": "",
+				// as Travis run on GCE, we have to alter env
+				"GCE_METADATA_HOST": "http://lego.wtf",
```
james-lawrence: hoboy.... that implies a security risk for travis....

ldez: not really because the vars env are saved and restore between each tests.

james-lawrence: i'd have to go play with it but if the GCE metadata is accessible in travis you can potentially pull ssh keys, instance metadata, etc. all of which can potentially be used to access travis' network.

james-lawrence:
np, ty for the library. =)

PR #1030, opened 2020-01-06
Successfully parse private key PEM blocks

### LINKED ISSUE #1029: Migrating from Certbot, lego can't parse some PEM-encoded keys
In the process of migrating thousands of certificates from a Certbot setup, I noticed that many of the private keys have PEM files that use these boundaries:

```
-----BEGIN PRIVATE KEY-----
-----END PRIVATE KEY-----
```

Lego fails to parse these because it is missing `RSA` or `EC`: https://github.com/go-acme/lego/blob/4f72de468dc3a3dbb083e7928a5065fabf7098b6/certcrypto/crypto.go#L80-L91

But the Go standard library does it differently, and succeeds: 
https://github.com/golang/go/blob/693748e9fa385f1e2c3b91ca9acbb6c0ad2d133d/src/crypto/tls/tls.go#L291-L308

(also see here, they loosely match on the substring "PRIVATE KEY": https://github.com/golang/go/blob/693748e9fa385f1e2c3b91ca9acbb6c0ad2d133d/src/crypto/tls/tls.go#L238)

Would you accept a PR to do something more like what the Go standard library does?

mholt:
This code is borrowed from the standard library; seemed like a good approach. Also adds support for ed25519 keys.

fix #1029

### DIFF
--- a/certcrypto/crypto.go
+++ b/certcrypto/crypto.go
@@ -80,19 +80,20 @@ func ParsePEMBundle(bundle []byte) ([]*x509.Certificate, error) {
 }
 
 // ParsePEMPrivateKey parses a private key from key, which is a PEM block.
+// Borrowed from Go standard library, to handle various private key and PEM block types.
+// https://github.com/golang/go/blob/693748e9fa385f1e2c3b91ca9acbb6c0ad2d133d/src/crypto/tls/tls.go#L291-L308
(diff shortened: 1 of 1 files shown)


ldez:
For the next time, could you use your fork instead the main repo, thanks.

mholt:
Awesome, thanks for the quick turnaround! (I'll try to keep a fork in mind next time)

PR #1041, opened 2020-01-11
namecheap: add sandbox support.

### LINKED ISSUE #1040: Add namecheap sandbox
I'm wondering if it's possible to add namecheap's sandbox so it can be used in dev environments.
https://www.namecheap.com/support/knowledgebase/article.aspx/763/63/what-is-sandbox

It would be the same as namecheap only difference I think is the api url 
    `https://api.sandbox.namecheap.com/xml.response`

One option would be to add an environment variable to current namecheap provider
`SANDBOX_MODE` (or something like that) be `false` by default and switching it to `true` swaps the url for us or something. Not sure what the best route would be.

api doc in case it's needed: https://www.namecheap.com/support/api/intro/

ldez:
Add sandbox support through the env var `NAMECHEAP_SANDBOX` (boolean)

Uses the production API URL by default.

Fixes #1040


jaycused:
@ldez thanks for the fast pr, I appreciate it!

PR #1044, opened 2020-01-15
designate: Don't clean up managed records like SOA and NS

stefreak:
The Designate plugin tries to clean up managed recordsets like SOA and NS.
In most Designate installations this is not allowed.

### DIFF
--- a/providers/dns/designate/designate.go
+++ b/providers/dns/designate/designate.go
@@ -240,7 +240,7 @@ func (d *DNSProvider) getRecord(zoneID string, wanted string) (*recordsets.Recor
 	}
 
 	for _, record := range allRecords {
-		if record.Name == wanted {
+		if record.Name == wanted && record.Type == "TXT" {
 			return &record, nil
 		}
 	}


ldez:
Hello,

Could you explain the state (WIP) of your PR?

stefreak:
@ldez thank you for looking at the PR. I was still testing the change.

Without this change, I get the following error message:

```
2020/01/16 10:10:34 [WARN] [*.ybk.cloud.syseleven.net] acme: error cleaning up: designate: error for _acme-challenge.ybk.cloud.syseleven.net. in CleanUp: Bad request with: [DELETE https://designate.cloud.syseleven.net:9001/v2/zones/10ce76b6-e702-439a-9e2e-b922a8560bde/recordsets/c00c6017-939d-4514-b637-3481be83474a], error message: {"message": "Managed records may not be deleted", "code": 400, "type": "bad_request", "request_id": "req-9ff909db-32c1-443e-9c1f-67714cd17c7f"}
```

with the change I could successfully get certificates using lego.

Would be really glad and happy user of lego if this could be merged!

providers/dns/designate/designate.go
```diff
@@ -240,7 +240,7 @@ func (d *DNSProvider) getRecord(zoneID string, wanted string) (*recordsets.Recor
 	}
 
 	for _, record := range allRecords {
-		if record.Name == wanted {
+		if record.Name == wanted && record.Type == "TXT" {
```
stefreak: @Idez if you prefer that, I can alternatively check if the record is not one of `SOA` or `NS` if you prefer.
```suggestion
		if record.Name == wanted && record.Type != "SOA" && record.Type != "NS" {
```

PR #1045, opened 2020-01-18
Add "LEGO_PATH" environment variable

hectorm:
Similar to [`CADDYPATH` in Caddy](https://caddyserver.com/v1/docs/cli) I think that lego should have an environment variable that allows setting the default path. This feature is especially useful in containers.

### DIFF
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -50,9 +50,10 @@ func CreateFlags(defaultPath string) []cli.Flag {
 			Usage: "(deprecated) Filename of the generated certificate.",
 		},
 		cli.StringFlag{
-			Name:  "path",
-			Usage: "Directory to use for storing the data.",
-			Value: defaultPath,
+			Name:   "path",
+			EnvVar: "LEGO_PATH",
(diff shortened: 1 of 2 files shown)


ldez:
Hello,

I'm not sure to see in which case this can be useful, because to use lego you have to use CLI flags.
The flag `--path` already allow to specify the path and it's possible to do `--path=${MYPATH}`

Could you give me concrete use cases?

hectorm:
Of course, the `--path` flag allows you to set this path and this PR doesn't add anything that couldn't be done before.

This proposal simply allows a base image to modify the default path to a value more suitable for container usage without the derived images having to be setting this path all the time.

hectorm:
I just saw your changes and that solution is absolutely better.

ldez:
To me, it looks like a edge case, I'm not really a fan of adding things only for a edge case.

But instead of hacking the default path, we can use the CLI library to get the job done, like that, it feels less like a case edge feature.

You have seen my change, that is fine with you so I am going to go on this approach.

PR #1050, opened 2020-01-30
route53: allow custom client to be provided

aidansteele:
This allows custom profiles, assumed roles, etc without modifying environment variables. Also instrumentation, custom proxies, etc.

Let me know if you want any changes made to suit project style, etc. Happy to make changes as desired!


aidansteele:
Thank you for reviewing the code and providing the feedback! I've made those changes now. Let me know if there's anything else.

PR #1053, opened 2020-02-08
zoneee: fix subdomains.

j005u:
The zoneeee dns provider fails with any kind of subdomain, due to the API being based on the zone name rather than the subdomain. 

I put together a simple patch to fix this based on work from other providers.

### DIFF
--- a/providers/dns/zoneee/client.go
+++ b/providers/dns/zoneee/client.go
@@ -8,6 +8,8 @@ import (
 	"io/ioutil"
 	"net/http"
 	"path"
+
+	"github.com/go-acme/lego/v3/challenge/dns01"
 )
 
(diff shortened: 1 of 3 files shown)


j005u:
Ah, sorry, forgot what I'd read last night in the contributors doc.

Fixed my formatting errors, however I'm pretty sure the zoneee tests are broken. I.E. they don't fail even though the client actually fails in usage with a subdomain such as something.batman.ee. Now when resolving the actual zone name, which fixes the dns provider for me in usage, the tests fail.

In reality, the client now works as expected for both domains and subdomains. Tested with production zone.ee credentials.

This is my first time touching go so I'm not capable enough to figure out what's wrong with the tests. Hoping someone can still put this to some use.

PR #1055, opened 2020-02-09
oraclecloud: fix subdomain support

### LINKED ISSUE #893: Oracle Cloud DNS challenge doesn't work correctly for subdomains
The Oracle Cloud DNS provider appears to work correctly for domains (e.g. example.com) but not subdomains (e.g. something.example.com).

I think the problem is the `zoneNameOrId` parameter to the patch domain records call: https://github.com/go-acme/lego/blob/a0d59b85fe4c7adcf9655ca16fbf2957f8998249/providers/dns/oraclecloud/oraclecloud.go#L99 Whilst `domain` is correct when generating a cert for example.com, when generating one for something.example.com, the zone name should generally be 'example.com'. 

I would suggest adding two additional _optional_ configuration variables: `OCI_ZONE_OCID` and `OCI_ZONE_NAME`. If `OCI_ZONE_OCID` is specified, it should be used for `zoneNameOrId`. Otherwise, if `OCI_ZONE_NAME` is specified, it should be used for `zoneNameOrId`. If neither are specified `domain` should be used as it is now.

chrisbulgaria:
~~#893 noticed oracle dns cannot be used for subdomains within a dns zone - only if the certificate is for the zone itself. Introduction of the ZONE_OCID and separating thereby the zone from the domain (which is still used if the zone_ocid is not provided and thereby providing backwards compatibility) solves this.~~

Fixes #893


ldez:
@chrisbulgaria Sorry for the late answer.

I think the problem can be solved in another way:

```go
// Present creates a TXT record to fulfill the dns-01 challenge
func (d *DNSProvider) Present(domain, token, keyAuth string) error {

	zoneNameOrID, err := dns01.FindZoneByFqdn(domain)
	if err != nil {
		return fmt.Errorf("oraclecloud: could not find zone for domain %q and fqdn %q : %w", domain, fqdn, err)
	}

  // ...
}
```

Could you validate that?

chrisbulgaria:
Hmm ... doesnt seem to work
error encountered while presenting token for DNS challenge: oraclecloud: could not find zone for domain "a.eks2.<redacted>.tk" and fqdn "_acme-challenge.a.eks2.<redacted>.tk." : could not find the start of authority for a.eks2.<redacted>.tk: dns: domain must be fully qualified

Cant see what is wrong - the SOA record is there at the base. 
Will dig deeper.

ldez:
sorry I forget to create the FQDN:
```
zoneNameOrID, err := dns01.FindZoneByFqdn(dns01.ToFqdn(domain))
```

chrisbulgaria:
thanks - this is of course much better, and works like a charm !

Chris

mattn:
I'm still having this issue.

ldez:
@mattn could you open a new issue?

PR #1058, opened 2020-02-09
dnsproviders: Implement NewConfigFromEnv()

### LINKED ISSUE #1054: DNS providers: Don't return half-populated configs from environment
I'm trying to make DNS providers with a custom config, but pre-populating default values from environment variables.

For example, take the Cloudflare provider: https://pkg.go.dev/github.com/go-acme/lego/v3@v3.3.0/providers/dns/cloudflare?tab=doc - the env variable logic is split into two places: `NewDefaultConfig()` and `NewDNSProvider()`. But these functions can't be used together!

Only `NewDNSProvider()` fills in default credentials from the environment, but it does not expose the config for further customization. So the config is only half-filled out, and we cannot use `NewDefaultConfig()` and `NewDNSProvider()` together.

If a default config was pre-populated from the environment and then returned from `NewDefaultConfig()`, this would make it possible for us to avoid putting DNS provider credentials in app config files. It'd be nice to call just one function and get a config that is filled out all the way from the environment. Otherwise, we'll have to copy+paste and repeat the env variable logic _for every DNS provider that does this_ in our own applications. 😢 

So, could we move all the environment variable logic into `NewDefaultConfig()` rather than having it split in two places like it is now? 😁

mholt:
This allows getting a config populated from env variables.
Without this, the only way to do this is to copy/duplicate all the env var logic.

`NewConfigFromEnv()` treats env variables as optional defaults.
Since the config struct is returned, its values can still be set/changed/read by the caller, which is immensely useful during application setup/config.

In practice, this allows us to keep credentials out of config files, while still being able to configure other elements of the DNS provider in code without relying on global env vars (which we can't do in Caddy 2, since we don't really have much/any global state -- so using a global value would lead to incorrect results).

Fixes #1054

I am just starting out with a few providers of my choice, but I am sure there will be more to follow!

### DIFF
--- a/providers/dns/cloudflare/cloudflare.go
+++ b/providers/dns/cloudflare/cloudflare.go
@@ -32,6 +32,28 @@ type Config struct {
 	HTTPClient         *http.Client
 }
 
+// NewConfigFromEnv returns a new config that is populated from environment variables.
+func NewConfigFromEnv() *Config {
+	values, err := env.GetWithFallback(
+		[]string{"CLOUDFLARE_EMAIL", "CF_API_EMAIL"},
+		[]string{"CLOUDFLARE_API_KEY", "CF_API_KEY"},
+	)
+	if err != nil {
+		values, _ = env.GetWithFallback(
+			[]string{"CLOUDFLARE_DNS_API_TOKEN", "CF_DNS_API_TOKEN"},
+			[]string{"CLOUDFLARE_ZONE_API_TOKEN", "CF_ZONE_API_TOKEN", "CLOUDFLARE_DNS_API_TOKEN", "CF_DNS_API_TOKEN"},
+		)
+	}
+
+	config := NewDefaultConfig()
+	config.AuthEmail = values["CLOUDFLARE_EMAIL"]
+	config.AuthKey = values["CLOUDFLARE_API_KEY"]
+	config.AuthToken = values["CLOUDFLARE_DNS_API_TOKEN"]
+	config.ZoneToken = values["CLOUDFLARE_ZONE_API_TOKEN"]
+
+	return config
+}
+
 // NewDefaultConfig returns a default configuration for the DNSProvider
 func NewDefaultConfig() *Config {
 	return &Config{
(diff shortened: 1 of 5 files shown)


ldez:
You have to allow maintainers to modify your PR.

mholt:
What do you want to change?

ldez:
The behavior is not good: `NewDefaultConfig()` must return error.

mholt:
Hmm, why is that? `NewDefaultConfig()` does not return an error.

I do not think this function should return an error.

Env vars aren't _required_ since the config is being returned so the caller can further populate its fields. If we want to verify a config and return errors, that should be done in `NewDNSProviderConfig()`.

ldez:
Do you have run the tests?

This function must return errors.

ldez:
I understand what you want to do but it is not compatible with the behavior that I want.

So we need to find another solution.

mholt:
`go test ./...` passes on my machine. How do I view what failed in the CI? I see thousands of lines of output in the Travis logs and https://github.com/go-acme/lego/pull/1058/checks?check_run_id=434919480 doesn't actually say what the failure is.

> I understand what you want to do but it is not compatible with the behavior that I want.

What behavior do you want, exactly?

ldez:
the tests fail on my computer, it's not related to the CI.

mholt:
Ah, I found it. (I've been having major issues with VS Code + gopls and modules lately for some reason.) One moment, will submit a corrective commit for the test.

ldez:
Don't fix the tests for now.

mholt:
@ldez Too late 😄 These tests assume that env variables are required, which is not the case since the struct can be filled out by the caller.

ldez:
You cannot remove the tests. Also I don't agree with that.

mholt:
> You cannot remove the tests. Also I don't agree with that.

Well, I _did_ remove the tests, because with this refactor they are no longer relevant.

Er, what don't you agree with, exactly?

ldez:
I think we do not know understood in the issue, we do not have the same version the behavior of the method.

Now I see what do you want, and I'm reconsidering your request.

ldez:
I cannot accept your PR, the error from `NewConfigFromEnv()` must be kept.

There are several thing that can produce those errors (not only missing env vars) and message details (in the missing env var case) are important.

For now, I refuse to change the behavior only for another application (even if it's Traefik or Caddy).

As solution, I think we can expose the env vars names to help you.

So for now, I will close this PR and continue to discuss on the issue  #1054.

PR #1065, opened 2020-02-21
Add a LEGO_NAMESERVER environment variable to set DNS servers during development

remilapeyre:
I'm using this to test https://github.com/remilapeyre/vault-acme.

### DIFF
--- a/challenge/dns01/nameserver.go
+++ b/challenge/dns01/nameserver.go
@@ -3,6 +3,7 @@ package dns01
 import (
 	"fmt"
 	"net"
+	"os"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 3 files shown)


ldez:
Hello,

Instead of introduce an env var into lego you can use the `AddRecursiveNameservers` function and put the env var in your code.

```go
provider, err := dns.NewDNSChallengeProviderByName(a.Provider)
if err != nil {
	return err
}
err = client.Challenge.SetDNS01Provider(provider, dns01.AddRecursiveNameservers(os.Getenv("MY_ENV_VAR")))
if err != nil {
	return err
}
```

https://github.com/remilapeyre/vault-acme/blob/93fee5e9a7a907fa1e386c3402b1a5558f1195ed/acme/path_certs.go#L182-L191

Thank you anyway.

remilapeyre:
Thanks, this was dumb.

PR #1070, opened 2020-02-26
Wrap errors.

ldez:
Use the go1.13 errors wrapping.

before:
```go
fmt.Errorf("test: %v", err)
```

after:
```go
fmt.Errorf("test: %w", err)
```

### DIFF
--- a/acme/api/account.go
+++ b/acme/api/account.go
@@ -31,12 +31,12 @@ func (a *AccountService) New(req acme.Account) (acme.ExtendedAccount, error) {
 func (a *AccountService) NewEAB(accMsg acme.Account, kid string, hmacEncoded string) (acme.ExtendedAccount, error) {
 	hmac, err := base64.RawURLEncoding.DecodeString(hmacEncoded)
 	if err != nil {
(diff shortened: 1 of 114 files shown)


mholt:
Wow, this is cool.

It's a really big change, though, and I don't have time to review each line right now; it looks pretty comprehensive! I do have a couple high-level questions by way of review:

- What is the motivation for this change?
- Do we intend to expose all the underlying error values to the (exported) API? Is that what this is doing, effectively? If so, are we bound by error values for compatibility?

I'm generally in favor of less opaque errors.

wyattjoh:
What's really neat is that the `%w` will implement errors with a `.Unwrap` (see https://golang.org/pkg/fmt/#Errorf, also more details here https://golang.org/pkg/errors/). The idea would be that we can find deeper errors that occurred.

ldez:
> What is the motivation for this change?

Just because I see that as a good practice instead of the simple `%v`

> Do we intend to expose all the underlying error values to the (exported) API? 

We will not expose all internal errors structures.
@wyattjoh give a good explanation of the behavior and the advantage.

PR #1072, opened 2020-02-29
feat: multi-arch Docker image.

### LINKED ISSUE #1071: Multi-arch Docker Images
Hello,
I used several Raspberry Pi's (armv71) and would like to use the lego Docker image.
Currently only an amd64 build is uploaded to Docker Hub.
It seems that it is possible to create builds for other architectures on an amd64 system (multi-arch docker images).
Is it possible to implement this feature?

ldez:
Allows to publish (on tags only) a multi-arch image.

I use a tool that I created to publish multi-arch image easily: https://github.com/ldez/seihon

I already added env vars to Travis CI: https://travis-ci.org/go-acme/lego/settings

Fixes #1071

https://hub.docker.com/r/goacme/lego

### DIFF
--- a/tmpl.Dockerfile
+++ b/tmpl.Dockerfile
@@ -1,3 +1,5 @@
+# Dockerfile template used by Seihon to create multi-arch images.
+# https://github.com/ldez/seihon
 FROM golang:1-alpine as builder
 
 RUN apk --update upgrade \
--- a/tmpl.Dockerfile
+++ b/tmpl.Dockerfile
@@ -1,0 +1,24 @@
+FROM golang:1-alpine as builder
+
+RUN apk --update upgrade \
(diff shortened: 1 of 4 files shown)


tmpl.Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM golang:1-alpine as builder
```
dmke: Can we add a comment to clarify what this file is for?

As far as I can tell, `seihon` reads it as template, when no CLI flag is given (`--template=./tmpl.Dockerfile`). This should be made somewhat explicit - either with a comment here, or an explicit call to `seihon` in the `.travis.yml`

ldez: I added a comment in the `tmpl.Dockerfile`.

PR #1073, opened 2020-03-04
Upgrade `cloud.google.com/go` to 0.53

lawliet89:
Includes https://github.com/googleapis/google-cloud-go/commit/fbf2f5121ef6008eb9b7e5c67aa0c9abfa7ac7fc
which fixes issues when used with GKE Workload Identity

Also ran `go mod tidy`.


ldez:
Hello,

could you give more precise information about issues related to that PR?

lawliet89:
That PR removes timeout reading from the GCE metadata server. Specifically, the GKE workload identity server takes longer than the GCE metadata server to return responses. Otherwise, this can lead to issues like the one [here](https://github.com/influxdata/telegraf/issues/7043). In my case, I have also seen similar errors when Traefik tries to CloudDNS records.

GCP support recommends upgrading to >= 0.52.

PR #1110, opened 2020-04-10
azure: Allow for the use of MSI

JulienBalestra:
Azure provide a way to associate an identity to an instance via [user-managed-identity]( https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/overview).

This PR allows lego to use that identity by updating the azure api to get its credential.

This should be retro-compatible to the previous way to get credentials.

### DIFF
--- a/providers/dns/azure/azure.go
+++ b/providers/dns/azure/azure.go
@@ -13,8 +13,6 @@ import (
 
 	"github.com/Azure/azure-sdk-for-go/services/dns/mgmt/2017-09-01/dns"
 	"github.com/Azure/go-autorest/autorest"
-	"github.com/Azure/go-autorest/autorest/adal"
-	"github.com/Azure/go-autorest/autorest/azure"
 	"github.com/Azure/go-autorest/autorest/azure/auth"
 	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/go-acme/lego/v3/challenge/dns01"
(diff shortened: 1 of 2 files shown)


providers/dns/azure/azure.go
```diff
@@ -240,20 +238,8 @@ func toRelativeRecord(domain, zone string) string {
...
-		}
-
-		spt, err := adal.NewServicePrincipalToken(*oauthConfig, config.ClientID, config.ClientSecret, azure.PublicCloud.ResourceManagerEndpoint)
-		if err != nil {
-			return nil, err
-		}
-
-		spt.SetSender(config.HTTPClient)
-		return autorest.NewBearerAuthorizer(spt), nil
+		return auth.NewClientCredentialsConfig(config.ClientID, config.ClientSecret, config.TenantID).Authorizer()
```
ldez, line 241:
```suggestion
		credentialsConfig := auth.NewClientCredentialsConfig(config.ClientID, config.ClientSecret, config.TenantID)

		spToken, err := credentialsConfig.ServicePrincipalToken()
		if err != nil {
			return nil, fmt.Errorf("failed to get oauth token from client credentials: %v", err)
		}

		spToken.SetSender(config.HTTPClient)

		return autorest.NewBearerAuthorizer(spToken), nil
```

ldez:
ping @JulienBalestra

PR #1142, opened 2020-05-05
Registrar.ResolveAccountByKey: Fix issues found testing with Pebble

chen-anders:
We use [pebble:latest](https://github.com/letsencrypt/pebble) as a way to integration-test our own LetsEncrypt certificate fetcher. 

One notable issue I found was that the `Registrar.ResolveAccountByKey` method errors out when we try to resolve an existing account with the following error: 

```
unable to resolve account by key: acme: error: 400 :: POST :: https://127.0.0.1:14000/my-account/1 :: urn:ietf:params:acme:e
rror:malformed :: Use POST-as-GET to retrieve account data instead of doing an empty update
```

Based on the pebble source [here](https://github.com/letsencrypt/pebble/blob/35691f87bc2df5a34577c4617975e269eec86f87/wfe/wfe.go#L987), our `POST-as-GET` gets interpreted as an update call and is failed as a malformed request. 

It seems like we used to be able to get account information this way, but this seems to have changed as a result of: https://github.com/letsencrypt/pebble/pull/171

### DIFF
--- a/registration/registar_test.go
+++ b/registration/registar_test.go
@@ -18,14 +18,7 @@ func TestRegistrar_ResolveAccountByKey(t *testing.T) {
 	defer tearDown()
 
 	mux.HandleFunc("/account", func(w http.ResponseWriter, _ *http.Request) {
-		w.Header().Set("Location", apiURL+"/account_recovery")
-		_, err := w.Write([]byte("{}"))
-		if err != nil {
-			http.Error(w, err.Error(), http.StatusInternalServerError)
-		}
-	})
-
-	mux.HandleFunc("/account_recovery", func(w http.ResponseWriter, _ *http.Request) {
+		w.Header().Set("Location", apiURL+"/account")
 		err := tester.WriteJSONResponse(w, acme.Account{
 			Status: "valid",
 		})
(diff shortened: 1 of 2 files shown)


registration/registar_test.go
```diff
@@ -18,14 +18,7 @@ func TestRegistrar_ResolveAccountByKey(t *testing.T) {
...
 	mux.HandleFunc("/account", func(w http.ResponseWriter, _ *http.Request) {
-		w.Header().Set("Location", apiURL+"/account_recovery")
-		_, err := w.Write([]byte("{}"))
-		if err != nil {
-			http.Error(w, err.Error(), http.StatusInternalServerError)
-		}
-	})
-
-	mux.HandleFunc("/account_recovery", func(w http.ResponseWriter, _ *http.Request) {
+		w.Header().Set("Location", apiURL+"/account")
```
chen-anders, line 21: Is the `Location` still needed or can it be blank?

chen-anders:
hey @ldez , is there anything else that needs to be done for this change to be reviewed?

chen-anders:
Thanks again! @ldez

Issue #1146, opened 2020-05-09
Add the same env vars to "run" as in "renew"


smtalk:
Currently we have these env vars filled on "renew":
```
LEGO_ACCOUNT_EMAIL: the email of the account.
LEGO_CERT_DOMAIN: the main domain of the certificate.
LEGO_CERT_PATH: the path of the certificate.
LEGO_CERT_KEY_PATH: the path of the certificate key.
```

However. on run time there is no easy way to run something (hook) for specific new cert file, because environment has nothing on it (unless I missed something in documentation?). It can be workarounded with find and mtime checks or things like that, but an official way of doing it would be great (just add the same env vars that we have in "renew" to "run", and this would be totally enough).

ldez:
Hello,

FYI, the env vars are only added in the hook for the renew command, neither by nor for the command it-self.

When you use the `run` command, this information is deterministic because it is the information that you give to the CLI.

smtalk:
If you define the command yourself - yes, you could know this, but not everything, if "run" is executed automatically of different environments - I'd say no.

And even if I define it manually, like:
`lego -m admin@domain.com --http -d *.domain.com -d domain2.comf -d www.domain.com run`

How would you know the end-resulting path and filename from the command name? :) Yes, I know it'd form _.domain.com, because it's how lego calls wildcard domain name files. Additionally - I would need to know that script CWD, to know where .lego directory was placed, then read the LEGO_PATH directive to know if it's not overridden (because from docs it seems `--path` isn't the only way to override it). And can I be always sure lego never changes the structure of .lego folder? (for example, after implementation of https://github.com/go-acme/lego/issues/873, or https://github.com/go-acme/lego/issues/838, because if it does - hook wouldn't find it anymore). 

It'd be more clear with `--filename` option (and that way we could say we know all the details we need), but that option is deprecated.

Maybe I'm just missing something, and as you said "this information is deterministic because it is the information that you give to the CLI" - the end-resulting full path to the cert is known already (without parsing the output of "lego list", of course, which would be very nice to have in JSON format as well)?

Thank you.

ldez:
> And can I be always sure lego never changes the structure of .lego folder?

yes and no: we follow semver, so if we create a breaking change we will also change the major version.

I created a PR to a hook on the run command: #1157

smtalk:
[...]

PR #1157, opened 2020-05-12
cli: add hook on the run command.

ldez:
Related to #1146, #1108, #845


dmke:
LGTM, on a first glance.

I'm currently a bit time constrained, so I'll add a proper review in the next few days.

PR #1158, opened 2020-05-12
Ensure we return a location during account updates

jaredledvina:
## Context
Currently, when `UpdateRegistration` https://github.com/go-acme/lego/blob/b1dba4f23d357e1b96ff4a3e8295412b55d5c330/registration/registar.go#L119-L140 is called, a resource leveraging the `Account.Location` is returned to the caller. 

However, the existing logic in `Update` https://github.com/go-acme/lego/blob/b1dba4f23d357e1b96ff4a3e8295412b55d5c330/acme/api/account.go#L60-L72 leverages the `ExtendedAccount` object which via https://github.com/go-acme/lego/blob/fa5555d3f4614ee039686bd15a5f1e806b3e3c40/acme/commons.go#L60-L65 drops the `Location` field resulting in the `UpdateRegistration` call losing the URI. 

## Proposed Change
This PR aligns the account `Update` call closer to that of `New` call https://github.com/go-acme/lego/blob/b1dba4f23d357e1b96ff4a3e8295412b55d5c330/acme/api/account.go#L14-L28 such that we ensure that we return the `Location` allowing `UpdateRegistration` to have a filled in `URI` field.


ldez:
Hello,

Could you give more details about this change?

Could you also update the test `TestRegistrar_UpdateAccount`.

thank you.

jaredledvina:
Hey @ldez,

I'm terribly sorry about that, it appears that my previous description got lost along the way. I just re-filled that in and hopefully it makes some sense. Let me know if there's any questions about this and I'll take a shot at updating the `TestRegistrar_UpdateAccount` test as well. Thanks so much for the really fast reply!

jaredledvina:
It's unclear to me why the tests isn't working here, seems like the newly registered account URI is empty which is unexpected for me. 


That aside, I went ahead and review https://tools.ietf.org/html/rfc8555#section-7.3.2 that states

> If the server accepts the update, it MUST return a response with a 200 (OK) status code and the resulting account object.

And then looking at https://tools.ietf.org/html/rfc8555#section-7.1.2, it's clear that the account location/URI is not actually part of the account object. 

I guess, that would mean that instead of `Update` returning the account URI/Location, one should follow that up w/ looking up the account URL w/ the key per https://tools.ietf.org/html/rfc8555#section-7.3.1. I'm curious anyone knows if that's the expected usage here? 

If so, we can close this and I'll refactor my code to not expect update to return a URI/Location on my end.

PR #1163, opened 2020-05-17
Add an example how to generate rsa keys

### LINKED ISSUE #1102: EC keys not accepted everywhere, --key-type option hard to find in docs
This one will be easy to chalk up as PEBKAC, but here's what I ran into this evening:

1) First time user of lego; generate wildcard cert using cloudflare dns handler. Works great.
2) Upload cert and key to Linode Nodebalancer config. Private key is rejected as "not being formatted correctly"
3) Try various other forms of the key, to no avail.
4) Generate a cert using certbot, works perfectly first try.
5) Friend discovers --key-type by reading issues and the source code.
5) I look through docs on https://go-acme.github.io/lego/usage/cli/. See no mention of --key-type. Then we discover that the tiny "CLI help" text under usage that I had initially read right past is actually a BUTTON that shows the entire helptext.

So I guess in summary, if lego is going to default to EC384 which doesn't seem to be (yet) supported everywhere, perhaps that could be made more obvious (to me at least haha) in the examples? A single example saying "Generate RSA Keys for services that don't support EC384" might have clued me in.

Either way, thanks for lego! Working quite well now after working that out.

herrphon:
Fix #1102

Add an example for rsa keys.

### DIFF
--- a/docs/content/usage/cli/examples.md
+++ b/docs/content/usage/cli/examples.md
@@ -77,7 +77,7 @@ lego --email="foo@bar.com" --http --csr=/path/to/csr.pem run
 
 (lego will infer the domains to be validated based on the contents of the CSR, so make sure the CSR's Common Name and optional SubjectAltNames are set correctly.)
 
-### Obtail a certificate of type rsa instead of the default ec keys
+### Obtain a certificate of type rsa instead of the default ec keys
 
 Elliptic curve keys are not supported by old systems, so you want to fall back on rsa:
 
--- a/docs/content/usage/cli/examples.md
+++ b/docs/content/usage/cli/examples.md
@@ -77,6 +77,15 @@ lego --email="foo@bar.com" --http --csr=/path/to/csr.pem run
 
 (lego will infer the domains to be validated based on the contents of the CSR, so make sure the CSR's Common Name and optional SubjectAltNames are set correctly.)
 
+### Obtail a certificate of type rsa instead of the default ec keys
+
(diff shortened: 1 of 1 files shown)


docs/content/usage/cli/examples.md
```diff
@@ -77,6 +77,15 @@ lego --email="foo@bar.com" --http --csr=/path/to/csr.pem run
 
 (lego will infer the domains to be validated based on the contents of the CSR, so make sure the CSR's Common Name and optional SubjectAltNames are set correctly.)
 
+### Obtain a certificate of type rsa instead of the default ec keys
+
+Elliptic curve keys are not supported by old systems, so you want to fall back on rsa:
```
dmke, line 82: RSA and EC are acronyms, please change to uppercase.

I would also like to specify what "old systems" are, or at least link to more information. GlobalSign has a neat overview [here](https://support.globalsign.com/ssl/ssl-certificates-life-cycle/ecc-compatibility). (The footnotes didn't survive being moved between CMS, from the looks of it though. It would be nicer, if something similar would exist on letsencrypt.org.)

From what I can gather, you'll need an RSA key, if you absolutely must support hard- or software which has reached or is near EOL. The only non-EOL entry on the list at GlobalSign is CentOS/RHEL, with the minimum version required being 6.5 (released in 2014).

dmke (changes requested):

ldez:
Closed due to lack of feedback and corrections

PR #1176, opened 2020-05-30
inwx: Two-Factor-Authentication

### LINKED ISSUE #1168: Support Two-Factor-Authentication for INWX provider
I'd love to use lego (for my traefik container) for my inwx domains but for now I have to disable TFA since lego does not seem to support it. Is it possible to additionally support this?
The used library seems to support it, even though I could not get it running (see nrdcg/goinwx#1) --- but I assume this is something someone with more go knowledge can do easily?!

From a corresponding python inwx provider it seems to be rather simple: https://github.com/oGGy990/certbot-dns-inwx/blob/master/certbot_dns_inwx/dns_inwx.py#L121-L128
And there also seem to be TOTP implementations for go readily available: https://github.com/pquerna/otp/blob/master/totp/totp.go

ldez:
Closes #1168


providers/dns/inwx/inwx.go
```diff
@@ -152,6 +173,23 @@ func (d *DNSProvider) CleanUp(domain, token, keyAuth string) error {
 		}
 	}()
 
+	if info.TFA == "GOOGLE-AUTH" {
+		if d.config.SharedSecret == "" {
+			return errors.New("two factor authentication but no shared secret is given")
+		}
+
+		var tan string
+		tan, err = totp.GenerateCode("SHARED_SECRET", time.Now())
```
dmke, line 182: This should be `totp.GenerateCode(d.config.SharedSecret, time.Now())`, shouldn't it?

dmke: Also, the 2fa code block here is identical to the one in `Present()`, we may want to move it into a separate function.

XenGi:
I still don't get how I would use that shared secret. I tried with the one that is used for TOPT ut that didn't work. What secret should I use and where do I get it?

Issue #1008, opened 2019-11-12
Issues creating certificates for subdomain with route53


armsby:
I have been trying to create a certificate using letsencrypt and route53 the certificate im trying to create is for 'server.sub.domain.com' when trying to use route53 it I get an error saying that it can not find the host zone id for sub.domain.com, I belive that is a bug as the domain it should be looking for is domain.com, and that does exist, there is no issues creating certificates for that domain.

I have also tested it with cloudflare for another domain and that works perfectly, so I belive that the problem is when the api call towards route53

MichaelMure:
It's happening to me as well with `v3.7.0`.

MichaelMure:
After some debugging, it looks to me that what is happening is:
- code flow somehow end up in `route53.DNSProvider.Present()` at the start of the challenge
- call to `d.getHostedZoneID(fqdn)` to figure out what the name of the hosted zone is
- call graph continue to `dns01.fetchSoaByFqdn()` to perform recursive DNS call to see if there is a SOA record. For example for `foo.bar.example.org`, it will query `foo.bar.example.org` then `bar.example.org` then `example.org`, which should have this SOA record.
- the problem happen when in `dns01.fetchSoaByFqdn()` a DNS query has a temporary failure (say, a timeout). **This error is not handled there, it just skip the node in the domain**.
- if this failure happen at the domain that should have the SOA record (`example.org`), the function will end up returning `org` instead of `example.org
- later, the AWS SDK call to find the Route53 hosted zone by name (`ListHostedZonesByName`) will be called with `org` instead of `example.org` and fail

MichaelMure:
So to me this is not a Route53 provider failure, this is a `dns01` one.

ldez:
If you have an intermittent timeout, I think you should check your network and its configuration.

ldez:
You can also simply configure the DNS timeout.
```
--dns-timeout
```

https://go-acme.github.io/lego/usage/cli/#usage

armsby:
it does not look to be a timeout issue, it is only when the host name is on a subdomain, I am updating certificates on both sides of the failures for the main domain and it is only the lego client that fails, I have no issues when using https://github.com/acmesh-official/acme.sh

MichaelMure:
> If you have an intermittent timeout, I think you should check your network and its configuration, and the nameservers that you are using.

[...]

PR #1178, opened 2020-06-02
fetchSoaByFqdn: retry DNS query

MichaelMure:
This is a very crude solution (I don't fully understand the failure modes of `dnsQuery`) to demonstrate https://github.com/go-acme/lego/issues/1008.

I suspect you will want something more robust, maybe react differently on different class of error, but it does solve (or greatly mitigate) the problem for me.


ldez:
Hello,

Thank you, but your implementation is too naive to be merged.

I will close your PR but I will continue the discussion in #1008.

PR #1179, opened 2020-06-02
binaries: Add MIPS arch.

### LINKED ISSUE #1174: Support MIPS & MIPS64 architecture
Would like to see builds (releases) for these architectures, if that's possible.

ldez:
Fixes #1174

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -24,7 +24,9 @@ builds:
       - arm
       - arm64
       - mips
+      - mipsle
       - mips64
+      - mips64le
     goarm:
       - 7
       - 6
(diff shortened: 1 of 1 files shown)


.goreleaser.yml
```diff
@@ -23,10 +23,15 @@ builds:
       - 386
       - arm
       - arm64
+      - mips
+      - mips64
```
dmke, line 27: There are also the little-endian architectures `mipsle` and `mips64le`.

dmke: Do we want to add a UPX compression step for MIPS binaries (see [example](https://github.com/goreleaser/goreleaser/issues/146#issuecomment-288793339))?

Typically, those chips run in embedded environment where available space is always critically low (and Go binaries are way to large).

(I'm also OK with leaving this as an exercise for the user).

ldez: UPX have some side effect, I prefer to let user play with that.

PR #1180, opened 2020-06-03
feat: retry on DNS queries.

### LINKED ISSUE #1008: Issues creating certificates for subdomain with route53

ldez:
Fixes #1008


MichaelMure:
Tested today, it does improve the situation. However, I would advice to also fail the process early [here](https://github.com/go-acme/lego/blob/master/challenge/dns01/nameserver.go#L182) if a DNS error still happen despite the retry.

ldez:
> However, I would advice to also fail the process early here if a DNS error still happen despite the retry.

This not needed because the DNSError don't produce a retry and by example a timeout can have several reasons.

challenge/dns01/nameserver.go
```diff
@@ -229,15 +231,39 @@ func dnsQuery(fqdn string, rtype uint16, nameservers []string, recursive bool) (
...
+	var errG error
 
 	for _, ns := range nameservers {
-		in, err = sendDNSQuery(m, ns)
-		if err == nil && len(in.Answer) > 0 {
+		bo := backoff.NewExponentialBackOff()
+		bo.Multiplier = 1.2
+		bo.InitialInterval = dnsTimeout
+		bo.MaxInterval = 2 * bo.InitialInterval
+		bo.MaxElapsedTime = 4 * bo.InitialInterval
```
dmke, line 241: This sets the max. wait time to 2 minutes if I invoke lego with `--dns-timeout=30`, doesn't it? Should `bo.InitialInterval`  be `dnsTimeout/4` instead?

challenge/dns01/nameserver.go
```diff
@@ -229,15 +231,39 @@ func dnsQuery(fqdn string, rtype uint16, nameservers []string, recursive bool) (
 	m := createDNSMsg(fqdn, rtype, recursive)
 
 	var in *dns.Msg
-	var err error
+	var errG error
```
dmke, line 234: What does `errG` stand for? Neither "backoff", nor "retry", nor "query" contain a "g".

challenge/dns01/nameserver.go
ldez: the initial interval must be the same as the timeout use by the DNS client.

challenge/dns01/nameserver.go
ldez: https://github.com/go-acme/lego/pull/1180/files#diff-2064c998dfa170d0767a3036f371c862R262

dmke: That doesn't explain the "G" in the variable name. I can see where it is used/assigned, but I wonder why it's named `errG` and not something like `errQuery` or `errBackoff`...

challenge/dns01/nameserver.go
dmke: > The initial interval must be the same as the timeout use by the DNS client

I see. Then the documentation of the `--dns-timeout` flag needs to be modified to something like "this is the minimum timeout given to resolve SOA queries; on network errors the actual timeout might be considerably longer".

Alternatively, we can reduce the default timeout down to something like 2 or 3s to approximate the previous behaviour (on that note, this should work in *most* cases, because DNS *usually* resolves fast enough). To accommodate #1008, we could also bump `bo.MaxInterval` and `bo.MaxElapsedTime` to `10*initialInterval` and `20*initialInterval` respectively.

challenge/dns01/nameserver.go
ldez: `G` because `Global`, but you are right I can improve this name.

challenge/dns01/nameserver.go
ldez: I can reduce the default timeout.

But I thank that multiply by 10 or 20 is too huge: the dnsQuery is used by several functions in a loop (by example a loop on ns) and this can become very very long.

challenge/dns01/nameserver.go
dmke: Ah, of course. That I did not think of. Thanks for clarification/correction :+1:

challenge/dns01/nameserver.go
dmke: Oh, I've missed the context; I thought this was part of the `fetchSoaByFqdn` methods. Sorry for the confusion on my part.

ldez:
I am wondering about the impact and side effects of this PR. I need to take a little more time to think.

PR #1200, opened 2020-06-27
constellix: fix search records API call.

ldez:
https://api-docs.constellix.com/?version=latest#81003e4f-bd3f-413f-a18d-6d9d18f10201

### DIFF
--- a/providers/dns/constellix/constellix.go
+++ b/providers/dns/constellix/constellix.go
@@ -122,40 +122,12 @@ func (d *DNSProvider) Present(domain, token, keyAuth string) error {
 
 	// TXT record entry already existing
 	if len(records) == 1 {
-		record, err := d.client.TxtRecords.Get(dom.ID, records[0].ID)
-		if err != nil {
-			return fmt.Errorf("constellix: failed to get TXT records: %w", err)
-		}
-
-		if containsValue(record, value) {
-			return nil
-		}
-
-		request := internal.RecordRequest{
-			Name:       record.Name,
-			TTL:        record.TTL,
-			RoundRobin: append(record.RoundRobin, internal.RecordValue{Value: fmt.Sprintf(`"%s"`, value)}),
-		}
-
-		_, err = d.client.TxtRecords.Update(dom.ID, record.ID, request)
-		if err != nil {
-			return fmt.Errorf("constellix: failed to update TXT records: %w", err)
-		}
-
-		return nil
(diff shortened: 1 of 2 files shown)


providers/dns/constellix/constellix.go
```diff
@@ -122,7 +122,10 @@ func (d *DNSProvider) Present(domain, token, keyAuth string) error {
```
dmke, line 122: These ~20 lines at the start are almost identical to the first ~20 lines of `CleanUp()`. Maybe we can DRY'up and move it into a helper method? Something like

```go
func (d *DNSProvider) preflightcheck(domain, keyAuth string) (authZone string, dom internal.Domain, records []internal.Records, err error) {

	authZone, err = dns01.FindZoneByFqdn(fqdn)
	if err != nil {
		err = fmt.Errorf("constellix: could not find zone for domain %q and fqdn %q : %w", domain, fqdn, err)
		return
	}

	dom, err = d.client.Domains.GetByName(dns01.UnFqdn(authZone))
	if err != nil {
		err = fmt.Errorf("constellix: failed to get domain: %w", err)
		return
	}

	recordName := getRecordName(fqdn, authZone)

	records, err = d.client.TxtRecords.Search(dom.ID, internal.Exact, recordName)
	if err != nil {
		err = fmt.Errorf("constellix: failed to search TXT records: %w", err)
		return
	}

	if len(records) > 1 {
		err = errors.New("constellix: failed to get TXT records")
		return err
	}
	return
}

func (d *DNSProvider) Present(domain, token, keyAuth string) error {
	authZone, dom, records, err := preflightcheck(domain, keyAuth string)
	if err != nil {
		return err
	}
	// ...
}
```

I'm usually not a fan of methods returning more than three values, but adding a `type preflight struct { authZone string, dom internal.Domain, records []internal.Records }` wouldn't clear things up by that much, wouldn't it?

ldez: I'm usually not a fan of technical factorization: I prefer "business/behavioral" factorization.
I'm also not a use fan of returning more than three values, named returns.

`preflight`method and struct are just a technical factorization, it can a be seen as a simplification from a maintenance point of view but also as a limitation of the capacities of evolution.

So I would prefer to extract methods from the "second" parts of the functions, because those sections are more "business" related (create/delete TXT records)

dmke: Sounds good to me :+1:

Issue #1217, opened 2020-07-16
Support Openstack ckouds.yaml


johscheuer:
Openstack supports to specify credentials via [clouds.yml](https://docs.openstack.org/python-openstackclient/pike/configuration/index.html) which is an additional way to handle credentials for Openstack. I would propose to add the following logic:

- If the environment variable `OS_CLOUD` is set try to fetch the credentials from the file
- If the environment variable `OS_CLOUD` is *not* set try to fetch the credentials from the `OS_` variables.

johscheuer:
I would start working on this (since we need it for our use case).

PR #1219, opened 2020-07-16
designate: add support for Openstack clouds.yaml

johscheuer:
This PR adds support for `clouds.yml` with the Openstack provider.

Fixes: https://github.com/go-acme/lego/issues/1217

### DIFF
--- a/providers/dns/designate/designate.go
+++ b/providers/dns/designate/designate.go
@@ -15,6 +15,7 @@ import (
 	"github.com/gophercloud/gophercloud/openstack"
 	"github.com/gophercloud/gophercloud/openstack/dns/v2/recordsets"
 	"github.com/gophercloud/gophercloud/openstack/dns/v2/zones"
+	"github.com/gophercloud/utils/openstack/clientconfig"
 )
 
 // Environment variables names.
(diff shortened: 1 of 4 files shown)


providers/dns/designate/designate.go
```diff
@@ -62,15 +64,32 @@ type DNSProvider struct {
...
+		})
+
+		if err != nil {
+			return nil, fmt.Errorf("designate: %w", err)
+		}
+
+		opts = *cloudOpts
+	} else {
+		// TODO this is broken -> variables in _FILE will be ignored!
+		_, err = env.Get(EnvAuthURL, EnvUsername, EnvPassword, EnvTenantName, EnvRegionName)
```
johscheuer, line 84: It seems like this never worked as intended since we don't use the fetched values and we don't set the variables found in `_FILE` as env. variable. I would propose to remove the feature and instead refer to the `clouds.yaml`. This has multiple benefits: we can drop the hardcoded variables in https://github.com/go-acme/lego/pull/1219/files#diff-91861d95ae59b1d7adb03d128b1c20efL35 and just use the `openstack.AuthOptionsFromEnv()` function to fetch the credentials (this function already does some sanity checks) and we would also get application credentials "for free"

johscheuer:
I fix the CI issues (just needed to adjust my local golandci-lint).

johscheuer:
Friendly ping @ldez :)

PR #1227, opened 2020-08-09
add --preferred-chain option to support "alternate" certificate links

### LINKED ISSUE #1197: Support "alternate" certificate links for selecting different signing Chains
See: 
https://community.letsencrypt.org/t/transition-to-isrgs-root-delayed-until-sep-29/125516
and
https://tools.ietf.org/html/rfc8555#section-7.4.2

The cert download post will supply additional "alternate" links in the future (might be already present. i'm not sure).
I think the CertificateService and maybe Certificate resource would need to be extended to support the additional certs/CA Chains at least for usage as a library (my use-case). Not sure about CLI though.

Songmu:
resolve #1197

I implemented it referenced from the following pull request of certbot.

https://github.com/certbot/certbot/pull/8080

This change is needed asap, because Let's Encrypt is to change its default certificate chain on September 29.

ref.
- https://letsencrypt.org/2019/04/15/transitioning-to-isrg-root.html
- https://github.com/certbot/certbot/issues/6971


certificate/certificates.go
```diff
@@ -287,24 +288,51 @@ func (c *Certifier) getForCSR(domains []string, order acme.ExtendedOrder, bundle
...
+				CertStableURL:     link,
+			}
+		}
+		if preferredChain != "" {
+			x509Certs, err := certcrypto.ParsePEMBundle(issuer)
+			if err != nil {
+				return false, err
+			}
+			if x509Certs[0].Subject.CommonName != preferredChain {
+				continue
```
alexzorin, line 320: As written, a user would not be able to do this:

    --preferred-chain="DST Root CA X3"

because the root certificate is not included in the chain, and as such, no `.Subject` would ever match.

This could be addressed by instead matching against the `.Issuer` field of *every* certificate in the bundle. This is how Certbot does it, though other clients have opted for even more ways to match (like SKI/AKI/hash).

Interactive example demonstrating this use-case against this PR:

    $ go run cmd/lego/main.go --server https://acme-staging-v02.api.letsencrypt.org/directory -d 976fb1d7fc17.ngrok.io --http --http.port=":8123" -m red@act.ed run --preferred-chain "Fake LE Root X2"
    2020/08/18 12:38:40 [INFO] [976fb1d7fc17.ngrok.io] acme: Obtaining bundled SAN certificate
    2020/08/18 12:38:41 [INFO] [976fb1d7fc17.ngrok.io] AuthURL: https://acme-staging-v02.api.letsencrypt.org/acme/authz-v3/95951549
    2020/08/18 12:38:41 [INFO] [976fb1d7fc17.ngrok.io] acme: authorization already valid; skipping challenge
    2020/08/18 12:38:41 [INFO] [976fb1d7fc17.ngrok.io] acme: Validations succeeded; requesting certificates
    2020/08/18 12:38:42 [INFO] lego has been configured to prefer certificate chains with issuer "Fake LE Root X2", but no chain from the CA matched this issuer. Using the default certificate chain instead.

ldez (approved):
Thanks :+1:

Songmu:
Sorry for sending an incomplete pull request.  I hadn't gotten around to fixing it, but thank you for completing it on your end. @ldez

FStelzer:
Thanks a lot for implementing this!

PR #1230, opened 2020-08-09
Add additional namecheap-specific infos

### LINKED ISSUE #1229: Add a warning for DNS providers with strong limitations on API
I know this is an odd request and checking out the definition files, there is not really a good space to put this information.

But I recently moved over some domains to a different DNS provider. One of my reasons to pick a provider was it's support with LEGO. I picked namecheap.com

I did not do my research properly. I just read 'API access supported' - so I moved my stuff over. My treafik dns-01 lets encrypt setup was not ready, but once I got to it, I found the API option and was surprised to see: You need at least 50$ balance and at least 20 registered domains before you can use the API.

That kind of sucks.

What is your opinion on adding such note like:

"Currently, API access is limited to users with at least 50$ balance and 20 resisted domains (2020)." somewhere on the documentation page?

dragetd:
Currently, only after trying to enable API access, you get to see the following requirements:
"Account balance of $50+, 20+ domains in your account, or purchases totaling $50+ within the last 2 years."

Closes #1229

### DIFF
--- a/docs/content/dns/zz_gen_edgedns.md
+++ b/docs/content/dns/zz_gen_edgedns.md
@@ -10,6 +10,7 @@ slug: edgedns
 <!-- THIS DOCUMENTATION IS AUTO-GENERATED. PLEASE DO NOT EDIT. -->
 
 Since: v3.9.0
+
 Akamai edgedns supercedes fastdns; implementing a DNS provider for solving the DNS-01 challenge using Akamai EdgeDNS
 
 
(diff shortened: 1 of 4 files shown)


dragetd:
E.g. this is the message as a screenshot: 
grafik


Since this kind of information might change over time - if we are going to include it, I thought it would be fair to add a date reference.

I also agree that this is not really the point of lego to document this. Would have helped me, but up to you if you want to inculde it!

PR #1232, opened 2020-08-18
Add optional programmatic configuration of the DNS provider

remilapeyre:
In https://github.com/remilapeyre/vault-acme/issues/19 we would like to be able to use instantiate the same provider multiple times with a different configuration. We could use `os.Setenv()` to change the environment variables before doing this but it is cumbersome and would require to have a Mutex to avoid race conditions.

This patch add a new `config` parameters where the values for the configuration is first looked into before looking in the environment variables. Callers that only want to use the environment variables can set this parameter to `nil` and the old behaviour of looking in the environment variables will be kept.

### DIFF
--- a/challenge/dns01/dns_challenge_manual.go
+++ b/challenge/dns01/dns_challenge_manual.go
@@ -15,7 +15,7 @@ const (
 type DNSProviderManual struct{}
 
 // NewDNSProviderManual returns a DNSProviderManual instance.
-func NewDNSProviderManual() (*DNSProviderManual, error) {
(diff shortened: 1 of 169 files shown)


ldez:
The topic as #605

The env vars are not really configurable in the way that your thinking : some providers are more complex to handle configuration than env vars inside lego.

Your approach can works for some providers but it's not a valid global approach.

ldez:
I would propose you to open an issue to explain your use case and be able to discuss around this topic.

As I explain in my previous message, this PR, even it the approach is interesting, can not be merge because it's not a valid global approach.

Thanks anyway.

PR #1234, opened 2020-08-20
Adjust default TTL to be twice the propagation time

ad-m:
I have the impression that the propagation time should be longer than TTL. Because the DNS cache stops propagation of changes for the period defined by TTL.


ad-m:
Tests and provider require revisit, but let's discuss first.

ldez:
Hello,

The TTL depends of each DNS provider, this default value it just a "quick ref".

Changing this value is a breaking change.

So I encourage you to open an issue on the subject so that twe can discuss it, and I will have to close this PR.

Thanks anyway.

PR #1237, opened 2020-08-26
desec: applies API rate limits.

### LINKED ISSUE #1236: respect deSEC.io's API rate limit (sleep for 0.2 seconds after each request)
When requesting a certificate with more than 6 SANs using the [deSEC.io](https://go-acme.github.io/lego/dns/desec/) DNS provider, lego will run into [deSEC.io's API rate limit](https://talk.desec.io/t/request-was-throttled/120/3):

```
 error: 429: detail: Request was throttled.
```

To avoid this there are two options:
* use deSEC's [bulk API request](https://desec.readthedocs.io/en/latest/dns/rrsets.html#bulk-operations) (update N records with a single request): 
* sleep for 0.2 seconds after each request

ldez:
Replace the internal implementation by the Go client.
Applies API rate limits.

fixes #1236


dmke (approved):
LGTM

PR #1244, opened 2020-09-03
exoscale: change dependency version.

ldez:
The versions higher than v0.23.0 use a weird dependency do github.com/golangci/lint-1

https://github.com/exoscale/egoscale/blob/0590bfb2400cfce7358202c5f6337fa02198b1e0/go.mod#L4
https://github.com/deepmap/oapi-codegen/blob/fae4b9e07979f25a896eaa21816a7a00bb9a472d/go.mod#L8

---

the v1.19.0 seems to be an error: https://pkg.go.dev/github.com/exoscale/egoscale?tab=versions

https://github.com/go-acme/lego/commit/50a87b031e7c01fa75b20286a35082c3c7e931e7#r41994779

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,7 @@ require (
 	github.com/cloudflare/cloudflare-go v0.13.2
 	github.com/cpu/goacmedns v0.0.3
 	github.com/dnsimple/dnsimple-go v0.63.0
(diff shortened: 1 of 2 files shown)


dmke (approved):
Weird, indeed.

PR #1255, opened 2020-09-23
Add Hover as DNS-style provider

chickenandpork:
Support for Hover (http://hover.com/) (was: TuCows) as a DNS provider

This method mimics a HTTP client because Hover lacks a formal/public API.  The work is based on adapting some reverse-engineering into Go.


ldez:
Hello, in order for a PR adding a DNS provider to be accepted, you have to:

- [ ] add a description to your PR
- [ ] be able to maintain this provider
- [ ] have a homogeneous design with the other providers
- [ ] add tests (units)
```shell
make test
```
- [ ] add tests ("live")
https://github.com/go-acme/lego/blob/c938de68f262c057542d44bcb67ef961cefcc60e/providers/dns/cloudflare/cloudflare_test.go#L125-L151

```shell
make test
```
- [ ] add a [provider descriptor](https://github.com/go-acme/lego/blob/master/providers/dns/cloudflare/cloudflare.toml)
- [ ] generate CLI help, documentation, and readme.
```bash
make generate-dns
```
- [ ] be able to do: _(and put the output of this command to a comment in your PR)_
```bash
rm -rf .lego

./lego -m your@email.com --dns YOUR_PROVIDER_NAME -d *.example.com -d example.com -s https://acme-staging-v02.api.letsencrypt.org/directory run
```
Note the wildcard domain is important.
- [ ] pass the linter ([golangci-lint](https://github.com/golangci/golangci-lint#install) must be installed):
```shell
make checks
```
- [ ] do `go mod tidy`

ldez:
Hello,

If this provider doesn't have an API or official API documentation, I'm not sure we will agree to add this provider.

Also `github.com/chickenandpork/hoverdnsapi` has a dependency on lego, we will not allow that.

ldez:
Here is what I recommend:

- create a document describing the API and put it in a repository or in lego
- create in lego a light client instead of an external library

Without this, it will not be possible for us to accept your PR.

chickenandpork:
>  I'm not sure we will agree to add this provider

Does this mean literally "I am not sure", or is it a soft way of saying "we will not" ?  If the PR has zero chance of ever being merged, then we should just abandon it now, but if there's a chance I'll be able to use it to maintain the DNS records I use on my services, that would be immensely useful, hence I'd prefer to continue if there's a chance of being accepted.

ldez:
When I said  "I am not sure", I meant that I needed time to think about it.

For the rest, I think I answered in my previous comment https://github.com/go-acme/lego/pull/1255#issuecomment-698024302

If you want to use your external library you will have to modify it in order to remove lego dependencies.

ldez:
Do my answers allow you to make progress on the subject?

chickenandpork:
> Do my answers allow you to make progress on the subject?

I appreciate that you're blunt and clear; none of this USA "I don't think" meaning actually a polite "heck no!".  I understand if my implementation can lead to later maintenance concerns.

Honestly, my immediate response was somewhere between "aw shoot, I need to discard the work and start again" with the comment that was made at the same time mine was, and was unsure what example of a "light client" I could follow to make it as consistent as possible.  I mean: the words make sense, but it seems there's a specific example in mind, and consistency helps review.

...but I see that you're willing to go forward -- maybe at least to get the functionality there? -- with the external library if I can clean it up.  If I remember correctly, there was some import of types to ensure compatibility, but I can likely remove those.

I can try to move forward this weekend if that's OK, and I do really appreciate the feedback, I regret going silent for a while.

ldez:
By "light" client I mean something like that:

- https://github.com/go-acme/lego/tree/master/providers/dns/arvancloud/internal
- https://github.com/go-acme/lego/tree/master/providers/dns/cloudns/internal
- https://github.com/go-acme/lego/tree/master/providers/dns/conoha/internal
- https://github.com/go-acme/lego/tree/master/providers/dns/dynu/internal
- ...

chickenandpork:
Hi; Took me a while to move the dependent code to a `./internal/` to break the dependency -- changed country, changed jobs, hardware packed up shipping containers forever, etc.

This update is rebased, has the external dependency added, and unit tests OK.  The underlying code still functions fine as a CLI tool as well before copying tinto this module.

ldez:
I had completely forgotten about this PR.

The implementation was not really Go idiomatic.

I reviewed the PR and refactored the content (only based on the previous code).
It was complex and long.

It's not HTML scrapping because the responses are JSON, this is an API.
But this is the frontend API because there is no real API.

As the PR is very old I don't know if the implementation is still accurate.

But if a Hover's user wants to test it:

<details><summary>How to test this PR?</summary>

1. You need [Go](https://go.dev/doc/install)
2. Checkout the PR:
    ```bash
    git clone https://github.com/chickenandpork/lego.git
    cd lego
    git checkout 20200501-dns-type-provider-for-tucows-hover
    ```
3. Compile lego:
    - if you have `make`: `make build`
    - if you don't make `make`: `go build ./cmd/lego -o dist/lego`
4. Run the following command with your own information (email, domain, credentials):
    ```bash
    XXX=xxx \
    ./dist/lego -m your_email@example.com --dns hover -d *.example.com -d example.com -s https://acme-staging-v02.api.letsencrypt.org/directory run
    ```
   **The wildcard domain is important**
5. Before each run of the command, you should clean your local environment:
    ```bash
    rm -rf .lego
    ```

</details>

ldez:
After reading https://gist.github.com/dankrause/5585907, it seems like the implementation will not work because the endpoints have changed and the API is not usable if the account has 2FA.

So, it's sad but I will close this PR, I hope they will create a real API.

chickenandpork:
Understood, @ldez …  My Go code was a bit immature but it seemed there’s was always some additional step needed to get this PR approved. :(  sure, the provider isn’t cooperating, but what should we do if the provider doesn’t care but we want to use Lego to help gen carts?

ldez:
> what should we do if the provider doesn’t care but we want to use Lego to help gen certs?

You can use a CNAME approach https://letsencrypt.org/2019/10/09/onboarding-your-customers-with-lets-encrypt-and-acme.html
You should use a domain on another provider as a "proxy".

For example, if you have `example.org` (Hover) and `example.com` (another DNS provider) you can create a CNAME on `example.org` called `_acme-challenge.example.org` pointing to `challenge.example.com`. This way, you can obtain certificates for `example.org` through the other provider.

Issue #1256, opened 2020-09-24
May you support dynv6.com as DNS provider?


SamLowryMOI:
Hello,

dynv6.com uses a RFC2136 compliant way for TXT updates (see https://dynv6.com/docs/apis). They named lego as software on this site, but I can't find a way to use lego for getting a wildcard cert from LetsEncrypt, or more precisely, use traefik under Docker for this.

Please help. Thank you.

ldez:
Hello,

dynv6 API can be used with:

- an HTTP REST API https://dynv6.com/docs/apis#http.
- the [RFC2136](https://tools.ietf.org/html/rfc2136) https://dynv6.com/docs/apis#dns-update

lego support the [RFC2136](https://tools.ietf.org/html/rfc2136) API but not the HTTP API.

So you have to use the following documentation to configure the provider: https://go-acme.github.io/lego/dns/rfc2136/

SamLowryMOI:
Hello Idez,

thanks for the very quick reply. I didn't expect a RFC2136 item in the provider list, sorry.

I tried my best for a few hours but I can't get it working. 
<pre><code>docker run -it 
	-e RFC2136_NAMESERVER=ns1.dynv6.com 
	-e RFC2136_TSIG_ALGORITHM=hmac-sha512 
	-e RFC2136_TSIG_KEY=tsig-123.dynv6.com 
	-e RFC2136_TSIG_SECRET=YourSHAREDsecret==
	-e RFC2136_TTL=60 
goacme/lego 
	--accept-tos 
	--dns rfc2136 
	--dns.resolvers 8.8.8.8:53 
	--domains bla.dynv6.net  
	--email me@bla.org run  

2020/09/24 10:25:52 No key found for account me@bla.org. Generating a P256 key.
2020/09/24 10:25:52 Saved key to /.lego/accounts/acme-v02.api.letsencrypt.org/me@bla.org/keys/me@bla.org.key
2020/09/24 10:25:53 [INFO] acme: Registering account for me@bla.org
!!!! HEADS UP !!!!

Your account credentials have been saved in your Let's Encrypt
configuration directory at "/.lego/accounts".

You should make a secure backup of this folder now. This
configuration directory will also contain certificates and
private keys obtained from Let's Encrypt so making regular
backups of this folder is ideal.
2020/09/24 10:25:53 [INFO] [bla.dynv6.net] acme: Obtaining bundled SAN certificate
[...]

PR #1259, opened 2020-09-27
rfc2136: ensure TSIG algorithm is fully qualified

dmke:
Issue #1256 has revealed an issue with the way we read the TSIG algorithm from the environment.

The RFC specifies the algorithm to be fully-qualified, but the CLI docs don't communicate this clearly to the end user.

This PR just wraps the algorithm in a `dns.Fqdn()` call, ensuring a trailing dot.


ldez (approved):
LGTM

PR #1262, opened 2020-09-30
DNSProvider API with CreateRecord/DeleteRecord

matthiasng:
This MR extends the Provider interface for all DNS providers to create DNS records from a fqdn and value.

See #720 for more informations


ldez:
duplicate of #883

PR #1270, opened 2020-10-17
fix: use http.Header.Set method instead of Add.

ldez:
Use `http.Header.Set` method instead of `Add` to overwrite all existing header values when it's useful.


dmke (approved):
LGTM

Issue #1254, opened 2020-09-23
Openstack designate doesnt work as described in the documentation


yfoelling:
Hello,

i have an issue with the designate DNS, i put in exactly the environments asked in from the documentation but theres an error every time.

```
root@letsencrypt-proxy-1:~# env | grep OS_
OS_AUTH_URL=https://api.foo.de:5000/v3
OS_REGION_NAME=RegionOne
OS_PROJECT_NAME=Designate_Sandbox
OS_TENANT_NAME=Designate_Sandbox
OS_PASSWORD=verysecret
OS_USERNAME=designate_sandbox
root@letsencrypt-proxy-1:~# lego --domains="[domain]" --email="[email]" --dns="designate" run
2020/09/23 12:51:43 designate: Missing environment variable [OS_PROJECT_ID]
```
(FYI: OS_TENANT_NAME is deprecated since a while now, but lego also wont start without it)

After that i put in the optional variable for project id and tried some combination, but it still doesn't work..
```
root@letsencrypt-proxy-1:~# export OS_PROJECT_ID=[PROJECT_ID]
root@letsencrypt-proxy-1:~# env | grep OS_
OS_AUTH_URL=https://api.foo.de:5000/v3
OS_REGION_NAME=RegionOne
OS_PROJECT_NAME=Designate_Sandbox
OS_TENANT_NAME=Designate_Sandbox
OS_PASSWORD=verysecret
OS_USERNAME=designate_sandbox
OS_PROJECT_ID=[PROJECT_ID]
root@letsencrypt-proxy-1:~# lego --domains="[domain]" --email="[email]" --dns="designate" run
2020/09/23 12:52:08 designate: failed to authenticate: You must provide exactly one of DomainID or DomainName to authenticate by Username
root@letsencrypt-proxy-1:~# unset OS_PROJECT_NAME
root@letsencrypt-proxy-1:~# lego ---domains="[domain]" --email="[email]" --dns="designate" run
2020/09/23 12:52:31 designate: failed to authenticate: You must provide exactly one of DomainID or DomainName to authenticate by Username

```
obviously, i put in the right values instead if [PROJECT_ID] ;-)

I tried a few more combinations, but none worked sadly.

Generally i have access to the API on the same node with the same environment variables, that i tried for lego.

root@letsencrypt-proxy-1:~# openstack zone list
```
+--------------------------------------+-----------------+---------+------------+--------+--------+
| id                                   | name            | type    |     serial | status | action |
[...]

PR #1279, opened 2020-11-03
Deprecate OS_TENANT_NAME as required field

skatsaounis:
This PR removes the `OS_TENANT_NAME` from the required environment variables since, if set, it breaks authentication in OpenStack deployments where `OS_PROJECT_NAME` and `OS_PROJECT_ID` should be used.

For the record, the following message is received in case `OS_TENANT_NAME` is set only to pass `lego` validation check:

`Error: designate: failed to authenticate: You must provide exactly one of DomainID or DomainName to authenticate by Username`

Fixes: #1254


ldez (approved):
Thank you :+1:

skatsaounis:
@ldez Thanks for merging this one.

One question here is, should it be considered an enhancement or bug fix? In my opinion, it is the latter since in OpenStack Train which I use, I cannot use go-acme without this fix. And current OpenStack version is Victoria (two versions ahead).

I do not know your release cycle but I would love to have a ×.×.1 release with this change in 😄

ldez:
The information is in the labels and the milestone of this PR.

>  I cannot use go-acme without this fix

How do you use lego? as a CLI ? as a lib?

skatsaounis:
Thank you very much for the info. Ok, then I have to wait for 4.1 release. Any estimation regarding its availability?

Actually, I am using it through Terraform acme provider. So the next step for me is to open a PR here https://github.com/vancluever/terraform-provider-acme/blob/master/go.mod#L9 to use a version with the fix in place.

I know that I will wait a long time until finally be able to consume the change but I am trying to speed up things the best way possible.

PR #1302, opened 2020-11-23
vultr: Update to govultr v2 API

iquiw:
Hi, I got the following error when renewing certificate with lego v4.1.2 which uses govultr v1 API.
```
acme: cleaning up failed: vultr: API call failed: gave up after 4 attempts, last error unavailable (resp == nil)
```
Instead of tracking the error, I updated to use v2 API and succeeded to renew certificate.

CC: @ddymko


ddymko:
The error you posted is due to API V1 having a very small rate limit. One way to get past this is to set the `VULTR_HTTP_TIMEOUT` a bit higher.

However, API v2 was recently taken out of a `developer preview` and fixes a lot of the headaches...the small rate limit for one. So it is definitely better to update to this version of GoVultr (you beat me to it!).

Also just to point out the list calls return paginated responses and the default limit is to return `25` items in the response.  This will cause issues for any account that has more than 25 domains or 25 records per domain. 

https://github.com/go-acme/lego/pull/1302/files#diff-8c15b3caf76ccde7ea6efb0970a7bed6d6011a8ea91fbb82b7282cccba703eb5L146
https://github.com/go-acme/lego/pull/1302/files#diff-8c15b3caf76ccde7ea6efb0970a7bed6d6011a8ea91fbb82b7282cccba703eb5L173

iquiw:
@ddymko Thank you so much for the review and insight!

> The error you posted is due to API V1 having a very small rate limit. One way to get past this is to set the VULTR_HTTP_TIMEOUT a bit higher.

I suspected it but configured wrong environment variable (`VULTR_POLLING_INTERVAL`)...

> However, API v2 was recently taken out of a developer preview and fixes a lot of the headaches...the small rate limit for one. So it is definitely better to update to this version of GoVultr (you beat me to it!).

Oh, I overlooked the state of API v2.

> Also just to point out the list calls return paginated responses and the default limit is to return 25 items in the response. This will cause issues for any account that has more than 25 domains or 25 records per domain.

I might be able to fix it, but have no test environment.
I will try it later anyway. Or if you can submit PR, I am happy to close this and wait it (no hurry).

ddymko:
@iquiw If you want to finish up the PR feel free...you are pretty far along and the only missing piece here is the paging on the list calls. If you want want me to finish up the PR I don't mind doing that either.

iquiw:
Pushed paging change. As written, I did not test it.

Very much appreciate if you can check it.

ldez (approved):
Thanks :+1:

PR #1310, opened 2020-11-28
Add 'since' tag to HyperOne provider

ad-m:
See changelog at https://github.com/go-acme/lego/releases/tag/v3.9.0 .


ldez (commented):
Good catch, thank you :+1:

ldez (approved):

PR #1311, opened 2020-11-28
namecheap: unify documentation to other providers

ad-m:
Namecheap is only one provider which list provider restriction on provider list: https://go-acme.github.io/lego/dns/

I think this information fits more into the "Additional" section, similar to AWS Route53 (https://go-acme.github.io/lego/dns/route53/). I can see that this is a bit subjective, but I am bringing it to your consideration.

### DIFF
--- a/providers/dns/namecheap/namecheap.toml
+++ b/providers/dns/namecheap/namecheap.toml
@@ -2,11 +2,11 @@ Name = "Namecheap"
 URL = "https://www.namecheap.com"
 Code = "namecheap"
 Since = "v0.3.0"
-Description = '''
 
-Configuration for [Namecheap](https://www.namecheap.com).
+Additional = '''
+## Description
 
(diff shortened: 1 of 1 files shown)


ldez:
It's a choice: the criteria for accessing the API are relatively opaque (no public information) and this limitations is, from my point of view, very poorly documented by Namecheap.

So I will declined your PR, thanks anyway.

related to #1229

ad-m:
Maybe we should have an indication whether the provider is paid or free? DNSimple charge $6/zone/month Amazon Route 53 & Azure DNS charge $0.50/zone/month. the criteria for accessing the AP. I If the user wants to use Lego and has a DNS zone with an incompatible provider, the price criterion may be important to delege required records, as well as the ease of access to the Namecheap API.

ldez:
With Namecheap the use of API is not free, it's not related to domain, it's not the same thing.

PR #1315, opened 2020-12-07
hyperone: Minor tweaks of documentation

ad-m:
I am responsible at HyperOne for maintaining Lego integration.

`HYPERONE_ZONE_URI` env is not implemented.


ldez (approved):
Thanks :+1:

PR #1327, opened 2021-01-10
errors: Fix instance not being printed

nh2:
The condition so far printed it only when it was known to be empty.

Found in https://github.com/NixOS/nixpkgs/issues/101445#issuecomment-757352452


ldez (approved):
Thanks :+1:

ejder38 (commented):
nh2:errors-missing-instance-print

PR #1331, opened 2021-01-26
cloudns: Improve reliability

### LINKED ISSUE #1330: ClouDNS: acme challenge fails frequently
ClouDNS allows for multiple TXT records of the same host. Which in my opinion is correct

This creates following issue:
- The dns-01 challenge fails, when there is one ore more pre-existing DNS entries (e.g. from previous attempts)

ClouDNS has more name servers than those that appear in the propagation phase (currently 8 NS for my zone, only 4 that appear authoritative to lego)
- This will cause the during secondary verification. It may fails fail with NXDOMAIN

A tested fix is available, but I was not yet able to push it to github

david-aggeler:
- ~~Cleanup all TXT records before setting new ones~~
- Wait for NS server sync with API
- Updated test for better coverage

Fixes #1330

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,6 @@
 
 - **[dnsprovider]** vult, dnspod: fix default HTTP timeout.
 - **[dnsprovider]** pdns: URL request creation.
-- **[dnsprovider]** cloudns: fix reliability
 - **[lib]** errors: Fix instance not being printed
 
 ## [v4.1.3] - 2020-11-25
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 
 - **[dnsprovider]** vult, dnspod: fix default HTTP timeout.
 - **[dnsprovider]** pdns: URL request creation.
+- **[dnsprovider]** cloudns: fix reliability
 - **[lib]** errors: Fix instance not being printed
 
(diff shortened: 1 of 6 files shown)


ldez:
Could you run the following command (you have to replace the email and the domains) and put the output of this command to a comment in your PR:
```bash
rm -rf .lego

./lego -m your@email.com --dns cloudns -d *.example.com -d example.com -s https://acme-staging-v02.api.letsencrypt.org/directory run
```

Note the wildcard domain is important.

david-aggeler:
Point taken. It will timeout on the 'DNS record propagation' on the '*' record. So I've tested the original implementation. That one runs error free, but only returns one certificate:

2021/01/26 20:11:09 [INFO] [*.docker-test.example.com, docker-test.example.com] acme: Validations succeeded; requesting certificates
2021/01/26 20:11:10 [INFO] [*.docker-test.example.com] Server responded with a certificate.

Is the wildcard registration of all child hosts in parallel to the root host registration really a use case? 
If so, I see a few options, because the current implementation is so unstable (and difficult to troubleshoot without code)

- Can/Should the provider be put into the authSolversSequential category?
- Can the CleanUpAll be done all outside and before the domain specific Present(). like in prober.parallelSolve()?

What's wrong in original scenario, where I only get one certificate?

ldez (approved):

PR #1334, opened 2021-01-27
fix: preferred chain only match root name

### LINKED ISSUE #1333: --preferred-chain compatibility with LE Android vs OpenSSL choice
In <https://community.letsencrypt.org/t/openssl-client-compatibility-changes-for-let-s-encrypt-certificates/143816> LE detail their upcoming use of alternate chains, to let people choose to be compatible with unmaintained OpenSSL instead of compatible with unmaintained Android.  Some TLS-protected services I manage will need to bias in favor of bad OpenSSL, not bad Android.

As far as I can tell from #1227 and #1298 and the current help text, the first chain found where any certificate in the chain has an Issuer CN matching the supplied string will be used.  This is the same logic which certbot used to use, until certbot/certbot#8596 at which point they switched to only matching the root name.

1. I can't tell for sure without something to test against, but by inspection it looks as though the newer certbot behavior is needed for this situation, because my desired chain is, from the `[sequence, of, CN]` perspective, a subset of the first chain; thus there is there's no way to match the desired chain without also matching the first chain, so the current behavior is not useful for this ACME server in this scenario
2. I can't see any options for introspection/logging which would let people step through, perhaps interactively, and see the available options.

I think in my ideal scenario, I'd be able to use `lego [...] run --interactive` or `lego [...] run --preferred-chain='?'` to trigger a flow where lego will pause after getting the available chains, show the chains, let someone type/paste a matcher string, then list the matched results and show which one will be used and get confirmation, before continuing through to get the final cert.  This flow would let folks see the available options and figure out the exact string to use, without excessive API requests.

Does this all make sense?  Have I gone horribly wrong somewhere in my understanding?

Thanks.

ldez:
Based on https://github.com/certbot/certbot/pull/8596

Fixes #1333

### DIFF
--- a/certificate/certificates.go
+++ b/certificate/certificates.go
@@ -540,10 +540,10 @@ func hasPreferredChain(issuer []byte, preferredChain string) (bool, error) {
 		return false, err
 	}
 
-	for _, cert := range certs {
-		if cert.Issuer.CommonName == preferredChain {
-			return true, nil
-		}
+	topCert := certs[len(certs)-1]
+
+	if topCert.Issuer.CommonName == preferredChain {
+		return true, nil
 	}
 
 	return false, nil


dmke:
That looks suspiciously simple for a solution/fix. I'll need some time to wrap my head around the problem to understand the fix. I hope to get to it today in the evening.

dmke (approved):
LGTM, the description in https://github.com/certbot/certbot/pull/8596 was helpful understanding this.

PR #1335, opened 2021-01-28
Log reason for propagation check failure

apeschar:
I had trouble diagnosing an issue where the authoritative nameserver was set incorrectly, but the challenge record resolved without issues.  This wasn't clear until I added some logging to clarify the reason for the propagation check failing.

I believe this would be helpful for others trying to troubleshoot the same issue.


ldez:
Hello,

I don't want to display the error on each iteration because it will be useless information in 99% of the case.

So I will decline your PR.

PR #1354, opened 2021-02-25
dnspod: improve search accuracy when a domain have more than 100 records

hjweddie:
When there are more than 100 records in the domain, Records.List function without providing record name, may fail to receive  the records of the _acme_challenge TXT record. Because dnspod records list api return maximum 100 records one time.  So CleanUp will not delete the TXT records.

We can provide record name into Records.List function. This will search _acme_challenge TXT record more accurately, and with less return records, less than 100.


ldez (approved):
Thanks :+1:

Issue #1218, opened 2020-07-16
Support Openstack Application Credentials


johscheuer:
Openstack support [application credentials](https://docs.openstack.org/keystone/train/user/application_credentials.html) to create a "technical" user with less permissions. It would be nice if lego would support these application credentials. This would be partially solved by https://github.com/go-acme/lego/issues/1217 since you could specify the applications credentials in the `clouds.yml` file but we should also support setting the credentials over environment variables.

PR #1359, opened 2021-02-27
gophercloud: update to v0.16.0

rissson:
This PR updates gophercloud. It is my first step towards implementing #1218. I will soon open a PR for it, which will depend on this one.


ldez:
Maybe you can update the lib in your next PR.

Also, the update of github.com/gophercloud/gophercloud to v0.16.0 doesn't require updating `github.com/gophercloud/utils` and `gopkg.in/yaml.v2`, it's not a major problem but you don't provide any information about that.

rissson:
First, thanks for your swift response! I'm not quite confident in my usage of go packaging. `gopkg.in/yaml.v2` updated itself when I run `go get -u github.com/gophercloud/gophercloud` and `go get -u github.com/gophercloud/utils`, maybe it's some kind of dependency, although that seems weird. Is there anything I can do to avoid that?

Also, about updating `github.com/gophercloud/utils`, it makes sense to do so, as `github.com/gophercloud/gophercloud` and `github.com/gophercloud/utils` are supposed to work together.

ldez:
I will close this PR in favor of  #1360

PR #1361, opened 2021-02-28
desec: fix error with default TTL

Knight1:
When I do not specify "DESEC_TTL=3600" the default behavior is this error. 

```
acme: error presenting token: desec: failed to create records: domainName=, recordName=_acme-challenge: 400: body: {"ttl":["Ensure this value is greater than or equal to 3600."]}
```

desec.io only allows TTL below 3600 if you contact there support and they approve this on your account. So the default behavior is always resulting in an error unless you contact support. 

See: 
https://desec.readthedocs.io/_/downloads/en/latest/pdf/
https://github.com/desec-io/desec-stack/issues/216

This PR set's the TTL to 3600 so there are no longer any errors.

### DIFF
--- a/providers/dns/desec/desec.go
+++ b/providers/dns/desec/desec.go
@@ -24,6 +24,10 @@ const (
 	EnvHTTPTimeout        = envNamespace + "HTTP_TIMEOUT"
 )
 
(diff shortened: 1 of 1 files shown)


ldez (approved):
Thanks :+1:

PR #1362, opened 2021-02-28
add DUCKDNS_SUBDOMAIN environment variable to support own domain.

ymdmkhk:
Allow Duck DNS sub domain to be specified in environment variable to support own domain name.


ldez:
seems to be a duplicate of #952

ldez:
The full support of CNAME has been added to v4.9.

PR #1363, opened 2021-03-01
Cleanup stale records

vanbroup:
The record ID is only available when cleaning up our own records and in the same run, this change cleans up stale records from previously failed or third party runs by looking up the record ID.


vanbroup:
@ldez looks like we got a build timed-out, can you re-run it?

PR #1364, opened 2021-03-01
Cleanup stale records

vanbroup:
The record ID is only available when cleaning up our own records and in the same run, this change cleans up stale records from previously failed or third party runs by looking up the record ID.


ldez:
Hello,

by default, lego clean the record even if the command failed, how do you produce those stale records?

vanbroup:
There are two reasons why you can run into problems without this change, first if you have used a different client that doesn't cleanup the records. Second, when the lego process gets interrupted before cleaning up.

ldez:
I don't think that the role of lego is to clean TXT records not created by himself.

A side effect will be that only it will not possible to run 2 instances of lego at the same time (ex: one for a wildcard and one for the root domain).

Otherwise, your PR only removes the first record.

We will decline your PR.

PR #1367, opened 2021-03-04
chore: update linter

ldez:
Update golangci-lint to v1.38.0.

### DIFF
--- a/providers/dns/joker/provider_dmapi.go
+++ b/providers/dns/joker/provider_dmapi.go
@@ -26,7 +26,7 @@ func newDmapiProvider() (*dmapiProvider, error) {
 		var errU error
 		values, errU = env.Get(EnvUsername, EnvPassword)
 		if errU != nil {
-			// nolint:errorlint
+			// nolint:errorlint,errorlint // false-positive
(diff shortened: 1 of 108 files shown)


providers/dns/joker/provider_dmapi.go
```diff
@@ -26,7 +26,7 @@ func newDmapiProvider() (*dmapiProvider, error) {
 		var errU error
 		values, errU = env.Get(EnvUsername, EnvPassword)
 		if errU != nil {
-			// nolint:errorlint
+			// nolint:errorlint,errorlint // false-positive
```
dmke, line 29: Is this a new syntax for the linter (i.e. `//nolint:$name,$name`) or do you *really* want to disable `errorlint` here?

dmke (approved):
I believe I've found a small typo, genuinely not sure (does `// nolint:errorlint,errorlint` do anything special?)

Otherwise LGTM, :shipit:

providers/dns/joker/provider_dmapi.go
ldez: just a mistake: :smile:

PR #1374, opened 2021-03-12
powerdns: several improvements

pieterlexis:
Hi folks,

This PR has several improvements to the PowerDNS DNS Provider.

### Server name
The PowerDNS API allows for a `server_name` other than 'localhost'. The auth server itself does not use it, but proxies that have multiple backends can. We are running such a proxy. The first commit in this series adds the `PDNS_SERVER_NAME` option to set this. The default remains the same.

### content-type in PATCH requests
Some proxies require the Content-Type header to be set. The second commit sets this header to 'application/json' when the method is not DELETE or GET.

### 'records' can be omitted in rrset
The API allows the 'records' field to be omitted in the request when the `changetype` is 'DELETE'. The third commit implements this omission.

### Error reporting
I was confused by an error report that had the word 'fetch' hardcoded. The last commit changes this to the HTTP Method used.

If there is anything you need more in this PR, let me know.

Cheers,

Pieter Lexis
PowerDNS

PR #1376, opened 2021-03-17
make: fix validate-docs

dmke:
In #1374 the PDNS documentation was amended, but the author forgot to re-generate the docs.

We do have a validation check in the main workflow, but it didn't catch the change because the underlying Makefile target was broken.

This commit replaces the Makefile conditional with an equivalent shell conditional.

<details><summary>Explanation</summary>

The conditional checking for a dirty working directory ran before generating the documentation. In effect, `make validate-doc` only
did the right thing on the second invokation.

Consider this simplified Makefile example:

```make
generate:
	aaaa
	
validate: generate
ifneq ($(shell bbbb),)
	cccc
else
	dddd
endif
```

The order of evaluation for `make validate` is:

1. detect `generate` target, no dependencies,
2. set body for `generate` target to `aaaa`,
3. detect `validate` target, with dependency `generate`,
4. execute `bbbb`, compare with empty string,
   - if not empty: set body for `validate` target to `cccc`
   - else: set body for `validate` target to `dddd`
5. invoke `generate` → execute `aaaa` (`generate` is a dependency of `validate`),
6. invoke `validate` → execute `cccc` or `dddd`.

This also means that step 4 is also (unnecessarily) executed when the `validate` target is never invoked.

</details>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,10 +72,12 @@ generate-dns:
 	go generate ./...
 
 validate-doc: generate-dns
-ifneq ($(shell git status --porcelain -- ./docs/ ./cmd/ 2>/dev/null),)
-	@echo 'The documentation must be regenerated, please use `make generate-dns`.'
-	@git status --porcelain -- ./docs/ ./cmd/ 2>/dev/null
-	@exit 2
-else
-	@echo 'All documentation changes are done the right way.'
-endif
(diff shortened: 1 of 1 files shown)


ldez (approved):
LGTM

PR #1381, opened 2021-04-07
nifcloud: Add a NIFCLOUD_HOSTED_ZONE parameter

kakakikikeke-fork:
I updated dns nifcloud provider. If NIFCLOUD_HOSTED_ZONE is set, preferentially use it as the zone name, otherwise use the domain value as the zone name.

ex)

```
NIFCLOUD_HOSTED_ZONE=your.hosted.zone \
NIFCLOUD_ACCESS_KEY_ID=xxxxxxx \
NIFCLOUD_SECRET_ACCESS_KEY=xxxxxxx \
lego --email your-name@mail.domain --dns nifcloud --domains *.sub.your.hosted.zone --accept-tos run
```

### DIFF
--- a/cmd/zz_gen_cmd_dnshelp.go
+++ b/cmd/zz_gen_cmd_dnshelp.go
@@ -1344,6 +1344,7 @@ func displayDNSHelp(name string) error {
 		ew.writeln()
 
 		ew.writeln(`Additional Configuration:`)
+		ew.writeln(`	- "NIFCLOUD_HOSETD_ZONE":	Hosted zone name`)
 		ew.writeln(`	- "NIFCLOUD_HTTP_TIMEOUT":	API request timeout`)
 		ew.writeln(`	- "NIFCLOUD_POLLING_INTERVAL":	Time between DNS propagation check`)
 		ew.writeln(`	- "NIFCLOUD_PROPAGATION_TIMEOUT":	Maximum waiting time for DNS propagation`)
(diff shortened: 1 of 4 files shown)


ldez:
Hello,

could you explain why do you create this PR? what's the concrete problem behind it?

kakakikikeke-fork:
Hi,

Thank you for reviews.

## Goal
Make it possible to obtain a certificate with a common name such as the following (because it cannot be obtained at present).

hostedZone = "your.hosted.zone"

* aaa.your.hosted.zone
* aaa.bbb.your.hosted.zone
* *.bbb.your.hosted.zone

## Bug
The hostedZone may not be found because you are searching for the nifcloud hostedZone using fqdn specified in --domains. If the common name of the certificate you want and the hierarchy of hostedZone are different, you can get it by specifying NIFCLOUD_HOSTED_ZONE. Also, since the NIFCLOUD_HOSTED_ZONE parameter is an optional parameter, it will work even if it is not specified as before.

ldez:
> The hostedZone may not be found because you are searching for the nifcloud hostedZone using fqdn specified in --domains. If the common name of the certificate you want and the hierarchy of hostedZone are different

How it is possible to produce this case?
Are you sure of the root cause?
Could you open an issue and provide detailed information?

kakakikikeke-fork:
I opened a issue https://github.com/go-acme/lego/issues/1382. Thanks.

kakakikikeke-fork:
Closed. I created new p-r https://github.com/go-acme/lego/pull/1383.

PR #1391, opened 2021-04-20
exoscale: fix DNS provider debugging

falzm:
This change fixes the `exoscale` DNS provider debug mode activation.


ldez (approved):
Thanks :+1:

PR #1403, opened 2021-05-14
lib: improve wait.For returns.

### LINKED ISSUE #1401: wait.For() can return ambigous errors
- [x] Yes, I'm using a binary release within 2 latest releases.
- [x] Yes, I've included all information below (version, config, etc).

`wait.For` returns nil error if deadline exceeded

### What did you see instead?

If certificate got deadline `For()` returns nil error wrapped by fmt: 

```
obtaining certificate: error: one or more domains had a problem:
[example.com] time limit exceeded: last error: %!w(<nil>)
```

So, this is not good: it's way more better if we can replace timeout to context and return DeadlineExceeded or bare error which returns from callback

### Steps to reproduce

1. idk, just put into `wait.For()` some callback, which never returns any error

### Details

<details><summary>Version of lego</summary>

discovered on `v4.2.0` reproduces as well on `v4.3.1`
</details>

<details><summary>Logs</summary>
no any logs, idk
</details>

ldez:
Fixes #1401

### DIFF
--- a/platform/wait/wait.go
+++ b/platform/wait/wait.go
@@ -1,6 +1,7 @@
 package wait
 
 import (
+	"errors"
 	"fmt"
 	"time"
 
@@ -16,6 +17,9 @@ func For(msg string, timeout, interval time.Duration, f func() (bool, error)) er
 	for {
 		select {
 		case <-timeUp:
+			if lastErr == nil {
+				return errors.New("time limit exceeded")
+			}
 			return fmt.Errorf("time limit exceeded: last error: %w", lastErr)
 		default:
 		}
--- a/platform/wait/wait_test.go
+++ b/platform/wait/wait_test.go
@@ -13,7 +13,7 @@ func TestForTimeout(t *testing.T) {
 		})
 	}()
 
-	timeout := time.After(5 * time.Second)
+	timeout := time.After(6 * time.Second)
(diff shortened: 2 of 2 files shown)


quenbyako:
@ldez i though about something different which is using context with timeout)

something like
```go 
func For(ctx context.Context, repeatInterval time.Duration, f func() (bool, error)) error {
...
}
```
also, i don't think that it's a good idea to return last error: if `f` returns error, it must be throwed explicitly, cause you can lost more important errors in the middle of updating process.

ldez:
For now, I don't want to change the signature of the function.

> also, i don't think that it's a good idea to return last error: if f returns error, it must be throwed explicitly, cause you can lost more important errors in the middle of updating process.

this is the goal of this function to **not** throw an error immediately and to only return the last error.

dmke (approved):

quenbyako:
> @ldez return the last error

Ehhh, okay, it just looks pretty weird for me, but okay, you're a member and i'll not argue for my subjective opinion, so alright)

Also, thanks a lot for so fast fix and update! I really appreciate that 🤗

dmke:
> it just looks pretty weird for me

Naming is hard. A better name for `For()` would have been `RetryUntil()`.

PR #1405, opened 2021-05-14
hostingde: autodetection of the zone name.

### LINKED ISSUE #1404: hostingde: Make HOSTINGDE_ZONE_NAME optional

### Detailed Description

With hosting.de, it is currently impossible to get one certificate for multiple second-level domains, as the zone name is required as an environment variable.

What I would expect is that for the domain `something.great.example.org`, it searches for zones named `something.great.example.org` and, if that doesn't exist, for `great.example.org` or `example.org`. `org` is not a valid zone name on hosting.de, so we can stop there.

ldez:
Fixes #1404


dmke (approved):
LGTM

PR #1407, opened 2021-05-22
lightsail: improve documentation

dicksonleong:
I was using the Lightsail provider and noticed that the documentation was lacking, so i added some description (based on route 53 documentation), hope can help others who will be using it in in the future :)


ldez (approved):
Thanks :+1:

Clement-Jean:
Would be great if we can a bit more detail to solve #1436

Issue #1432, opened 2021-06-14
Azure DNS-01: allow supplying the ZONE name instead of gathering it from DNS


fibbs:
- [ x] Yes, I've searched similar issues on GitHub and didn't find any.

### Detailed Description

I am trying to build a setup where I want to use dns-01 challenges to generate certificates on a subdomain for fqdns that are not reachable from outside of the corporate network. Also, I don't want the public DNS to have any public records that might expose the corporate network internal structure. For this, I am using Corporate DNS to redirect *.dev.mycompany.org as a CNAME to my (Docker) server which runs a reverse proxy (traefik) which uses lego (lib) to auto-generate certs using dns-01 challenge. In external, public DNS I have an NS record delegating the subdomain dev.mycompany.org to an Azure DNS zone. This publicly available zone is empty, I want lego to update this zone with the appropriate TXT records for certificate validation.

Now, this fails due to the fact that the lego dns provider module for azure tries to gather the name of the zone by traversing through the DNS and finding the SOA record for the zone managing the host name for which I want to generate a certificate for, and uses this zone to instruct Azure API to create the TXT within.

Now, the server running lego on (in fact, the container) is inside my corporate environment, uses the internal DNS, thus does not have the NS record being available in the public DNS. This causes lego's dns01.FindZoneByFqdn function to not find this "sub domain" as the responsible zone but the "upper" zone, in my case "mycompany.org". This provokes an error at Microsoft Azure side: 

> Failure responding to request: StatusCode=403 -- Original Error: autorest/azure: Service returned an error. Status=403 Code=\"AuthorizationFailed\" Message=\"The client '<client object id>' with object id '<client object id>' does not have authorization to perform action 'Microsoft.Network/dnsZones/read' over scope '/subscriptions/<subscription id>/resourceGroups/<resource group>/providers/Microsoft.Network/dnsZones/mycompany.org' or the scope is invalid.

This is true, the zone is invalid, because it lego tries to instruct Azure to add a record into the zone "mycompany.org" instead of "dev.mycompany.org".

As a solution to this, I would wish to have an additional ENV variable, i. e. `AZURE_ZONE` which would skip the DNS resolve of the zone and just use the name I would supply.

I hope I was able to explain this request clearly.

fibbs:
Pull request here: https://github.com/go-acme/lego/pull/1433

dmke:
Side note:

> I don't want the public DNS to have any public records that might expose the corporate network internal structure.

[...]

PR #1433, opened 2021-06-19
azure: zone name as environment variable

fibbs:
This pull request solves issue #1432 (feature request).

In some cases you might want to use acme to fetch certificates while the server doing so does not have access to (public) DNS and therefore cannot fetch the SOA record of the domain. This happens to me with a subdomain. I have this subdomain in the authoritative DNS server of another provider, forwarding one sub-domain to an azure DNS hosted zone. This NS Record is not set at my company's internal DNS server, because I do not want to have the records inside the sub-domain to be visible from the internet.

For this to work, I need to supply the zone name within which lego should create the TXT validation records instead of letting lego gather this from DNS. This is what this pull request solves by adding an additional (optional) environment variable AZURE_ZONE_NAME.


providers/dns/azure/azure.toml
```diff
@@ -20,6 +20,7 @@ Example = ''''''
     AZURE_PROPAGATION_TIMEOUT = "Maximum waiting time for DNS propagation"
     AZURE_TTL = "The TTL of the TXT record used for the DNS challenge"
     AZURE_METADATA_ENDPOINT = "Metadata Service endpoint URL"
+	AZURE_ZONE_NAME = "Zone name to use inside Azure DNS service to add the TXT record in"
```
dmke, line 23: Nit: 4 spaces instead of tabs please.

providers/dns/azure/azure.go
```diff
@@ -169,9 +170,15 @@ func (d *DNSProvider) Present(domain, token, keyAuth string) error {
...
-		return fmt.Errorf("azure: %w", err)
+	var zone string
+	if env.GetOrFile(EnvZoneName) != "" {
+		zone = env.GetOrFile(EnvZoneName)
+	} else {
+		var err error
+		zone, err = d.getHostedZoneID(ctx, fqdn)
+		if err != nil {
+			return fmt.Errorf("azure: %w", err)
+		}
```
dmke, line 181: This should be moved into `getHostedZoneID`.

dmke (commented):
This sounds like a useful feature to have, not only for the Azure provider, but for all DNS providers. I believe `lego --dns.providers=...` should be able to this already, though (if not, we should move the logic of `dns01.FindZoneByFqdn` for a more generic solution).

I've left some remarks on this PR, for the case we don't want to change `dns01.FindZoneByFqdn`.

providers/dns/azure/azure.go
fibbs: Agree, done so.

providers/dns/azure/azure.toml
fibbs: oops. fixed.

fibbs:
> This sounds like a useful feature to have, not only for the Azure provider, but for all DNS providers. I believe `lego --dns.providers=...` should be able to this already, though (if not, we should move the logic of `dns01.FindZoneByFqdn` for a more generic solution).

To be honest, I have not checked how exactly other of the providers work exactly. I was supposing that this "get the actual zone within which the TXT record is to be added" was something specific to how the structure is in Azure DNS was, and that probably other provider's code could be working completely different. So, I do believe that fixing this is good enough only for Azure, as I don't know whether the function `dns01.FindZoneByFqdn` might be used by something completely different though.

Anyway, that's of course something you as the developers of this project have much more insights into. I would be happy with either solution.

> I've left some remarks on this PR, for the case we don't want to change `dns01.FindZoneByFqdn`.

Fixed :)

ldez:
@fibbs the message of @dmke is for me :wink:

providers/dns/azure/azure.go
```diff
@@ -169,9 +170,9 @@ func (d *DNSProvider) Present(domain, token, keyAuth string) error {
 	ctx := context.Background()
 	fqdn, value := dns01.GetRecord(domain, keyAuth)
 
-	zone, err := d.getHostedZoneID(ctx, fqdn)
-	if err != nil {
-		return fmt.Errorf("azure: %w", err)
+    zone, err := d.getHostedZoneID(ctx, fqdn)
+    if err != nil {
+            return fmt.Errorf("azure: %w", err)
```
dmke, line 175: In Go code, we do want tabs, not spaces :-)

Please run `gofmt -s -w providers/dns/azure/azure.go` to correct this automatically.

providers/dns/azure/azure.go
```diff
@@ -242,6 +244,10 @@ func (d *DNSProvider) CleanUp(domain, token, keyAuth string) error {
 
 // Checks that azure has a zone for this domain name.
 func (d *DNSProvider) getHostedZoneID(ctx context.Context, fqdn string) (string, error) {
+	if env.GetOrFile(EnvZoneName) != "" {
+	    return env.GetOrFile(EnvZoneName), nil
```
dmke, line 248: Since `GetOrFile` might read a file, its better to cache that return value:

```suggestion
	if zone := env.GetOrFile(EnvZoneName); zone != "" {
	    return zone, nil
```

Performance doesn't matter here, but this is more idiomatic Go code.

ldez (approved):

ldez:
@sergiomcalzada no this PR does not break anything, it's impossible.

ldez:
This PR is a part of v4.5 (created one year ago).

The content of this PR is clear and simple: there are no possible side effects.

So your problem is related to something else

ldez:
@sergiomcalzada Please open a dedicated issue.

sergiomcalzada:
Tracked it to the origin
https://github.com/vancluever/terraform-provider-acme/issues/268

Will try to manage from there the issue, but seems related to the latest lego update

PR #1438, opened 2021-06-27
lightsail: wrong Region env var name

Clement-Jean:
"LIGHTSAIL_" was mentioned two times in the key for lightsail provider


ldez (approved):
LGTM

PR #1439, opened 2021-06-28
aws: detailed credentials

Clement-Jean:
Make the Lightsail and Route53 documentation less confusing and more detailed.


ldez (approved):
LGTM

PR #1440, opened 2021-06-28
provider: index supported provider via custom type

burgesQ:
### why 

I'm working on introducing a dns01 challenge resolver to a project  using the lego library.

I realize no ways existed to (de)serialize / verify the a given provider before the creation of the  DNSResolver.. So I started implemting this feature (to read / dump to config file, loop over cli enum flag for viper etc ...)  

While working on introducing the type, I realize I could publish it to the library via MR :) 

+ I also realize down the path that I  could refactor the huge switch case from `./provider/dns/dns_provider.go` to  a simpler indexing over map  

### what 

A new type is introduced: `SupportedProvider` - which is an alias to int.

And array of const over iota is used to set the value of every provider.

The type implement the interface Stringer + the MarshalJSON / UnmarshalJSON one,
useful for json (de-)serialization, or even cli flag feature (see
 github.com/thediveo/enumflag).

A special error is yield in case of invalid value provided to set
the type: `ErrUnsupportedProvider`.

A factory signature is introduce (`providerFactory`, which is an alias to
`func() (challenge.Provider, error)`) to dump the switch case in favor of an
index over a `map[supportedProvder]ProviderFactory`.

two other new methods: `IsProviderSupported` and `GetSupportedProvider`

### enhancement 

- [ ] update the cli tools to loop over the factory type instead of raw str (didn't check if viper was used for the cli tools ?) (cc github.com/thediveo/enumflag)


ldez:
Hello,

your approach is interesting but there are some elements that I don't agree with, ex:
- methods that return interface
- map to store the providers
- etc
 
If you want to create a "custom" provider you just have to recreate/wrap the `NewDNSChallengeProviderByName` method.

I will decline your PR, thank you.

PR #1462, opened 2021-08-09
alidns: support ECS instance RAM role

bestmike007:
ECS instance RAM role w/ Go SDK: https://www.alibabacloud.com/help/doc-detail/66223.html


bestmike007:
@ldez no idea why tests fail:

image

ldez (approved):
Thank you :+1:

PR #1494, opened 2021-09-29
edgedns: fix Present and CleanUp logic

janosi:
The current Present logic of the EdgeDNS provider tries to create the TXT record even if it has just updated the existing TXT record.
The current CleanUp logic of the EdgeDNS provider deletes the TXT record even if it found that there are still challenges in the TXT record that should remain there.

This commit is to fix these issues.


ldez (approved):
Thanks

PR #1501, opened 2021-10-03
Add support for ZeroSSL account registration

fln:
This commit extends lego library and cli tool to support issuing
certificates from ZeroSSL without having to manually create an account.

Without this commit ZeroSSL can be used but users need to manually
create ZeroSSL account and start `lego` in EAB (External Account
Binding) mode.

From the `lego` cli tool perspective this commit:

Detects if `lego` ir running with ZeroSSL ACME directory `--server
https://acme.zerossl.com/v2/DV90` and uses ZeroSSL API to issue keys for
EAB. There is no need to provide `--eab`, `--kid`, `--hmac` values
anymore.

From the library perspective this commit:

Creates new method `RegisterWithZeroSSL()` in the `registration`
package which takes care of creating ZeroSSL account with a given email.
Internally it re-uses `RegisterWithExternalAccountBinding()` method
after KID and HMAC are retrieved from ZeroSSL registration endpoint.


ashkitten:
i would also like this feature. i've actually rebased this patch and am using it already for myself.

jinjianming:
Why haven't there been any updates

ldez:
Closes by #2837

PR #1508, opened 2021-10-12
rackspace: change zone ID to string

chrishepner:
Attempting a DNS challenge request with the Rackspace provider fails with an error in the following form:
```
2021/10/11 21:07:31 Could not obtain certificates:
[example.com] [example.com] acme: error presenting token: rackspace: json: cannot unmarshal string into Go struct field HostedZone.domains.id of type int
```

Querying https://dns.api.rackspacecloud.com/v1.0/account/domains?name=example.com shows that the domains.id field is now being returned as a string, rather than an int:
```
{"totalEntries":1,"domains":[{"id":"123456", ...}]}
```

This commit accordingly changes HostedZone and depending code to expect a string rather than an int.

### DIFF
--- a/providers/dns/rackspace/client.go
+++ b/providers/dns/rackspace/client.go
@@ -62,7 +62,7 @@ type ZoneSearchResponse struct {
 
 // HostedZone HostedZone.
 type HostedZone struct {
-	ID   int    `json:"id"`
+	ID   string `json:"id"`
 	Name string `json:"name"`
 }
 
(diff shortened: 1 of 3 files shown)


ldez:
Hello,

the documentation seems to say the opposite:
https://docs.rackspace.com/docs/cloud-dns/v1/api-reference/domains#list-domains-by-name

```
Status: 200 OK
Date: Thu, 28 Jul 2011 21:54:21 GMT
X-API-VERSION: 1.0.17
Content-Type: application/json
Content-Length: 202

{
  "domains" : [ {
    "name" : "region2.example.net",
    "id" : 2725352,
    "updated" : "2011-06-23T20:21:06.000+0000",
    "created" : "2011-06-23T19:24:27.000+0000"
  } ],
  "totalEntries" : 114
}
```

ldez:
But the API documentation was not updated for 5 years:
https://github.com/rackerlabs/docs-cloud-dns/tree/master/api-docs/v1/api-reference/methods

:thinking: maybe they changed their API and forget to update the documentation...

robshep:
I queried this with Rackspace.

> "In late September/early October our DNS team had pushed out an update that changes any newly added domains to use a new alphanumeric string instead of the shorter integer.  This is not a change we can roll back as these are the new style of UUID that the DNS API will be using going forward.  As of yet the documentation does not reflect this but we are hoping to get it updated in the near future.  I have created an internal report to help push this along."

hannesdejager:
This seems to be documented [here](https://docs.rackspace.com/docs/cloud-dns/v1/release-notes/#v1-0-35-environment-changes-september-21-2021).

When will a release with this fix be forthcoming? I'm using Traefik thats using Lego and I'm a bit stuck with this now.

PR #1520, opened 2021-11-01
chore: homogeneous tests setup and close servers.

ldez:
- fix missing `server.Close`
- refactor method to setup tests

### DIFF
--- a/providers/dns/servercow/internal/client_test.go
+++ b/providers/dns/servercow/internal/client_test.go
@@ -12,19 +12,21 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func setupAPIMock() (*Client, *http.ServeMux, func()) {
-	handler := http.NewServeMux()
-	svr := httptest.NewServer(handler) // FIME t.Cleanup(server.Close)
+func setupTest(t *testing.T) (*Client, *http.ServeMux) {
+	t.Helper()
+
+	mux := http.NewServeMux()
+	server := httptest.NewServer(mux)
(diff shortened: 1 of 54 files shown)


providers/dns/servercow/internal/client_test.go
```diff
@@ -14,7 +14,7 @@ import (
 
 func setupAPIMock() (*Client, *http.ServeMux, func()) {
 	handler := http.NewServeMux()
-	svr := httptest.NewServer(handler)
+	svr := httptest.NewServer(handler) // FIME t.Cleanup(server.Close)
```
dmke, line 17: Is this FIXME intentional?

dmke (changes requested):
Thank you for bringing `t.Cleanup` to my attention. I didn't know it even existed - this is a great feature (I'll start using this in my own code)!

In `servercow/internal/client_test.go` there's a FIXME left over (I think). Will the tests fail if there's no `t.Cleanup(server.Close)`? The code doesn't look like it would, though.

providers/dns/servercow/internal/client_test.go
ldez: oops!

ldez: if you like `t.Cleanup`, you can take a look to `t.Setenv()`

dmke (approved):

providers/dns/servercow/internal/client_test.go
dmke: Neat!

PR #1561, opened 2022-01-15
azure: add support for Azure Private Zone DNS

knutejoh:
I've added support for Azure Private Zone DNS to be able to utilize private zones when running a private ACME server inside Azure


knutejoh:
@ldez Just tested your structual changes locally with our Azure setup, and it's still working as expected :)

ldez (approved):

PR #1565, opened 2022-01-18
feat: extended support of cert pool

### LINKED ISSUE #1490: Add option to have LEGO_CA_CERTIFICATES add to the local cert pool instead of replacing it
### Welcome

### How do you use lego?

Through Traefik

### Detailed Description

As described in https://community.traefik.io/t/lets-encrypt-x509-certificate-signed-by-unknown-authority/11112 I had been wondering for two months why Traefik refused to talk to the default Let's Encrypt servers while `curl` and even a custom `go` test program had no problems. Eventually I noticed that I had set `LEGO_CA_CERTIFICATES` to my local CA's root cert, assuming this would allow the cert in addition to the default certs. In reality, this fully replaces the default cert pool.

I could imagine the following solutions:
* Have `LEGO_CA_CERTIFICATES` add to the default pool by default
* Add a new environment variable `LEGO_ADDITIONAL_CA_CERTIFICATES` that adds to the default pool
* Allow multiple file names like `LEGO_CA_CERTIFICATES=/etc/traefik/acme.crt,/etc/ssl/certs/ca-certificates.crt`

ldez:
- `LEGO_CA_CERTIFICATES` now accepts multiple file paths to be added by using `os.PathListSeparator` (`:` on POSIX, `;` on Windows) as a separator.
- `LEGO_CA_SYSTEM_CERT_POOL` initiates the cert pool with a copy of the system cert pool.

Fixes #1490

### DIFF
--- a/lego/client_config.go
+++ b/lego/client_config.go
@@ -7,6 +7,8 @@ import (
 	"net"
 	"net/http"
 	"os"
+	"strconv"
+	"strings"
 	"time"
 
 	"github.com/go-acme/lego/v4/certcrypto"
@@ -16,9 +18,14 @@ import (
 const (
 	// caCertificatesEnvVar is the environment variable name that can be used to
 	// specify the path to PEM encoded CA Certificates that can be used to
-	// authenticate an ACME server with a HTTPS certificate not issued by a CA in
+	// authenticate an ACME server with an HTTPS certificate not issued by a CA in
 	// the system-wide trusted root list.
+	// Multiple file paths can be added by using `,` as a separator.
 	caCertificatesEnvVar = "LEGO_CA_CERTIFICATES"
+
+	// caSystemCertPool is the environment variable name that can be used to define
+	// if the certificates pool must use a copy of the system cert pool.
+	caSystemCertPool = "LEGO_CA_SYSTEM_CERT_POOL"
 
 	// caServerNameEnvVar is the environment variable name that can be used to
 	// specify the CA server name that can be used to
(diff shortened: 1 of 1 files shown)


dmke (approved):
I would like to suggest `os.PathListSeparator` (`:` on POSIX, `;` on Windows) instead of a comma for path separation. It may not be intuitive, but it already has precedence (in `$PATH`).

I'd be OK, if we leave it as it, though.

jmcombs:
Is there any documentation links that can be shared that can help explain the requirements around `LEGO_CA_SYSTEM_CERT_POOL` and `LEGO_CA_CERTIFICATES`? I am trying to append a Root CA from a self-hosted ACME server (Step CA). I understand that `LEGO_CA_SYSTEM_CERT_POOL` needs to be set as `true` but would love to know the certificate requirements for `LEGO_CA_CERTIFICATES`. Is the exported CA Chain in PEM format acceptable or do I need to provide the `crt` and private `key` file?

dmke:
The file(s) `LEGO_CA_CERTIFICATES` point to must be each a series of one or more PEM-encoded certificates. 

Example: `LEGO_CA_CERTIFICATES=/path/to/chain.pem`, where `chain.pem` might look like:

```
-----BEGIN CERTIFICATE-----
Certificate A
-----END CERTIFICATE-----
-----BEGIN CERTIFICATE-----
Certificate B
-----END CERTIFICATE-----
-----BEGIN CERTIFICATE-----
etc.
-----END CERTIFICATE-----
```

If it is more convenient for you, you can point `LEGO_CA_CERTIFICATES` to multiple files: `LEGO_CA_CERTIFICATES=/path/to/root.pem:/path/to/intermediate1.pem:/path/to/intermediate2.pem` (use `;` instead of `:` to separate files on Windows).

In both cases the order should not matter, but when in doubt start at the root certificate.

PR #1593, opened 2022-02-17
Always return CSR as part of Resource

relvira:
Return the generated CSR in the certificate `Resource` after an `Obtain` or `ObtainForCSR`.

There are cases where users of this library will like to perform some additional operations with the generated CSR after an obtain was run, for example: submitting the CSR to an external tool that monitors certificate-transparency logs, in order to make sure the certificate issued by the library is legitimate.

I am not very familiar with this codebase so I might've missed something. I updated existing tests to make sure the CSR in Resource exists.

Please let me know if there's anything else I can do to speed up this work (or if there's something terribly wrong with my approach).

Thank you.

### DIFF
--- a/certificate/certificates.go
+++ b/certificate/certificates.go
@@ -273,6 +273,7 @@ func (c *Certifier) getForCSR(domains []string, order acme.ExtendedOrder, bundle
 		Domain:     commonName,
 		CertURL:    respOrder.Certificate,
 		PrivateKey: privateKeyPem,
+		CSR:        csr,
 	}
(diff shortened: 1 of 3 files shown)


relvira:
thanks @ldez for your comment, I will definitely give that a try and report back. Wouldn't it make sense to have the CSR returned as part of the certificates struct anyway? a CSR is always generated and it was slightly confusing to have the attribute available in the struct but it was empty after all.

Looking forward to hear your thoughts!

PR #1595, opened 2022-02-19
docs: fix typos for usage hostingde example

johanneskastl:


johanneskastl:
Please let me know if I should squash the commits or if there is anything else I should change. Would be happy to do so...

ldez (approved):
thanks

PR #1596, opened 2022-02-20
Fix panic in certcrypto.ParsePEMPrivateKey

csmith:
If ParsePEMPrivateKey is called with input that is not actually
a valid PEM block (e.g. because a file got corrupted), pem.Decode
will return a nil block which is immediately dereferenced causing
a panic.

Guard against this case returning an error, and add some tests
to validate.


ldez (approved):
Thank :+1:

PR #1607, opened 2022-03-19
Update zz_gen_cloudflare.md

YIN-Renlong:
Update the execution detail of the command `CLOUDFLARE_DNS_API_TOKEN=...` to avoid misuse. Since `CLOUDFLARE_DNS_API_TOKEN=...` is not a conventional command line, its executability needs to be noted.

### DIFF
--- a/docs/content/dns/zz_gen_cloudflare.md
+++ b/docs/content/dns/zz_gen_cloudflare.md
@@ -27,8 +27,7 @@ lego --email myemail@example.com --dns cloudflare --domains my.example.org run
 
 # or
 
-CLOUDFLARE_DNS_API_TOKEN=1234567890abcdefghijklmnopqrstuvwxyz \
-lego --email myemail@example.com --dns cloudflare --domains my.example.org run
+CLOUDFLARE_DNS_API_TOKEN=1234567890abcdefghijklmnopqrstuvwxyz lego --email myemail@example.com --dns cloudflare --domains my.example.org run
 ```
 
 


ldez:
Hello,

> Since CLOUDFLARE_DNS_API_TOKEN=... is not a conventional command line, its executability needs to be noted.

There is no convention involved in this section, it's the right syntax for a multiline command.

Maybe I don't understand your point, but I will decline your PR for now.

YIN-Renlong:
Hi, thanks for the reply. In fact, when I first execute this command line:

`CLOUDFLARE_DNS_API_TOKEN=1234567890abcdefghijklmnopqrstuvwxyz \`

then
`lego --email myemail@example.com --dns cloudflare --domains my.example.org run`

An error will occur:
`cloudflare: some credentials information are missing: CLOUDFLARE_EMAIL,CLOUDFLARE_API_KEY or some credentials information are missing: CLOUDFLARE_DNS_API_TOKEN,CLOUDFLARE_ZONE_API_TOKEN`

Only when I execute a complete line, the command line will execute successfully:

`CLOUDFLARE_DNS_API_TOKEN=1234567890abcdefghijklmnopqrstuvwxyz lego --email myemail@example.com --dns cloudflare --domains my.example.org run
`
For this reason, I believe it is necessary to simplify the usage of the command line, or at least need to clarify the scope of usage. What do you think?

ldez:
It's just standard shell/bash, nothing exotic here.

The 2 lines are here for several reasons:
- it's the right syntax for a multiline command
- you can peek only one line if you already exported your env vars
- the readability is better

Issue #1626, opened 2022-04-18
BlueCat provider cleanup error 


dittman1:
### Welcome

- [X] Yes, I'm using a binary release within 2 latest releases.
- [X] Yes, I've searched similar issues on GitHub and didn't find any.
- [X] Yes, I've included all information below (version, config, etc).

### What did you expect to see?

The TXT record removed after validation.

### What did you see instead?

There is an error when performing cleanup and the TXT record remains in BlueCat.

### How do you use lego?

Binary

### Reproduction steps

I run the command:

```
./lego --server="https://acme-staging-v02.api.letsencrypt.org/directory" --dns="bluecat" --domains="myhost.mydomain" --email="me@mydomain.net" run
```

### Version of lego

```console
lego version 4.6.0 linux/amd64
```

### Logs

```
2022/04/13 09:31:29 [WARN] [bam.mydomain] acme: cleaning up failed: bluecat: request failed with HTTP status code 404 Full message: {"status":404,"reason":"Not Found","code":"ResourceNotFound","message":"The requested resource was not found"}
```

ldez:
Hello,

I improved the Bluecat implementation, mainly because the error message was a bit unclear.

Are you able to test my PR #1627?

dittman1:
Let me see if I can build it.

dittman1:
I was unable to built it.  I don't normally use go so I installed the packages on RHEL 8.  Make fails because golangci-lint doesn't exist and I can't find it for RHEL 8.  I tried installing the suggested package buildah-tests but that doesn't provide it.

dittman1:
I think I was able to build it, but the test version doesn't even get the TXT record added it:

`# ./lego-test --server="https://acme-staging-v02.api.letsencrypt.org/directory" --dns="bluecat" --domains=“myhost.mydomain” --email=“me@mydomain” run
2022/04/18 11:37:36 [INFO] [myhost.mydomain] acme: Obtaining bundled SAN certificate
2022/04/18 11:37:36 [INFO] [myhost.mydomain] AuthURL: https://acme-staging-v02.api.letsencrypt.org/acme/authz-v3/2220594883
2022/04/18 11:37:36 [INFO] [myhost.mydomain] acme: Could not find solver for: tls-alpn-01
2022/04/18 11:37:36 [INFO] [myhost.mydomain] acme: Could not find solver for: http-01
2022/04/18 11:37:36 [INFO] [myhost.mydomain] acme: use dns-01 solver
[...]

PR #1627, opened 2022-04-18
bluecat: rewrite provider implementation

ldez:
- improve error messages
- split client and provider implementations
- improve status code handling
- add API documentation link
- add example


providers/dns/bluecat/internal/client.go
```diff
@@ -0,0 +1,286 @@
...
+		return err
+	}
+
+	authResp := string(authBytes)
+	if strings.Contains(authResp, "Authentication Error") {
+		return fmt.Errorf("request failed: %s", strings.Trim(authResp, "\""))
+	}
+
+	// Upon success, API responds with "Session Token-> BAMAuthToken: dQfuRMTUxNjc3MjcyNDg1ODppcGFybXM= <- for User : username"
+	c.token = regexp.MustCompile("BAMAuthToken: [^ ]+").FindString(authResp)
```
dmke, line 68: Maybe move `regexp.MustCompile` to a package-level variable?

providers/dns/bluecat/internal/client.go
```diff
@@ -0,0 +1,286 @@
...
+	}
+
+	authBytes, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return err
+	}
+
+	authResp := string(authBytes)
+	if !strings.Contains(authResp, "successfully") {
+		msg := strings.Trim(authResp, "\"")
```
dmke, line 99: I have somehow developed an aversion against backslashes :)

```suggestion
		msg := strings.Trim(authResp, `"`)
```

providers/dns/bluecat/internal/client.go
```diff
@@ -0,0 +1,286 @@
...
+		return nil, fmt.Errorf("status code: %d, message: %s", resp.StatusCode, string(data))
+	}
+
+	var txtRec EntityResponse
+	err = json.NewDecoder(resp.Body).Decode(&txtRec)
+	if err != nil {
+		return nil, fmt.Errorf("JSON decode: %w", err)
+	}
+
+	return &txtRec, err
```
dmke, line 185: For the sake of consistency, this should be:

```suggestion
	if err = json.NewDecoder(resp.Body).Decode(&txtRec); err != nil {
		return nil, fmt.Errorf("JSON decode: %w", err)
	}

	return &txtRec, nil
```

dmke (changes requested):
I've found some minor stylistic issues, but overall this looks very clean now.

dmke (approved):

dmke (approved):

dmke (approved):

providers/dns/bluecat/bluecat.go
```diff
@@ -60,10 +60,13 @@ type DNSProvider struct {
...
-// BLUECAT_SERVER_URL should have the scheme, hostname, and port (if required) of the authoritative Bluecat BAM server.
-// The REST endpoint will be appended.
-// In addition, the Configuration name and external DNS View Name must be passed in BLUECAT_CONFIG_NAME and BLUECAT_DNS_VIEW.
+// Credentials must be passed in the environment variables:
+//  - BLUECAT_SERVER_URL
+// 		It should have the scheme, hostname, and port (if required) of the authoritative Bluecat BAM server.
+// 		The REST endpoint will be appended.
+//  - BLUECAT_USER_NAME and BLUECAT_PASSWORD
+//  - BLUECAT_CONFIG_NAME (the Configuration name)
+//  - BLUECAT_DNS_VIEW (external DNS View Name)
```
dmke, line 69: Lists indented by space is rendered on pkg.go.dev in a single line (this [SO post](https://stackoverflow.com/q/52744468/766783) still applies).

One workaround is to use tabs (this will render a code block, so semantically, thats not ideal):

```suggestion
//	- BLUECAT_SERVER_URL
//	  It should have the scheme, hostname, and port (if required) of the authoritative Bluecat BAM server.
//	  The REST endpoint will be appended.
//	- BLUECAT_USER_NAME and BLUECAT_PASSWORD
//	- BLUECAT_CONFIG_NAME (the Configuration name)
//	- BLUECAT_DNS_VIEW (external DNS View Name)
```

An alternative is to use manual bullets:

```suggestion
// • BLUECAT_SERVER_URL
//
//  It should have the scheme, hostname, and port (if required) of the authoritative Bluecat BAM server.
//  The REST endpoint will be appended.
//
// • BLUECAT_USER_NAME and BLUECAT_PASSWORD
//
// • BLUECAT_CONFIG_NAME (the Configuration name)
//
// • BLUECAT_DNS_VIEW (external DNS View Name)
```

dmke (commented):
This provider seems to resist a bit too much to changes, doesn't it?

Do you have access to their product and can throw a live test against it?

Anyhow, I've added a remark to the doc string on `NewDNSProvider`. Feel free to ignore that, as it (might) resolve itself with https://github.com/golang/go/issues/51082.

ldez:
> This provider seems to resist a bit too much to changes, doesn't it?

The previous implementation was not really clean, I wanted to be more clean and precise but the API documentation is weak.

> Do you have access to their product and can throw a live test against it?

I don't have access to their product, I follow the feedback from #1626.

dmke:
> but the API documentation is weak

Well said :)

> I don't have access to their product, I follow the feedback from https://github.com/go-acme/lego/issues/1626.

Ah, I see.

PR #1628, opened 2022-04-20
gcore: change dns api url

lipovmi:


ldez:
Hello,

Can you provide a bit more context? (who are you, a reference to the official site about this change, etc.)
Why do you use 2 Github accounts for this PR?

PR #1657, opened 2022-06-15
feat: sleep at renewal

### LINKED ISSUE #1656: small random sleep at renewal to avoid load spikes
### Welcome

### How do you use lego?

Other

### Detailed Description

Hiya! I'm an engineer at Let's Encrypt. Lately we've been having severe load spikes at 00:00 UTC every day. We've run into similar issues in the past, and historically the issue has been large numbers of users manually configuring their cron jobs to run renewals at exactly midnight every day. It gets worse - because many of these jobs fail due to overload conditions, the next day the load spike may be bigger!

In Certbot the problem is partly solved by packaging - all the major packaging for Certbot arranges for it to run at a random time throughout the day. However, there are still people who manually setup Certbot in crontab, and they often choose 00:00 UTC to run it. So, some years ago Certbot added some code:

https://github.com/certbot/certbot/issues/6391
https://github.com/certbot/certbot/issues/6596
https://github.com/certbot/certbot/pull/6599

When `certbot renew` is run non-interactively, it will sleep a random amount of time up to 8 minutes. This helps spread out the midnight load spike significantly.

Right now lego-cli is the top participant in the load spike. During the first 30 seconds after 00:00 UTC today, lego-cli accounted for 33k new-order requests, while the next biggest contributor accounted for only 5.6k new-order requests. For all requests (not just new-order) lego-cli accounted for 173k vs 19k for the next biggest contributor (60% of the total).

Considering all lego-cli traffic to Let's Encrypt, it's very spiky:

image

So, two questions:

 - Could you implement a randomized delay for non-interactive renewals, similar to Certbot's?
 - Are you aware of any major integrations or packaging for lego-cli that includes a cron job or systemd unit that runs at 00:00?
 
 Thanks,
 Jacob

ldez:
Add a sleep between 0 to 8 minutes during the call of the `renew` command like [certbot](https://github.com/certbot/certbot/blob/284023a1b7672be2bd4018dd7623b3b92197d4b0/certbot/certbot/_internal/renewal.py#L472).

Fixes #1656

### DIFF
--- a/cmd/cmd_renew.go
+++ b/cmd/cmd_renew.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"crypto"
 	"crypto/x509"
+	"math/rand"
 	"time"
 
 	"github.com/go-acme/lego/v4/certcrypto"
@@ -68,6 +69,10 @@ func createRenew() *cli.Command {
 				Name:  "always-deactivate-authorizations",
 				Usage: "Force the authorizations to be relinquished even if the certificate request was successful.",
 			},
+			&cli.BoolFlag{
+				Name:  "no-random-sleep",
+				Usage: "Do not add a random sleep before the renewal. We do not recommend using this flag if you are doing your renewals in an automated way.",
+			},
 		},
 	}
 }
@@ -130,6 +135,14 @@ func renewForDomains(ctx *cli.Context, client *lego.Client, certsStorage *Certif
 		if errR != nil {
 			return errR
 		}
+	}
+
+	if ctx.Bool("no-random-sleep") {
+		// https://github.com/go-acme/lego/issues/1656
+		// https://github.com/certbot/certbot/blob/284023a1b7672be2bd4018dd7623b3b92197d4b0/certbot/certbot/_internal/renewal.py#L472
+		sleepTime := rand.New(rand.NewSource(time.Now().UnixNano())).Int63n(int64(8 * time.Minute))
+		log.Infof("renewal: random delay of %s", time.Duration(sleepTime))
+		time.Sleep(time.Duration(sleepTime) * time.Second)
 	}
 
 	request := certificate.ObtainRequest{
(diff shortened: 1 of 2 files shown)


cmd/cmd_renew.go
```diff
@@ -132,6 +137,14 @@ func renewForDomains(ctx *cli.Context, client *lego.Client, certsStorage *Certif
 		}
 	}
 
+	if ctx.Bool("no-random-sleep") {
+		// https://github.com/go-acme/lego/issues/1656
+		// https://github.com/certbot/certbot/blob/284023a1b7672be2bd4018dd7623b3b92197d4b0/certbot/certbot/_internal/renewal.py#L472
+		sleepTime := rand.New(rand.NewSource(time.Now().UnixNano())).Int63n(int64(8 * time.Minute))
+		log.Infof("renewal: random delay of %s", time.Duration(sleepTime))
+		time.Sleep(time.Duration(sleepTime) * time.Second)
```
dmke, line 145: Here happens a lot in the first line. Maybe rewrite this to something like:

```suggestion
		const jitter = 8 * time.Minute
		rndSource := rand.NewSource(time.Now().UnixNano())
		rnd := rand.New(rndSource)
		sleepTime := time.Duration(rnd.Int63n(int64(jitter)))
		
		log.Infof("renewal: random delay of %s", sleepTime)
		time.Sleep(time.Duration(sleepTime))
```

In any case, in `time.Sleep(time.Duration(sleepTime) * time.Second)` the multiplication by 10<sup>9</sup> (`time.Second`) must be removed, otherwise `time.Sleep` might pause the execution for a few hundred years.

dmke (changes requested):
I've found some defects :)

cmd/cmd_renew.go
ldez: yes I iterated and forget to remove the multiplication

dmke (approved):

ldez:
@dmke you can open a PR.

For the problem with go1.18 we already use an updated version of `golang.org/x/sys`.

PR #1659, opened 2022-06-17
cli/renew: skip random sleep if stdout is a terminal

dmke:
This aligns our behaviour with that of certbot.

- Issue: #1656
- Amends: #1657


ldez (approved):
LGTM

PR #1660, opened 2022-06-21
docs: new logo

ldez:
- new colorful logo
- drop references to ACME v1


dmke (changes requested):
I like the new logo.

However we should really use SVG images: [browser support is ubiquitous](https://caniuse.com/svg), and Github even allows selecting different images for dark and light themes. See

- https://github.blog/changelog/2021-11-24-specify-theme-context-for-images-in-markdown/
- https://driesvints.com/blog/investigating-dark-mode-for-svgs-in-github-readmes

(If you can send me the source file(s), I can help out here).

dmke:
I just noticed that the README embeds the colorful logo, which should be fine in either light or dark mode, so the point on selecting the right image depending on the theme is moot.

I'd still like to use SVGs, though, as they have another advantage: They render crisp with any device resolution. Bitmap graphics either tend to get blurry on HiDPI screens, and HiDPI images for low-density screens usually are unnecessarily large.

ldez:
I know what is SVG and how it works :wink: 
I use a 4K screen.
And I also know what the problems of SVG in a browser are.

> I just noticed that the README embeds the colorful logo, which should be fine in either light or dark mode, so the point on selecting the right image depending on the theme is moot.

I created the logo with colors that work with both modes, it is intentional.

FYI, I worked as a designer for 2 years.

ldez:
> Is the green fingerprint in the white logo also intentional?

Yes it's not a random green, it's the "ACME green" https://github.com/ietf-wg-acme/logos
When you use a logo from a "brand", it's better to follow their colors.

I will change the blue.

dmke:
> When you use a logo from a "brand" it's better to follow their colors.

When using brand logos, it is often frowned upon if you change the appearance (here: just taking the fingerprint, but omitting the hexagon).

ldez:
I changed the white logo and added SVG.

ldez:
The white logo must be a variation of the colored logo.
I already use SVGO (the SVGs are not a source SVG).

dmke:
> The white logo must be a variation of the colored logo.

As I said, I haven't touched the colorful logo *yet*. Adding color to the `logo-white.svg` should however be as simple as injecting a stylesheet:

```html
<style>
/* TODO: lookup the correct color values */
.lego-letter-l, .lego-letter-e { fill: yellow }
.lego-letter-g, .lego-letter-o { fill: blue }
.acme-logo-hex { fill: green; stroke-width: 0 }
</style>
```

> I already use SVGO

Sadly, SVGO does not apply transformations, nor remove unused gradients, nor clear unused ID and class attributes.

> I don't want that the ACME logo takes the main place: the lock is the main piece.

I see. I'll try to improve the it further (tomorrow, that is).

ldez:
> As I said, I haven't touched the colorful logo yet.

I know this is why I removed my comment.

I want to keep my source files clean, so I don't want to play with CSS.

I will take one of your ideas: reduced stroke width of "circuits"

dmke (approved):
I've shaved another few hundred bytes off the already minified files.

ldez:
Thank you for your suggestions and feedback :heart:

PR #1670, opened 2022-07-01
callback dns provider

attiss:
This PR contains a new DNS provider implementation: `callback`.

---

`callback` provider can be used for custom DNS solutions when lego is used as a library. The provider accepts two callback functions:
- one that gets called by `Present()` - the function should take care of registering the TXT records with the provided contents
- one that gets called by `CleanUp()` - the function should take care of deregistering the TXT records with the provided contents

Real world use case:
- in-house DNS solutions (without publicly available APIs)

### DIFF
--- a/providers/dns/callback/callback.go
+++ b/providers/dns/callback/callback.go
@@ -1,0 +1,77 @@
+// Package callback implements a DNS provider for solving the DNS-01 challenge using callback functions.
+package callback
+
+import (
+	"errors"
+	"time"
+
+	"github.com/go-acme/lego/v4/challenge/dns01"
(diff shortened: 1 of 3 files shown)


ldez:
Hello,

You can already use the following providers: `[exec](https://go-acme.github.io/lego/dns/exec/)`, or `[httpreq](https://go-acme.github.io/lego/dns/httpreq/)`

If you are using lego as a lib you can provide your own provider implementation, you don't need "callback".

I will decline your PR.

attiss:
Hi @ldez,

Fair point, it's much more straight forward to implement my custom provider by satisfying that interface.

Thank you for the quick feedback!

PR #1673, opened 2022-07-12
gcloud: update golang.org/x/oauth2

lucacasonato:
This updates x/oauth2/google. The new version supports
"external_account" JSON key files, which are needed when using workload
identity federation on GCP.


ldez (approved):
thanks :+1:

PR #1677, opened 2022-07-16
add recursive CNAME lookup support

danthegoodman1:
Right now the `LEGO_EXPERIMENTAL_CNAME_SUPPORT` only follows one CNAME record. If this CNAME record points to another, then the chain breaks and the DNS record created is incorrect.

This change allows it to keep following CNAME records as long as it continues to find one up to a limit so that CNAME loops don't hang the service.

And thanks for the awesome work on this repo!!


ldez (approved):
Thanks :+1:

PR #1679, opened 2022-07-17
pdns: notify secondary servers after updates

esclear:
This change addresses the PowerDNS DNS provider.

When using the API with a setup where there is not only the primary but also some secondary servers, the secondary servers were not sent a `NOTIFY` and would not initiate an AXFR transfer.

Thus, the challenge would not propagate to all DNS servers which would in return cause failed verifications.

This change uses the [`/notify`](https://doc.powerdns.com/authoritative/http-api/zone.html#put--servers-server_id-zones-zone_id-notify) endpoint of the PowerDNS API to notify the secondaries.

### DIFF
--- a/providers/dns/pdns/client.go
+++ b/providers/dns/pdns/client.go
@@ -137,6 +137,17 @@ func (d *DNSProvider) getAPIVersion() (int, error) {
 	return latestVersion, err
 }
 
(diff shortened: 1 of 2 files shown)


ldez (approved):
Thanks :+1:

PR #1687, opened 2022-08-02
digitalocean: configurable base URL

rienafairefr:
I wanted to proxy the calls to DigitalOcean API made for dns ACME challenge, and the go-acme/lego `digitalocean` provider has hardcoded https://api.digitalocean.com, this PR should handle that hopefully.

### DIFF
--- a/cmd/zz_gen_cmd_dnshelp.go
+++ b/cmd/zz_gen_cmd_dnshelp.go
@@ -561,7 +561,7 @@ func displayDNSHelp(name string) error {
 		ew.writeln()
 
 		ew.writeln(`Additional Configuration:`)
-		ew.writeln(`	- "DO_API_URL":	The URL of the API (default https://api.digitalocean.com)`)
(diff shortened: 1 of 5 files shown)


ldez:
Hello,

why do you this change?

ldez:
Hello,

why do you want to proxy the calls to DigitalOcean API made for DNS ACME challenge?
What kind of information did you expect?

rienafairefr:
I'm experimenting with limiting token disclosure risk by passing API calls through a passthrough proxy (similar to https://github.com/google/magic-github-proxy). This way I could put another token somewhere that would be allowed to do the necessary calls for an acme challenge on a certain domain, and not a iota more.

rienafairefr:
I could workaround using a more custom provider I think (like the httpreq), but I figured I could send the PR either way, some providers here have a customizable base_url.

ldez (approved):
LGTM

Issue #1693, opened 2022-08-09
Support for provider: yandex cloud


kluevandrew:
### Welcome

- [X] Yes, I've searched similar issues on GitHub and didn't find any.
- [X] Yes, the DNS provider exposes a public API.
- [X] Yes, I know that the lego maintainers don't have an account in all DNS providers in the world.
- [X] Yes, I'm able to create a pull request and be able to maintain the implementation.
- [X] Yes, I'm able to test an implementation if someone creates a pull request to add the support of this DNS provider.

### How do you use lego?

Through Traefik

### Link to the DNS provider

https://cloud.yandex.com

### Link to the API documentation

https://cloud.yandex.com/en/docs/dns/quickstart

### Additional Notes

Contains tested pull request

ldez:
Hello,

What is the difference between Yandex and Yandex Cloud?

https://go-acme.github.io/lego/dns/yandex/

kluevandrew:
> Hello,
> 
> What is the difference between Yandex and Yandex Cloud?
> 
> https://go-acme.github.io/lego/dns/yandex/

*Yandex Cloud* - is a cloud provider like Google Cloud Engine or Amazon AWS, 
*Yandex* - is a closed *pdd.yadex.ru*.

Project pdd.yandex.ru was closed in 2019, it was like Google Workspace but with DNS hosting. 
All domains was moved to *Yandex 360 for buisness* (new project like Google Workspace)
*Yandex 360 for business* also contains DNS Hosting, but it does not have public api as i know 

I'm not an employee of Yandex, i works at VK (funny, but we are competitors), so i can mistake in year when project was closed and may mistake about yandex360 API

paskal:
FYI https://pdd.yandex.ru still works for users. I haven't used `go-acme/lego` and its a Yandex provider, but I recently reviewed and verified the work of such a provider in https://github.com/AnalogJ/lexicon.

PR #1715, opened 2022-09-17
feat: deprecate Yandex and DNSPod

ldez:
Yandex is replaced by Yandex Cloud.
DNSPod is replaced by Tencent Cloud.

https://github.com/go-acme/lego/issues/1693#issuecomment-1209638944


dmke (approved):

paskal:
pdd.yandex.ru still works as well as their API, and I don't think there is a reason to name it deprecated while it works.

PR #1718, opened 2022-09-17
feat: take out CNAME support from experimental features

ldez:
drop `LEGO_EXPERIMENTAL_CNAME_SUPPORT`


ldez:
Windows... a file system with "ephemeral" errors and also with a dumb DNS resolving system...

So I will try to understand and fix the problem but I think that I will be forced to keep the option only because Windows is a piece of s***.

ldez:
So I was forced to increase the DNS timeout only for Windows...

dmke (approved):

PR #1719, opened 2022-09-18
docs: un-deprecate Yandex

ldez:
- https://github.com/go-acme/lego/pull/1715#issuecomment-1250371197
- https://github.com/go-acme/lego/issues/1693#issuecomment-1250363604

cc @paskal

### DIFF
--- a/providers/dns/yandex/yandex.toml
+++ b/providers/dns/yandex/yandex.toml
@@ -1,7 +1,7 @@
 Name = "Yandex PDD"
 Description = '''
 '''
-URL = "https://yandex.com/"
+URL = "https://pdd.yandex.com"
 Code = "yandex"
 Since = "v3.7.0"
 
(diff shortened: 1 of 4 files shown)


paskal (approved):

paskal:
Only one thing: I would propose describing that provider explicitly as "Yandex PDD", even if the name stays just "yandex". I think just "yandex" is okay as there was no "yandex cloud DNS" before recently, but the description better be clear about it.

providers/dns/yandex/yandex.toml
```diff
@@ -1,4 +1,4 @@
-Name = "Yandex"
+Name = "Yandex PDD"
 Description = '''
 '''
 URL = "https://yandex.com/"
```
paskal, line 4: Sorry for not catching it last time. I propose to change it to `https://yandex.com/dev/domain/doc/reference/dns-add.html` or `https://pdd.yandex.com`: the current URL is the wrong one.

paskal (approved):

dmke (approved):

Issue #1649, opened 2022-05-22
support RFC8738


Greyh4t:
### Welcome

- [X] Yes, I've searched similar issues on GitHub and didn't find any.

### How do you use lego?

Binary

### Detailed Description

When I use ZeroSSL to create a certificate for an IP address, lego gives an error
````
2022/05/22 22:13:29 Could not obtain certificates:
	acme: error: 400 :: POST :: https://acme.zerossl.com/v2/DV90/newOrder :: urn:ietf:params:acme:error:rejectedIdentifier :: Invalid DNS identifier [<my ip address>]
````
here is my command
```
./lego --server https://acme.zerossl.com/v2/DV90 --eab --kid <kid> --hmac <hmac> --email=<email> --domains=<my ip address> --path ./ --accept-tos --http --http.port "0.0.0.0:80" run
```
I found that [RF8738](https://datatracker.ietf.org/doc/html/rfc8738)  defines IP Identifier Validation Extension，and lego seems not support it，can you add support for it?

ninetian:
the acme support IP SSL,but zerossl not allow request ip ssl on the acme API , so you got error

can try this https://github.com/tinkernels/zerossl-ip-cert
or login zerossl.com to get it

tested it's work

appliedprivacy:
Support for IP address based certificates will become relevant for all operators of encrypted DNS services due to the requirements in DDR
https://datatracker.ietf.org/doc/draft-ietf-add-ddr/ (section: 4.2 Verified Discovery).

orangepizza:
had to make new PR: visit #1838

PR #1720, opened 2022-09-20
support for ipaddress in SAN

orangepizza:
a tiny edits on few functions that adapt type of identifier/ SAN type if input 'domain' can be parsed as IP address.
not bothered to change name of variable []domains to some better name: kind don't want to follow all those function calls to pass full identifier 

test on pebble that this is enough for ipv4 and ipv6 in -d.

### DIFF
--- a/acme/api/internal/sender/useragent.go
+++ b/acme/api/internal/sender/useragent.go
@@ -10,5 +10,5 @@ const (
 	// ourUserAgentComment is part of the UA comment linked to the version status of this underlying library package.
 	// values: detach|release
 	// NOTE: Update this with each tagged release.
-	ourUserAgentComment = "release"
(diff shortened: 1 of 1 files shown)


ldez:
Related to #1649

orangepizza:
while I creating tls-alpn testing I found that current tls-alpn test creating wrong testing certificate (SAN DNSname included port number in it, moke verifier had same error too (not triming port number in domain fro localhost:23578 so test passed. this only happends only in test as problem was sending wrong name to auth - otherwise TLS-ALPN client would be broken

ldez:
Hello,

do you plan to recreate a PR based on your new branch? https://github.com/orangepizza/lego/tree/ipaddr

orangepizza:
yes I already did : look #1838

PR #1727, opened 2022-10-05
Route 53: Update IAM policy example to grant least privilege access

paulgalow:
[AWS recently released new IAM permission conditions](https://aws.amazon.com/about-aws/whats-new/2022/09/amazon-route-53-support-dns-resource-record-set-permissions/) to restrict access to Route 53 beyond the scope of a hosted zone down to individual records. This makes it possible to finally create least privilege IAM policies for restricting access to only the [specific DNS TXT record](https://letsencrypt.org/docs/challenge-types/#dns-01-challenge) required. I have [written a blog post](https://paulgalow.com/aws-route-53-iam-policy-letsencrypt-dns) with more context and examples.

I'd like to propose an updated IAM policy which takes into account the new capabilities to lock down access as much as possible. I have successfully tested this policy with Traefik.

One downside I see with the example at hand is that users will now need to provide more information in their IAM policies, i.e. their hosted zone ID and their domain name.

Curious to hear your thoughts.

### DIFF
--- a/docs/content/dns/zz_gen_route53.md
+++ b/docs/content/dns/zz_gen_route53.md
@@ -82,32 +82,51 @@ See also:
 
 ## Policy
 
-The following AWS IAM policy document describes the permissions required for lego to complete the DNS challenge.
+The following AWS IAM policy document describes least privilege permissions required for lego to complete the DNS challenge. Replace `Z11111112222222333333` with your hosted zone ID and `example.com` with your domain name.
 
 ```json
 {
-   "Version": "2012-10-17",
-   "Statement": [
-       {
-           "Sid": "",
-           "Effect": "Allow",
-           "Action": [
-               "route53:GetChange",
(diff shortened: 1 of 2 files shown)


ldez:
Hello,

Your PR is interesting, can you keep the previous snippet and add your snippet in another code section because I think the 2 approaches can be used.

ldez:
@paulgalow the file `docs/content/dns/zz_gen_route53.md` is a generated file.

https://github.com/go-acme/lego/blob/7cfa075581991478504d4716ab9eda1e7c4c0653/docs/content/dns/zz_gen_route53.md?plain=1#L12-L14

You have to move your changes to `providers/dns/route53/route53.toml` and run `make generate-dns`.

paulgalow:
Thanks for pointing me into the right direction, @ldez. Now that you've mentioned it, it is totally obvious to me that I was editing the wrong file. 🤦‍♂️

PR #1733, opened 2022-10-10
cname: stop trying to traverse cname if none have been found

jbrunink:
It would try to resolve CNAME's over and over even though there was an empty answer. This resolves that. Feedback is appreciated. :-) Thank you!


jbrunink:
Hold on, I need to LRN2GIT. :/

ldez:
Hello,

You spotted something interesting but the implementation wasn't the best so I kept the idea but simplified the implementation.

ldez (approved):
LGTM

PR #1734, opened 2022-10-10
hurricane: fix CNAME support

jbrunink:
I've made a start with fixing up the Hurricane Electric DNS provider. It will now resolve the CNAME with `dns01.GetRecord` and then use the result as `hostname` to the DynDNS API of Hurricane Electric. To my understanding this is how it should work.

Feedback is appreciated and thank you. 😄

### DIFF
--- a/providers/dns/hurricane/hurricane.go
+++ b/providers/dns/hurricane/hurricane.go
@@ -86,9 +86,9 @@ func NewDNSProviderConfig(config *Config) (*DNSProvider, error) {
 
 // Present updates a TXT record to fulfill the dns-01 challenge.
 func (d *DNSProvider) Present(domain, _, keyAuth string) error {
-	resolvedDomain, txtRecord := dns01.GetRecord(domain, keyAuth)
+	fqdn, txtRecord := dns01.GetRecord(domain, keyAuth)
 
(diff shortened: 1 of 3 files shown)


ldez:
@jbrunink what is the problem? can you give me an explanation?

ldez:
The context:

For example, the domain `example.org` is owned by Cloudflare with the following configuration
```
example.org
CNAME _acme-challenge.example.duckdns.org
```

And the domain `example.duckdns.org` is owned by DuckDNS.

The goal is to create a TXT record inside duckdns.org:

```
example.duckdns.org
TXT YSBmYWtlIHZhbHVlIGZvciB0aGUgVFhUIHJlY29yZA==
```

---

In my implementation, the `hostname` is based on the FQDN extracted from the CNAME (`_acme-challenge.example.duckdns.org.`), so the `domain` will be `example.duckdns.org` and the `hostname` will be `_acme-challenge.example.duckdns.org`.
The token will be related to the  `domain` (`example.duckdns.org`) from duckdns.org.

ldez:
@jbrunink is it more clear with my explanation?

ldez (approved):
Thank you

PR #1747, opened 2022-10-31
lib: close response body

iwind:
with http.Client, even call url with HEAD method, resp.Body is not null and open, if we did not close it, next requests will fail in some conditions.


ldez (approved):

ldez:
The body cannot be close like that because the body is use later.

PR #1763, opened 2022-11-25
Option to entirely disable DNS01 propagation check

lnobach:
Here is a challenge option to entirely skip the propagation check after the DNS01 challenge has been presented by the DNS provider. This option should only be used by DNS providers guaranteeing that the challenge is present at their authoritative
nameservers after returning from the Present(..) function.

## Motivation

Needed by (almost) air-gapped platforms running lego which
- are not able to reach the authoritative nameserves of the domains (timeout or no route to host),
- are not able to check the TXT record over their resolver (propagation too slow despite low TTL),
- instead have another provider-specific method ensuring propagation.
 
## Example usage

```go
myprovider = GetMyProviderEnsuringPropagation()
err = client.Challenge.SetDNS01Provider(myprovider, dns01.DisablePropagationCheck())
if err != nil {
	return err
}
```

### DIFF
--- a/challenge/dns01/dns_challenge.go
+++ b/challenge/dns01/dns_challenge.go
@@ -114,29 +114,34 @@ func (c *Challenge) Solve(authz acme.Authorization) error {
 		return err
 	}
 
-	fqdn, value := GetRecord(authz.Identifier.Value, keyAuth)
+	if c.preCheck.disabled {
+		log.Infof("[%s] acme: DNS record propagation check disabled.", domain)
+	} else {
+		fqdn, value := GetRecord(authz.Identifier.Value, keyAuth)
 
-	var timeout, interval time.Duration
-	switch provider := c.provider.(type) {
-	case challenge.ProviderTimeout:
-		timeout, interval = provider.Timeout()
-	default:
-		timeout, interval = DefaultPropagationTimeout, DefaultPollingInterval
-	}
-
-	log.Infof("[%s] acme: Checking DNS record propagation using %+v", domain, recursiveNameservers)
-
-	time.Sleep(interval)
-
-	err = wait.For("propagation", timeout, interval, func() (bool, error) {
-		stop, errP := c.preCheck.call(domain, fqdn, value)
-		if !stop || errP != nil {
-			log.Infof("[%s] acme: Waiting for DNS record propagation.", domain)
+		var timeout, interval time.Duration
+		switch provider := c.provider.(type) {
+		case challenge.ProviderTimeout:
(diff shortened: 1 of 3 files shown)


ldez:
Hello,

You can already disable the propagation check by using `dns01.WrapPreCheck()`

Example: https://github.com/traefik/traefik/blob/a3e4c85ec0bbc57957473692d9006f7eab186175/pkg/provider/acme/provider.go#L304-L315

lnobach:
Hi @ldez,
thanks for the quick response.

I probably misunderstood the various ways in which you can use the WrapPreCheck function. I will try out if it does a similar job and close the PR if it does.

lnobach:
Hi,

WrapPreCheck replaces an individual propagation check step with a custom function, but not the entire pre-check process.
However by just creating a function immediately returning `true, nil`, the check process can be successfully stopped with the first attempt:

```go
myprovider = GetMyProviderEnsuringPropagation()
err = client.Challenge.SetDNS01Provider(myprovider,
	dns01.WrapPreCheck(func(domain, fqdn, value string, check dns01.PreCheckFunc) (bool, error) {
		log.Debug("Skipping lego's DNS01 propagation check (ignore the previous two log lines).")
		return true, nil
	}))
//...
```

One thing that is left is the initial sleep before the loop kicks in, taking unnecessary time:

https://github.com/go-acme/lego/blob/1b56aa0d2f9b14d35814765bbd16b0bc0c08818e/challenge/dns01/dns_challenge.go#L129

This can be avoided with the ProviderTimeout interface:

```go
func (p *MyProviderEnsuringPropagation) Timeout() (timeout, interval time.Duration) {
	return 1, 0
}
```
The other thing that is left is some noise in the logs telling that DNS record propagation check is done using specific nameservers, which is not.

https://github.com/go-acme/lego/blob/1b56aa0d2f9b14d35814765bbd16b0bc0c08818e/challenge/dns01/dns_challenge.go#L127

```
2022/11/25 12:48:37 [INFO] [foo-bar.example.com] acme: Checking DNS record propagation using [127.0.0.53:53]
2022/11/25 12:48:37 [INFO] Wait for propagation [timeout: 1ns, interval: 0s]
DEBU[0052] Skipping lego's DNS01 propagation check (ignore the previous two log lines).  module=ca-acme
```

I see the solution with `wrapPreCheck(..)` more as a workaround, but it is suitable for our purpose, so I wouldn't mind if you close the PR without merge. Feel free to decide.

ldez:
For now, I will decline the PR, thank you anyway.

PR #1766, opened 2022-11-27
fix: return an error when extracting record name

ldez:
return an error when extracting record name instead of panic

### DIFF
--- a/providers/dns/constellix/constellix.go
+++ b/providers/dns/constellix/constellix.go
@@ -109,7 +109,7 @@ func (d *DNSProvider) Present(domain, token, keyAuth string) error {
 		return fmt.Errorf("constellix: failed to get domain (%s): %w", authZone, err)
 	}
 
-	recordName, err := getRecordName(fqdn, authZone)
+	recordName, err := dns01.ExtractSubDomain(fqdn, authZone)
 	if err != nil {
 		return fmt.Errorf("constellix: %w", err)
 	}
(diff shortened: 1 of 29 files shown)


providers/dns/constellix/constellix.go
```diff
@@ -263,6 +269,12 @@ func containsValue(record *internal.Record, value string) bool {
...
-func getRecordName(fqdn, authZone string) string {
-	return fqdn[0 : len(fqdn)-len(authZone)-1]
+func getRecordName(fqdn, authZone string) (string, error) {
+	end := len(fqdn) - len(authZone) - 1
+
+	if len(fqdn) < end || end < 0 {
+		return "", fmt.Errorf("%d is lower than the length of the fqdn (fqdn: %s, authZone: %s)", end, fqdn, authZone)
+	}
+
+	return fqdn[0:end], nil
```
dmke, line 279: This function is repeated four times. Can we move it into an utilities package? `dns/providers/internal` already exists, so having a "records" sub package there wouldn't hurt that much.

dmke (changes requested):

providers/dns/constellix/constellix.go
ldez: The name is generic, we have several different implementations, so I don't want to create confusion.

dmke: I didn't know we had a "generic" package. For me, the important bit is that there are four copy-pasted implementations of non-trivial behaviour without tests.

dmke: Q: Is it possible to have `fqdn == "foo.example.com."` and `authZone == "example.org"`? 

If so, this function returns `"foo", nil`, and I'm not sure if this is another bug lingering here.

dmke: Ah, I believe I can answer that myself: `authZone` is the domain containing the SOA record for `fqdn`. IIRC, this requires `fqdn` to be a sub-domain of `authZone`.

ldez: I'm not a huge fan of "utils" package. I have to think.

ldez: I created a function `ExtractSubDomain` inside the `dns01` package.

dmke (approved):
LGTM

PR #1778, opened 2022-12-04
fix: return an error when extracting record name

ldez:
Same as #1766


providers/dns/hosttech/hosttech.go
```diff
@@ -114,9 +114,14 @@ func (d *DNSProvider) Present(domain, token, keyAuth string) error {
 		return fmt.Errorf("hosttech: could not find zone for domain %q (%s): %w", domain, authZone, err)
 	}
 
+	subDomain, err := dns01.ExtractSubDomain(fqdn, freemyip.RootDomain)
```
dmke, line 117: Is `freemyip.RootDomain` really correct here?

dmke (approved):

providers/dns/hosttech/hosttech.go
ldez: I was a copy-paste error

PR #1779, opened 2022-12-05
ibmcloud: add support for subdomains

thomasmhofmann:
I noticed that requesting wildcard certificates for subdomains does not work for the ibmcloud (softlayer) dns solver.
The problem is that the softlayer API getDomainID is used and the domain name is passed as an argument.
For wildcard certificates for subdomains this argument will be the subdomain. The subdomain is not a separate entity in softlayer DNS so no record and thus no domain ID is found.

This change will try to find the "parent" of the subdomain in case the domain that was used to find the domain ID has more than two parts.
When the parent can be found the correct acme challenge entries will be made and a wildcard certificate for a subdomain will be created successfully.

I did live tests with one of my domains in softlayer and with the change I am able to retrieve the wildcard cert for the subdomain(s). I tried:
- `*.toplevel.domain`
- `*.1.toplevel.domain`
- `*.1.2.toplevel.domain`
- `toplevel.domain`
- `2.toplevel.domain`
- `2.3.toplevel.domain`

I am a golang beginner so in case something is wrong with the code please tell me. 
Thanks.

### DIFF
--- a/providers/dns/ibmcloud/internal/wrapper.go
+++ b/providers/dns/ibmcloud/internal/wrapper.go
@@ -79,17 +79,6 @@ func getDomainID(service services.Dns_Domain, domain string) (*int, error) {
 	return getDomainID(service, parent)
 }
 
-func getParentDomain(domainParts []string) string {
-	numParts := len(domainParts)
-	var sb strings.Builder
-	for i := 1; i < numParts; i++ {
-		sb.WriteString(domainParts[i])
-		if i < numParts-1 {
-			sb.WriteString(".")
-		}
-	}
-	return sb.String()
-}
 
 func findTxtRecords(service services.Dns_Domain, fqdn string) ([]datatypes.Dns_Domain_ResourceRecord, error) {
 	var results []datatypes.Dns_Domain_ResourceRecord
--- a/providers/dns/ibmcloud/internal/wrapper.go
+++ b/providers/dns/ibmcloud/internal/wrapper.go
@@ -72,13 +72,12 @@ func getDomainID(service services.Dns_Domain, domain string) (*int, error) {
 	// So in case a subdomain like `sub.toplevel.tld` was used try again using the parent domain
 	// (strip the first part in the domain string -> `toplevel.tld`).
 	_, parent, found := strings.Cut(domain, ".")
-	if !found || strings.Index(parent, ".") == -1 {
+	if !found || !strings.Contains(parent, ".") {
 		return nil, fmt.Errorf("no data found for domain: %s", domain)
 	}
 
 	return getDomainID(service, parent)
 }
-
 
 func findTxtRecords(service services.Dns_Domain, fqdn string) ([]datatypes.Dns_Domain_ResourceRecord, error) {
 	var results []datatypes.Dns_Domain_ResourceRecord
--- a/providers/dns/ibmcloud/internal/wrapper.go
+++ b/providers/dns/ibmcloud/internal/wrapper.go
@@ -40,7 +40,7 @@ func (w Wrapper) CleanupTXTRecord(fqdn, domain string) error {
 
 	domainID, err := getDomainID(service, domain)
 	if err != nil {
-		return fmt.Errorf("failed to get domain ID for domain: %w", err)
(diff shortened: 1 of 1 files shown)


providers/dns/ibmcloud/internal/wrapper.go
```diff
@@ -66,7 +67,27 @@ func getDomainID(service services.Dns_Domain, domain string) (*int, error) {
...
+	numParts := len(domainParts)
+	var sb strings.Builder
+	for i := 1; i < numParts; i++ {
+		sb.WriteString(domainParts[i])
+		if i < numParts-1 {
+			sb.WriteString(".")
+		}
+	}
+	return sb.String()
 }
```
ldez, line 91:
```suggestion
func getParentDomain(domain string) (string, error) {
	domainParts := dns.SplitDomainName(domain)
	if len(domainParts) <= 2 {
		return "", fmt.Errorf("no data found for domain: %s", domain)
	}

	return strings.Join(domainParts[1:], "."), nil
}
```

providers/dns/ibmcloud/internal/wrapper.go
```diff
@@ -66,7 +67,27 @@ func getDomainID(service services.Dns_Domain, domain string) (*int, error) {
 		return r.Id, nil
 	}
 
-	return nil, fmt.Errorf("no data found of domain: %s", domain)
+	// The domain was not found by name. For subdomains this is not unusal in softlayer.
+	// So in case a subdomain like sub.toplevel.domain was used try again using the
+	// parent domain (strip the first part in the domain string -> toplevel.domain).
```
ldez, line 72:
```suggestion
	// The domain was not found by name.
	// For subdomains this is not unusual in softlayer.
	// So in case a subdomain like `sub.toplevel.tld` was used try again using the parent domain
	// (strip the first part in the domain string -> `toplevel.tld`).
```

providers/dns/ibmcloud/internal/wrapper.go
```diff
@@ -66,7 +67,27 @@ func getDomainID(service services.Dns_Domain, domain string) (*int, error) {
...
+	domainParts := strings.Split(domain, ".")
+	if len(domainParts) > 2 {
+		return getDomainID(service, getParentDomain(domainParts))
+	} else {
+		return nil, fmt.Errorf("no data found for domain: %s", domain)
+	}
```
ldez, line 78:
```suggestion
	parent, err := getParentDomain(domain)
	if err != nil {
		return nil, err
	}

	return getDomainID(service, parent)
```

providers/dns/ibmcloud/internal/wrapper.go
```diff
@@ -39,7 +40,7 @@ func (w Wrapper) CleanupTXTRecord(fqdn, domain string) error {
 
 	domainID, err := getDomainID(service, domain)
 	if err != nil {
-		return fmt.Errorf("failed to get domain ID: %w", err)
+		return fmt.Errorf("failed to get domain ID for domain: %w", err)
```
ldez, line 43:
```suggestion
		return fmt.Errorf("failed to get domain ID: %w", err)
```

ldez (changes requested):

providers/dns/ibmcloud/internal/wrapper.go
thomasmhofmann: My understanding is that Cut will return the before and after parts in regards to the sperator.
So if domain string is "toplevel.domain" before will be "toplevel" and after will be "domain" (found will be true).

So for this case the getDomainID function is called allthough it does not make sense to call it because "domain" itself will not yield a result. This is why the initial code had the len(domainParts) > 2 check.

This will probably not cause harm but a call to getDomainID that will always fail.

Thanks for looking into this.

ldez: FYI my algo has the same behavior as yours, but it is 10 times faster.

ldez (approved):
Thanks :+1:

thomasmhofmann:
Thanks for quick review and working with me on this!

PR #1785, opened 2022-12-12
docs: generate CLI help

dmke:
## Description

To prevent bit rot, this adds a bit of `go generate` code to build the `lego --help` output on the [CLI options page](https://go-acme.github.io/lego/usage/cli/options/) directly from the `*urfave/cli.App`.

Technically, `internal/cli_help/main.go` instantiates the CLI app similar to the way `cmd/lego/main.go` does it, but omits some details (like the version number or the CWD prefix to `--path`). It then simulates a few help commands (`lego help`, etc.) and dumps the results into `docs/data/zz_cli_help.toml`, from where Hugo embeds it into the `Options.md` via a new shortcode ("`{{< clihelp >}}`").

(I don't really like the shortcode detour, but Markdown content files are not allowed to access variables directly.)

The initial `go generate ./...` call (e.g. from `make generate-dns`) might now take a moment longer, because the initialization of the CLI app requires a few more CPU cycles.

## Other changes

- I've tried to reduce the line length of the `*cli.Flag` definitions (in `cmd/cmd_*.go`) to a reasonable limit. Feel free to omit 3249cd7f entirely if you feel this is a pointless endeavour :)
- In order to capture the output of `lego dnshelp`, a small modification of `displayDNSHelp()` (generated by `internal/dnsdocs`) was needed: It now accepts an `io.Writer`.
- `make validate-docs` now suppresses the output of the `git diff --exit-code` conditional. I found its output confusing.

## Preview

https://mm.dmke.org/lego/usage/cli/options/ (manually updated)


internal/dnsdocs/cli_help/generator.go
```diff
@@ -0,0 +1,96 @@
...
+	// - skip DNS help and provider list, as initialization takes time,
+	//   and we don't generate `lego dns --help` here
+	app := cli.NewApp()
+	app.Name = "lego"
+	app.HelpName = "lego"
+	app.Usage = "Let's Encrypt client written in Go"
+	app.Flags = cmd.CreateFlags("./.lego")
+	app.Commands = cmd.CreateCommands()
+	return app
+}()
```
dmke, line 80: :thinking: Maybe this code should live in `cmd/`, and reused by `cmd/lego/main.go`? Something like

<details><summary>cmd/cmd_app.go</summary>

```go
func InitApp(path string) *cli.App {
	app := cli.NewApp()
	app.Name = "lego"
	app.HelpName = "lego"
	app.Usage = "Let's Encrypt client written in Go"
	app.Flags = CreateFlags(path)
	app.Commands = CreateCommands()
	return app
}
```

</details>
<details><summary>cmd/lego/main.go</summary>

```go
func main() {
	var defaultPath string
	cwd, err := os.Getwd()
	if err == nil {
		defaultPath = filepath.Join(cwd, ".lego")
	}

	app := cmd.InitApp(defaultPath)

	app.Version = version
	cli.VersionPrinter = func(c *cli.Context) {
		fmt.Printf("lego version %s %s/%s\n", c.App.Version, runtime.GOOS, runtime.GOARCH)
	}
	app.EnableBashCompletion = true
	app.Before = cmd.Before

	err = app.Run(os.Args)
	if err != nil {
		log.Fatal(err)
	}
}
```

</details>
<details><summary>internal/dnsdocs/cli_help/generator.go</summary>

```go
var lego = cmd.InitApp("./.lego")
```

</details>

OTOH, how likely is it that the CLI app initialization changes so much, that the documentation becomes unusable? The most important part (printing the flags) would be very much unaffected by most changes to the `*cli.App`.

dmke: I think, I'll leave it as is. "[A little copying is better than a little dependency.](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=9m28s)"

ldez (approved):
LGTM

Issue #1777, opened 2022-12-01
Possible bad propagation check with dns-01 challenge


oseiberts11:
### Welcome

- [X] Yes, I'm using a binary release within 2 latest releases.
- [X] Yes, I've searched similar issues on GitHub and didn't find any.
- [X] Yes, I've included all information below (version, config, etc).

### What did you expect to see?

A sucessful certificate generation.

### What did you see instead?

We had an error message from Let's Encrypt:
`2022/11/30 10:24:59 error: one or more domains had a problem:
[cloud.syseleven.de] acme: error: 400 :: urn:ietf:params:acme:error:dns :: DNS problem: SERVFAIL looking up TXT for _acme-challenge.cloud.syseleven.de - the domain's nameservers may be malfunctioning",`
but the propagation check with 2 name servers apparently passed:
```
2022/11/30 10:23:40 [INFO] [cloud.syseleven.de] acme: Could not find solver for: tls-alpn-01
2022/11/30 10:23:40 [INFO] [cloud.syseleven.de] acme: Could not find solver for: http-01
2022/11/30 10:23:40 [INFO] [cloud.syseleven.de] acme: use dns-01 solver
2022/11/30 10:23:40 [INFO] [cloud.syseleven.de] acme: Preparing to solve DNS-01
2022/11/30 10:23:50 [INFO] [cloud.syseleven.de] acme: Trying to solve DNS-01
2022/11/30 10:24:00 [INFO] [cloud.syseleven.de] acme: Checking DNS record propagation using [8.8.8.8:53 4.4.4.4:53]
2022/11/30 10:24:10 [INFO] Wait for propagation [timeout: 10m0s, interval: 10s]
2022/11/30 10:24:10 [INFO] [cloud.syseleven.de] acme: Waiting for DNS record propagation.
2022/11/30 10:24:20 [INFO] [cloud.syseleven.de] acme: Waiting for DNS record propagation.
2022/11/30 10:24:30 [INFO] [cloud.syseleven.de] acme: Waiting for DNS record propagation.
2022/11/30 10:24:47 [INFO] [cloud.syseleven.de] acme: Cleaning DNS-01 challenge
```
With a successful result, before the last line, there would have been a message like 
`[cloud.syseleven.de] The server validated our request`.

Note that we used `4.4.4.4` which was a working public name server in the past, but apparently no longer. It must have stopped working relatively recently. We discovered this while trying to debug this. As of now, it does not respond to any query.

[...]

PR #1786, opened 2022-12-12
docs: expand description for --dns.resolvers

dmke:
This extends the documentation for the `--dns.resolvers` flag as suggested in #1777.

The commits are currently based off of PR #1785, so depending on whether that gets merged, I'll rebase this PR accordingly.

@oseiberts11: Would you like to proof-read the new section? You can find a preview here: https://mm.dmke.org/lego-b649657b/usage/cli/options/#dns-resolvers-and-challenge-verification

### DIFF
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -104,7 +104,8 @@ func CreateFlags(defaultPath string) []cli.Flag {
 		},
 		&cli.StringSliceFlag{
 			Name: "dns.resolvers",
-			Usage: "Set the resolvers to use for performing recursive DNS queries." +
+			Usage: "Set the resolvers to use for performing (recursive) CNAME resolving and apex domain determination." +
+				" For DNS challenge verification, the authoritative DNS server is queried directly." +
 				" Supported: host:port." +
 				" The default is to use the system resolvers, or Google's DNS resolvers if the system's cannot be determined.",
 		},
--- a/docs/content/usage/cli/Options.md
+++ b/docs/content/usage/cli/Options.md
@@ -54,16 +54,34 @@ When using a DNS challenge provider (via `--dns <name>`), Lego tries to ensure t
 
 This involves a few DNS queries to different servers:
 
-1. Determining the apex domain and resolving CNAMES.
+1. Determining the DNS zone and resolving CNAMEs.
 
-   The apex domain is usualy the 2nd level domain for TLDs like `.com` or `.de`, but can also be the 3rd level domain for `.co.uk` domains.
-   It contains the SOA record, from which Lego extracts the authoritative name server (MNAME, i.e. the primary name server).
+   The DNS zone for a given domain is determined by the SOA record, which contains the authoritative name server for the domain and all its subdomains.
+   For simple domains like `example.com`, this is usually `example.com` itself.
+   For other domains (like `fra.eu.cdn.example.com`), this can get complicated, as `cdn.example.com` may be delegated to the CDN provider, which means for `cdn.example.com` must exist a different SOA record.
+
+   To find the correct zone, Lego requests the SOA record for each DNS label (starting on the leaf domain, i.e. the left-most DNS label).
+   If there is no SOA record, Lego requests the SOA record of the parent label, then for its parent, etc., until it reaches the apex domain.[^apex]
    Should any DNS label on the way be a CNAME, it is resolved as per usual.
 
-   To resolve the apex domain and CNAMEs, Lego sends queries to the configured DNS resolvers.
-   These are, by default, the system name servers, and fallback to Google's DNS servers, should they be absent. You can override this behaviour with the `--dns.resolvers` flag.
+   In the default configuration, Lego uses the system name servers for this, and falls back to Google's DNS servers, should they be absent.
 
 2. Verifying the challenge token.
 
    The `_acme-challenge.<yourdomain>` TXT record must be correctly installed.
    Lego verifies this by directly querying the authoritative name server for this record (as detected in the previous step).
+
+Strictly speaking, this verification step is not necessary, but helps to protect your ACME account.
+Remember that some ACME providers impose a rate limit on certain actions (at the time of writing, Let's Encrypt allows 300 new certificate orders per account per 3 hours).
+
+There are also situations, where this verification step doesn't work as expected:
+
+- A "split DNS" setup gives different answers to clients on the internal network (Lego) vs. on the public internet (Let's Encrypt).
+- With "hidden master" setups, Lego may be able to directly talk to the primary DNS server, while the `_acme-challenge` record might not have fully propagate to the (public) secondary servers, yet.
+
+The effect is the same: Lego determined the challenge token to be installed correctly, while Let's Encrypt has a different view, and rejects the certificate order.
+
+In these cases, you can instruct Lego to use a different DNS resolver, using the `--dns.resolvers` flag.
+You should prefer one on the public internet, otherwise you might be susceptible to the same problem.
+
+[^apex]: The apex domain is the domain you have registered with your domain registrar. For gTLDs (`.com`, `.fyi`) this is the 2nd level domain, but for ccTLDs, this can either be the 2nd level (`.de`) or 3rd level domain (`.co.uk`).
--- a/docs/content/usage/cli/Options.md
+++ b/docs/content/usage/cli/Options.md
@@ -47,3 +47,23 @@ If you are using this option, make sure you proxy all of the following traffic t
 This traffic redirection is only needed as long as lego solves challenges. As soon as you have received your certificates you can deactivate the forwarding.
 
 [^header]: You must ensure that incoming validation requests contains the correct value for the HTTP `Host` header. If you operate lego behind a non-transparent reverse proxy (such as Apache or NGINX), you might need to alter the header field using `--http.proxy-header X-Forwarded-Host`.
+
+## DNS Resolvers and Challenge Verification
+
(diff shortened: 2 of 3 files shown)


oseiberts11:
@dmke Thanks! This looks great to me.

cmd/flags.go
```diff
@@ -102,8 +103,11 @@ func CreateFlags(defaultPath string) []cli.Flag {
 			Usage: "By setting this flag to true, disables the need to wait the propagation of the TXT record to all authoritative name servers.",
 		},
 		&cli.StringSliceFlag{
-			Name:  "dns.resolvers",
-			Usage: "Set the resolvers to use for performing recursive DNS queries. Supported: host:port. The default is to use the system resolvers, or Google's DNS resolvers if the system's cannot be determined.",
+			Name: "dns.resolvers",
+			Usage: "Set the resolvers to use for performing (recursive) CNAME resolving and apex domain determination." +
+				" For DNS challenge verification, the authoritative DNS server is queried directly." +
+				" Supported: host:port." +
+				" The default is to use the system resolvers, or Google's DNS resolvers if the system's cannot be determined.",
```
jsumners, line 110: Please also include an example. It's not clear how to provide multiple values for this setting. Do you provide a CSV or something else?

jsumners, line 110: I'm still not clear what this means. What does setting this value gain a user?

dmke: (I'll need to provide a bit of context, please bear with me for a moment. I'll try to be brief, but that also requires to simplify things. This is the ELI5 version.)

---

The ACME protocol involves a few steps of verification, in order to issue you a TLS certificate. One of the earliest steps is to ask the ACME provider (usually Let's Encrypt), if `lego` can have such a certificate for `www.example.com`, to which Let's Encrypt answers, "sure, if you can prove you own that domain; here is a token and place it someplace only accessible to owners".

Proving ownership can happen in different ways, but the bulk of the code in this repository is concerned with the "DNS challenge" (dns-01). This means, `lego` must place the aforementioned token as TXT record on `_acme-challenge.www.example.com`.

Now, DNS being the redundant and resilient system it is, is not known for fast propagation of changes. Layers of caching servers are setup between you and the name server responsible to deliver you that TXT record (on your PC, your router, your ISP have certainly installed a recursive resolver). Each layer may (or may not!) obey the caching time for certain DNS answers, even for negative answers ("I don't know about `_acme-challenge.www.example.com`, and the authoritative name server has told me to save this answer for 600 seconds").

This makes the verifying step tedious. `lego` could skip this step, and leave it to Let's Encrypt. But should setting up the TXT record fail (or just take a moment), Let's Encrypt may just respond with "you didn't setup the token, you don't get a certificate", and deduct one point from your request quota (remember that Let's Encrypt rate limits your account to 300 new orders per 3 hours). If you try and retry and retry to obtain a certificate, your quota may be exhausted quicker than you think.

The next less naïve approach would be to wait a moment. But that begs the question: for how long? Should every user generally suffer from overly long waiting periods, just because in some cases the propagation takes a bit longer? Certainly not.

So... what does `lego` do? Easy: We're querying the authoritative DNS server directly, avoiding any caches. The not-so-easy part is determining which server that is.

For a domain like `example.com`, we can just lookup the SOA record for `example.com`, and extract the MNAME from it (in DNS lingo, `example.com` is the apex domain for `example.com`, because it contains the SOA record). This however gets a little more complicated for domains like `zeus.fra.ger.eu.cdn.example.com`, where the whole `*.cdn.example.com` zone might be delegated to another server (the apex domain might be `cdn.example.com`, not `example.com`).

Here, `lego` can only ask first `zeus.fra.ger.eu.cdn.example.com` for its SOA record, then `fra.ger.eu.cdn.example.com`, etc., until it finds the authoritative name server.

The final step is to query that name server, "hey what's the TXT record for `_acme-challenge.zeus.fra.ger.eu.cdn.example.com`? If it doesn't provide the correct answer, `lego` will wait a moment (`--dns-timeout`) and try again (I forgot how often).

(The story then continues with a similar verification on Let's Encrypt's side, and the exchange of the TLS certificate, should LE be satisfied.)

---

What does setting [the `--dns.resolvers`] value gain a user?

Since `lego`, by default, uses the system's DNS resolver to determine the apex domain, this may lead to a wrong answer. Within data centers you might have DNS setup in a way which gives you a different answer than queries in from the outside world ("split DNS", "hidden master").

A false-positive here is not better than the first approach described above (i.e. let Let's Encrypt figure it out). A false-negative here could mean that your system resolver is not able at all to get the `_acme-challenge` record - and `lego` would wait and wait and finally fail.

Another reason can be that you don't trust your system's/router's/ISP's name server. I know from some residential ISPs here in Germany where in the past they wanted to be helpful and present users with a "Did you mean alice-dsl.de? Here are some ads"-page, if the end-user had a typo when entering a domain into the browsers URL bar (/rant). Of course such a behaviour breaks `lego`'s verification process as well.

In such situations, you need to instruct `lego` to use a specific (set of) known-good resolver(s) (ideally but not necessarily on the public internet). The `--dns.resolvers` flag is for this specific requirement.

dmke: If you run `lego --help` (or see [here](https://github.com/go-acme/lego/pull/1786/files#diff-fe0fb6c80906c79ff932549d18a30a1cff5846e1ac80a7d64a30baa965a0d74fR28), you'll find that you can just specify the `--dns.resolvers` flag multiple times:

```
$ lego --help

...

   --dns.resolvers value [ --dns.resolvers value ]
              Set the resolvers to use for performing (recursive) CNAME resolving and
              apex domain determination. For DNS challenge verification, the authoritative
              DNS server is queried directly. Supported: host:port. The default is to use
              the system resolvers, or Google's DNS resolvers if the system's cannot be
              determined.
```

(I've reformatted it a bit to cope with the restricted space imposed by Github.)

jsumners: Is that new? I have v4.8.0 (via https://github.com/void-linux/void-packages/blob/master/srcpkgs/lego/template) and I see:

```
> lego --help
**snip**
--dns.resolvers value        Set the resolvers to use for performing recursive DNS queries. Supported: host:port. The default is to use the system resolvers, or Google's DNS resolvers if the system's cannot be determined.
**snip**
```

Still, I think it would be good to state that the option can be repeated multiple times for multiple resolvers. This is the first time I can recall seeing this bracket syntax and it isn't super clear.

dmke: Hm, curious. I'll have to investigate...

dmke: Looks like we've updated the CLI argument parser library in Lego 4.9.0 from [2.3.0 to 2.14.0](https://github.com/urfave/cli/compare/v2.3.0...v2.14.0), which could explain differences in the output.

jsumners: 👍

docs/content/usage/cli/Options.md
```diff
@@ -131,3 +47,23 @@ If you are using this option, make sure you proxy all of the following traffic t
...
+   It contains the SOA record, from which Lego extracts the authoritative name server (MNAME, i.e. the primary name server).
+   Should any DNS label on the way be a CNAME, it is resolved as per usual.
+
+   To resolve the apex domain and CNAMEs, Lego sends queries to the configured DNS resolvers.
+   These are, by default, the system name servers, and fallback to Google's DNS servers, should they be absent. You can override this behaviour with the `--dns.resolvers` flag.
+
+2. Verifying the challenge token.
+
+   The `_acme-challenge.<yourdomain>` TXT record must be correctly installed.
+   Lego verifies this by directly querying the authoritative name server for this record (as detected in the previous step).
```
jsumners, line 69: image

dmke:
> @ldez marked this pull request as ready for review 2 minutes ago

Github doesn't let me approve my own PR, but the changes LGTM.

ldez:
I just rebased the PR, there are no changes :wink:

docs/content/usage/cli/Options.md
```diff
@@ -47,3 +47,23 @@ If you are using this option, make sure you proxy all of the following traffic t
...
 
 [^header]: You must ensure that incoming validation requests contains the correct value for the HTTP `Host` header. If you operate lego behind a non-transparent reverse proxy (such as Apache or NGINX), you might need to alter the header field using `--http.proxy-header X-Forwarded-Host`.
+
+## DNS Resolvers and Challenge Verification
+
+When using a DNS challenge provider (via `--dns <name>`), Lego tries to ensure the ACME challenge token is properly setup before instructing the ACME provider to perform the validation.
+
+This involves a few DNS queries to different servers:
+
+1. Determining the apex domain and resolving CNAMES.
```
ldez, line 57: The goal is to find the "zone", it's not really the apex, it can be the 3rd, 4th, 5th, 6th, etc. level

dmke:
> I just rebased the PR, there are no changes :wink:

Sorry, I've been looking at the changes of #1785.

docs/content/usage/cli/Options.md
dmke: Isn't the apex domain the domain containing the SOA record?

ldez: Depend on the definition of the apex.
We try to find the domain that handles the subdomain, it can be a subdomain.
An SOA record can be placed on a subdomain.
For me, the apex is the ltd + the "first domain", but maybe my own definition is wrong.

ldez: I hadn't seen your edit.

dmke: I've reworded the section and added a bit more content ([preview](https://mm.dmke.org/lego-b649657b/usage/cli/options/#dns-resolvers-and-challenge-verification)).

ldez (approved):
LGTM

PR #1788, opened 2022-12-18
tencentcloud: support punycode domain

huiyifyj:
Fixes https://github.com/go-acme/lego/discussions/1787

https://github.com/go-acme/lego/blob/6695fcc87344a920efcf0a504f299913a7869b54/providers/dns/tencentcloud/client.go#L20-L23
https://github.com/go-acme/lego/blob/6695fcc87344a920efcf0a504f299913a7869b54/providers/dns/tencentcloud/client.go#L39-L44

The tencent cloud sdk `DescribeDomainList` api.

The `Name` fields returned by `DescribeDomainList` api will not be converted into punycode format, can use `Punycode` field.


ldez (approved):
LGTM

PR #1797, opened 2023-01-04
update go-jose package to latest

drakkan:
go-jose v2 is not actively developed anymore

https://github.com/square/go-jose/tree/master#go-jose

jose v3 is also looking for help but at least it should be maintained


ldez (approved):
LGTM

PR #1802, opened 2023-01-07
Add -4 and -6 flags

### LINKED ISSUE #1801: Provide -4 and -6 flags to use IPv4 and IPv6 respectively
### Welcome

### How do you use lego?

Binary

### Detailed Description

This is related to several issues. A sampling of them are:

+ https://github.com/go-acme/lego/issues/1406
+ https://github.com/go-acme/lego/issues/1777

What I believe is happening is that lego is attempting to verify the challenge record propagation and *strongly* preferring IPv6 if it is enabled on the host. If the remote authoritative nameserver is not responding to an IPv6 connection, lego simply loops until the challenge token expires and thus never completes issuing a certificate.

I arrived at this conclusion by starting a certificate issuance, watching the requests (`tcpdump -i ${IFACE} udp port 53 | rg acme`) to see that outgoing queries for the TXT record are happening to the auth server over IPv6. I then tried the same query myself on both IPv4 and IPv6:

```sh
❯ dig @ns-249-c.gandi.net _acme-challenge.example.com txt +yaml
-
  type: MESSAGE
  message:
    type: AUTH_RESPONSE
    query_time: !!timestamp 2023-01-06T14:16:03.573Z
    response_time: !!timestamp 2023-01-06T14:16:03.612Z
    message_size: 109b
    socket_family: INET
    socket_protocol: UDP
    response_address: 217.70.187.250
    response_port: 53
    query_address: 0.0.0.0
    query_port: 43704
    response_message_data:
      opcode: QUERY
      status: NOERROR
      id: 27604
      flags: qr aa rd
      QUESTION: 1
      ANSWER: 1
      AUTHORITY: 0
      ADDITIONAL: 1
      OPT_PSEUDOSECTION:
        EDNS:
          version: 0
          flags:
          udp: 1232
      QUESTION_SECTION:
        - _acme-challenge.example.com. IN TXT
      ANSWER_SECTION:
        - _acme-challenge.example.com. 300 IN TXT "gpHXXn-bXxreJ860cysiinMfymNPPwPnq2aA3rNfdhY"
```

```sh
❯ dig @ns-249-c.gandi.net _acme-challenge.example.com txt +yaml -6

; <<>> DiG 9.16.22 <<>> @ns-249-c.gandi.net _acme-challenge.smurf.pw txt +yaml -6
; (1 server found)
;; global options: +cmd
-
  type: DIG_ERROR
  message: |
    connection timed out; no servers could be reached
```

Typically, a tool such as this (as shown by the `dig` usage above) provides a mechanism to indicate the IP stack to use through the switches `-4` and `-6`. If I understand https://pkg.go.dev/net#Dialer correctly, it looks like passing `tcp` or `tcp6` to the `network` parameter of `Dialer.Control` is all that is needed in the underlying framework to make these flags possible.

Note, the Gandi nameservers do respond on port 53 over IPv6. My actual problem is something with AT&T's terrible network.

jsumners:
If completed, this PR will resolve #1801. At the moment I'm opening a draft PR to gauge interest in the overall proposal.

<details><summary>(outdated lint errors)</summary>

I do not understand these lint errors:

```
cmd/flags.go:12: File is not `gci`-ed with --skip-generated -s standard,default (gci)
                        Name: "ipv4only",
cmd/flags.go:14: File is not `gci`-ed with --skip-generated -s standard,default (gci)
                        Usage: "Use IPv4 only. This flag is ignored if ipv6only is also specified.",
cmd/flags.go:17: File is not `gci`-ed with --skip-generated -s standard,default (gci)
                        Name: "ipv6only",
cmd/flags.go:19: File is not `gci`-ed with --skip-generated -s standard,default (gci)
                        Usage: "Use IPv6 only. This flag is ignored if ipv4only is also specified.",
```

I do not know what "gci-ed" means.

I also don't understand this one because `recursiveNameservers` is also a global:

```
challenge/dns01/nameserver.go:31:5: networkStack is a global variable (gochecknoglobals)
var networkStack = "both"
```

I'm open to this not being a global if I can get some guidance on how it should be passed in from the CLI flags processing down to where the DNS query is actually invoked.

</details>

Items that remain:
- [x] Update https://github.com/go-acme/lego/blob/e982d5a77fb5263978eaf7ea2db24b0cb5ce46cf/cmd/setup_challenges.go#L44
- [x] Update https://github.com/go-acme/lego/blob/e982d5a77fb5263978eaf7ea2db24b0cb5ce46cf/cmd/setup_challenges.go#L86

### DIFF
--- a/challenge/dns01/nameserver.go
+++ b/challenge/dns01/nameserver.go
@@ -26,6 +26,24 @@ var defaultNameservers = []string{
 // recursiveNameservers are used to pre-check DNS propagation.
 var recursiveNameservers = getNameservers(defaultResolvConf, defaultNameservers)
 
+// NetworkStack is used to indicate which IP stack should be used for DNS
+// queries. Valid values are DefaultNetworkStack, IPv4Only, and IPv6Only.
+type NetworkStack int
+
+const (
+	// DefaultNetworkStack indicates that both IPv4 and IPv6 should be allowed.
+	// This setting lets the OS determine which IP stack to use.
+	DefaultNetworkStack NetworkStack = iota
+	// IPv4Only forces DNS queries to only happen over the IPv4 stack.
+	IPv4Only
+	// IPv6Only forces DNS queries to only happen over the IPv6 stack.
+	IPv6Only
+)
+
+// currentNetworkStack is used to define which IP stack will be used. The default is
+// both IPv4 and IPv6. Set to IPv4Only or IPv6Only to select either version.
+var currentNetworkStack = DefaultNetworkStack
+
 // soaCacheEntry holds a cached SOA record (only selected fields).
 type soaCacheEntry struct {
 	zone      string    // zone apex (a domain name)
@@ -65,6 +83,11 @@ func AddRecursiveNameservers(nameservers []string) ChallengeOption {
 		recursiveNameservers = ParseNameservers(nameservers)
 		return nil
 	}
+}
+
+// SetNetworkStack defines the IP stack that will be used for DNS queries.
+func SetNetworkStack(network NetworkStack) {
+	currentNetworkStack = network
 }
 
 // getNameservers attempts to get systems nameservers before falling back to the defaults.
@@ -249,12 +272,33 @@ func createDNSMsg(fqdn string, rtype uint16, recursive bool) *dns.Msg {
 	return m
 }
 
+// getNetwork interprets the NetworkStack setting in relation to the desired
+// protocol. The proto value should be either "udp" or "tcp".
+func getNetwork(proto string) string {
+	// The dns client passes whatever value is set in [dns.Client.Net] to
+	// the [net.Dialer] (https://github.com/miekg/dns/blob/fe20d5d/client.go#L119-L141).
+	// And the [net.Dialer] accepts strings such as "udp4" or "tcp6"
+	// (https://cs.opensource.google/go/go/+/refs/tags/go1.18.9:src/net/dial.go;l=167-182).
+	if currentNetworkStack == IPv4Only {
+		return proto + "4"
+	}
+	if currentNetworkStack == IPv6Only {
+		return proto + "6"
+	}
+	return proto
+}
+
 func sendDNSQuery(m *dns.Msg, ns string) (*dns.Msg, error) {
-	udp := &dns.Client{Net: "udp", Timeout: dnsTimeout}
+	network := getNetwork("udp")
+	udp := &dns.Client{Net: network, Timeout: dnsTimeout}
 	in, _, err := udp.Exchange(m, ns)
 
-	if in != nil && in.Truncated {
-		tcp := &dns.Client{Net: "tcp", Timeout: dnsTimeout}
+	network = getNetwork("tcp")
+	// We can encounter a net.OpError if the nameserver is not listening
+	// on UDP at all, i.e. net.Dial could not make a connection.
+	_, isOpErr := err.(*net.OpError)
+	if (in != nil && in.Truncated) || isOpErr {
+		tcp := &dns.Client{Net: network, Timeout: dnsTimeout}
 		// If the TCP request succeeds, the err will reset to nil
 		in, _, err = tcp.Exchange(m, ns)
 	}
--- a/challenge/dns01/nameserver.go
+++ b/challenge/dns01/nameserver.go
@@ -26,24 +26,6 @@ var defaultNameservers = []string{
 // recursiveNameservers are used to pre-check DNS propagation.
 var recursiveNameservers = getNameservers(defaultResolvConf, defaultNameservers)
 
-// NetworkStack is used to indicate which IP stack should be used for DNS
-// queries. Valid values are DefaultNetworkStack, IPv4Only, and IPv6Only.
-type NetworkStack int
-
-const (
-	// DefaultNetworkStack indicates that both IPv4 and IPv6 should be allowed.
-	// This setting lets the OS determine which IP stack to use.
-	DefaultNetworkStack NetworkStack = iota
-	// IPv4Only forces DNS queries to only happen over the IPv4 stack.
-	IPv4Only
-	// IPv6Only forces DNS queries to only happen over the IPv6 stack.
-	IPv6Only
-)
-
-// currentNetworkStack is used to define which IP stack will be used. The default is
-// both IPv4 and IPv6. Set to IPv4Only or IPv6Only to select either version.
-var currentNetworkStack = DefaultNetworkStack
-
 // soaCacheEntry holds a cached SOA record (only selected fields).
 type soaCacheEntry struct {
 	zone      string    // zone apex (a domain name)
@@ -83,11 +65,6 @@ func AddRecursiveNameservers(nameservers []string) ChallengeOption {
 		recursiveNameservers = ParseNameservers(nameservers)
 		return nil
 	}
-}
-
-// SetNetworkStack defines the IP stack that will be used for DNS queries.
-func SetNetworkStack(network NetworkStack) {
-	currentNetworkStack = network
 }
 
 // getNameservers attempts to get systems nameservers before falling back to the defaults.
@@ -272,28 +249,12 @@ func createDNSMsg(fqdn string, rtype uint16, recursive bool) *dns.Msg {
 	return m
 }
 
-// getNetwork interprets the NetworkStack setting in relation to the desired
-// protocol. The proto value should be either "udp" or "tcp".
-func getNetwork(proto string) string {
-	// The dns client passes whatever value is set in [dns.Client.Net] to
-	// the [net.Dialer] (https://github.com/miekg/dns/blob/fe20d5d/client.go#L119-L141).
-	// And the [net.Dialer] accepts strings such as "udp4" or "tcp6"
-	// (https://cs.opensource.google/go/go/+/refs/tags/go1.18.9:src/net/dial.go;l=167-182).
-	if currentNetworkStack == IPv4Only {
-		return proto + "4"
-	}
-	if currentNetworkStack == IPv6Only {
-		return proto + "6"
-	}
-	return proto
-}
-
 func sendDNSQuery(m *dns.Msg, ns string) (*dns.Msg, error) {
-	network := getNetwork("udp")
+	network := currentNetworkStack.Network("udp")
 	udp := &dns.Client{Net: network, Timeout: dnsTimeout}
 	in, _, err := udp.Exchange(m, ns)
 
-	network = getNetwork("tcp")
+	network = currentNetworkStack.Network("tcp")
 	// We can encounter a net.OpError if the nameserver is not listening
 	// on UDP at all, i.e. net.Dial could not make a connection.
 	var opErr *net.OpError
--- a/challenge/dns01/nameserver.go
+++ b/challenge/dns01/nameserver.go
@@ -296,8 +296,8 @@ func sendDNSQuery(m *dns.Msg, ns string) (*dns.Msg, error) {
 	network = getNetwork("tcp")
 	// We can encounter a net.OpError if the nameserver is not listening
 	// on UDP at all, i.e. net.Dial could not make a connection.
-	_, isOpErr := err.(*net.OpError)
-	if (in != nil && in.Truncated) || isOpErr {
+	var opErr *net.OpError
+	if (in != nil && in.Truncated) || errors.As(err, &opErr) {
 		tcp := &dns.Client{Net: network, Timeout: dnsTimeout}
 		// If the TCP request succeeds, the err will reset to nil
 		in, _, err = tcp.Exchange(m, ns)
--- a/challenge/dns01/nameserver.go
+++ b/challenge/dns01/nameserver.go
@@ -251,17 +251,16 @@ func createDNSMsg(fqdn string, rtype uint16, recursive bool) *dns.Msg {
 
 func sendDNSQuery(m *dns.Msg, ns string) (*dns.Msg, error) {
 	network := currentNetworkStack.Network("udp")
-	udp := &dns.Client{Net: network, Timeout: dnsTimeout}
-	in, _, err := udp.Exchange(m, ns)
+	client := &dns.Client{Net: network, Timeout: dnsTimeout}
+	in, _, err := client.Exchange(m, ns)
 
-	network = currentNetworkStack.Network("tcp")
 	// We can encounter a net.OpError if the nameserver is not listening
 	// on UDP at all, i.e. net.Dial could not make a connection.
 	var opErr *net.OpError
 	if (in != nil && in.Truncated) || errors.As(err, &opErr) {
-		tcp := &dns.Client{Net: network, Timeout: dnsTimeout}
+		client.Net = currentNetworkStack.Network("tcp")
 		// If the TCP request succeeds, the err will reset to nil
-		in, _, err = tcp.Exchange(m, ns)
+		in, _, err = client.Exchange(m, ns)
 	}
 
 	return in, err
--- a/cmd/setup_challenges.go
+++ b/cmd/setup_challenges.go
@@ -41,16 +41,21 @@ func setupChallenges(ctx *cli.Context, client *lego.Client) {
 	}
 }
 
-func setupHTTPProvider(ctx *cli.Context) challenge.Provider {
+func resolveHTTP01Network(ctx *cli.Context) http01.ProviderNetwork {
 	var network http01.ProviderNetwork
 	switch {
 	case ctx.IsSet("ipv4only") && ctx.IsSet("ipv6only"):
 		network = http01.DefaultNetwork
 	case ctx.IsSet("ipv4only"):
-		network = http01.Tcp4Network
+		network = http01.TCP4Network
 	case ctx.IsSet("ipv6only"):
-		network = http01.Tcp6Network
+		network = http01.TCP6Network
 	}
+	return network
+}
+
+func setupHTTPProvider(ctx *cli.Context) challenge.Provider {
+	network := resolveHTTP01Network(ctx)
 
 	switch {
 	case ctx.IsSet("http.webroot"):
@@ -99,9 +104,9 @@ func setupTLSProvider(ctx *cli.Context) challenge.Provider {
 	case ctx.IsSet("ipv4only") && ctx.IsSet("ipv6only"):
 		network = tlsalpn01.DefaultNetwork
 	case ctx.IsSet("ipv4only"):
-		network = tlsalpn01.Tcp4Network
+		network = tlsalpn01.TCP4Network
 	case ctx.IsSet("ipv6only"):
-		network = tlsalpn01.Tcp6Network
+		network = tlsalpn01.TCP6Network
 	}
 
 	switch {
--- a/cmd/setup_challenges.go
+++ b/cmd/setup_challenges.go
@@ -81,13 +81,15 @@ func setupHTTPProvider(ctx *cli.Context) challenge.Provider {
 			log.Fatal(err)
 		}
 
-		srv := http01.NewProviderServer(host, port, network)
+		srv := http01.NewProviderServer(host, port)
+		srv.SetNetwork(network)
 		if header := ctx.String("http.proxy-header"); header != "" {
 			srv.SetProxyHeader(header)
 		}
 		return srv
 	case ctx.Bool("http"):
-		srv := http01.NewProviderServer("", "", network)
+		srv := http01.NewProviderServer("", "")
+		srv.SetNetwork(network)
 		if header := ctx.String("http.proxy-header"); header != "" {
 			srv.SetProxyHeader(header)
 		}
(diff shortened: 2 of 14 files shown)


dmke:
> I do not know what "gci-ed" means.

In the context of linters, this usually refers to the [`gci`](https://github.com/daixiang0/gci) package bundled in [`golangci-lint`](https://golangci-lint.run/usage/linters). However, `gci` is concerned with the order of package imports, not with the formatting of the struct fields:

```go
&cli.BoolFlag{
	Name: "ipv4only",
	Aliases: []string{"4"},
	Usage: "Use IPv4 only. This flag is ignored if ipv6only is also specified.",
},
&cli.BoolFlag{
	Name: "ipv6only",
	Aliases: []string{"6"},
	Usage: "Use IPv6 only. This flag is ignored if ipv4only is also specified.",
},
```

which should probably be:

```go
&cli.BoolFlag{
	Name:    "ipv4only",
	Aliases: []string{"4"},
	Usage:   "Use IPv4 only. This flag is ignored if ipv6only is also specified.",
},
&cli.BoolFlag{
	Name:    "ipv6only",
	Aliases: []string{"6"},
	Usage:   "Use IPv6 only. This flag is ignored if ipv4only is also specified.",
},
```

Run `gofmt -w cmd/flags.go` to auto-correct these.

> I also don't understand this one because recursiveNameservers is also a global:

```
challenge/dns01/nameserver.go:31:5: networkStack is a global variable (gochecknoglobals)
var networkStack = "both"
```

Global variables introduce state on the package level, which makes using it as a library harder, and also complicates tests. Unfortunately, avoiding globals is sometimes only achievable with a major refactoring.

For `recursiveNameservers` (and a few others) we have an exception here:

https://github.com/go-acme/lego/blob/1cad41db658621daca9c0950c731a66308ae5e4b/.golangci.toml#L126-L128

IMHO, the `networkStack` variable looks like a candidate for another global.

challenge/dns01/nameserver.go
```diff
@@ -26,6 +26,10 @@ var defaultNameservers = []string{
 // recursiveNameservers are used to pre-check DNS propagation.
 var recursiveNameservers = getNameservers(defaultResolvConf, defaultNameservers)
 
+// networkStack is used to define which IP stack will be used. The default is
+// both IPv4 and IPv6. Set to "ipv4" for IPv4 only, and "ipv6" for IPv6 only.
+var networkStack = "both"
```
dmke, line 31: I believe an enum would be better here:

```suggestion
type networkStack int

const (
	DefaultNetworkStack networkStack = iota
	IPv4Only
	IPv6Only
)

// currentNetworkStack is used to define which IP stack will be used. The default is
// both IPv4 and IPv6. Set to IPv4Only or IPv6Only to select either version.
var currentNetworkStack = DefaultNetworkStack
```

jsumners:
> However, `gci` is concerned with the order of package imports, not with the formatting of the struct fields:

I don't know how you were able to deduce that from the error message, but that seems to have been the trouble. It seems my IDE decided to not auto-format on save like it said it was going to do 😕

> IMHO, the `networkStack` variable looks like a candidate for another global.

👍 I'll add the exception.

challenge/dns01/nameserver.go
```diff
@@ -26,9 +26,17 @@ var defaultNameservers = []string{
...
 
-// networkStack is used to define which IP stack will be used. The default is
-// both IPv4 and IPv6. Set to "ipv4" for IPv4 only, and "ipv6" for IPv6 only.
-var networkStack = "both"
+type networkStack int
+
+const (
+	defaultNetworkStack networkStack = iota
+	ipv4Only
+	ipv6Only
```
dmke, line 34: Please export the values (but keep the type private):

```suggestion
	DefaultNetworkStack networkStack = iota
	IPv4Only
	IPv6Only
```

This eliminates ambiguities on the call-site (and allows to drastically simplify the setter). Since the type stays private, users can't easily/accidentally define additional valid values (maybe via `reflect.TypeOf`, but that won't be accidental anymore).

cmd/setup_challenges.go
```diff
@@ -111,6 +111,19 @@ func setupDNS(ctx *cli.Context, client *lego.Client) {
...
+	case ctx.IsSet("ipv4only") && ctx.IsSet("ipv6only"):
+		// If both flags are set then it's as good as not providing either flag,
+		// so we default to the OS choice.
+		dns01.SetNetworkStack("both")
+
+	case ctx.IsSet("ipv4only"):
+		dns01.SetNetworkStack("ipv4")
+
+	case ctx.IsSet("ipv6only"):
+		dns01.SetNetworkStack("ipv6")
```
dmke, line 124: ... and change this to:

```suggestion
	case ctx.IsSet("ipv4only") && ctx.IsSet("ipv6only"):
		// If both flags are set then it's as good as not providing either flag,
		// so we default to the OS choice.
		dns01.SetNetworkStack(dns01.DefaultNetworkStack)

	case ctx.IsSet("ipv4only"):
		dns01.SetNetworkStack(dns01.IPv4Only)

	case ctx.IsSet("ipv6only"):
		dns01.SetNetworkStack(dns01.IPv6Only)
```

challenge/dns01/nameserver.go
```diff
@@ -67,6 +79,21 @@ func AddRecursiveNameservers(nameservers []string) ChallengeOption {
...
+	case "ipv4":
+		currentNetworkStack = ipv4Only
+
+	case "ipv6":
+		currentNetworkStack = ipv6Only
+
+	default:
+		currentNetworkStack = defaultNetworkStack
+	}
+}
```
dmke, line 95: Simplify the function to:

```suggestion
// SetNetworkStack defines the IP stack that will be used for DNS queries.
func SetNetworkStack(network networkStack) {
	currentNetworkStack = network
}
```

dmke (changes requested):
<details><summary>Put on hold</summary>

I don't want the user to guess whether it was `SetNetworkStack("ipv4")`, or `SetNetworkStack("IPv4")`. Exposing the valid values eliminiates this ambiguity.

</details>

challenge/dns01/nameserver.go
jsumners: I'd also have to make the `networkStack` type public. I made these private in order to appease the linter and follow your original suggestion.

challenge/dns01/nameserver.go
ldez: I'm not a fan of a Go "enum": it's not a real enum, it's just a weak list of values.

jsumners: @ldez what do you want to see?

dmke: As this PR currently only updates the `dns01` package (and leaves `http01` and `tlsalpn01` as a TODO), maybe we can find a way to pass the stack selection to `dns01.NewChallenge`, `http01.NewChallenge` and `tlsalpn01.NewChallenge` as an option?

Looking at `dns01.NewChallenge`, it already accepts a vararg for options:

https://github.com/go-acme/lego/blob/1cad41db658621daca9c0950c731a66308ae5e4b/challenge/resolver/solver_manager.go#L51-L54

Maybe the `networkStack` can move into the `cmd` package (in whatever form), and passing the selection as option to the challenge solver can be done (centrally) in `setupChallenges`?

https://github.com/go-acme/lego/blob/1cad41db658621daca9c0950c731a66308ae5e4b/cmd/setup_challenges.go#L25-L26

jsumners: > (and leaves `http01` and `tlsalpn01` as a TODO)

It does not leave them as todo items. This PR is in draft because I wanted guidance prior to actually working on the full solution. At this point I'm getting conflicting feedback and would like a clear direction. I need this feature and am willing to put in the work, but I don't want to flail around to get it done.

dmke: Sorry for the miscommunication, English isn't my first language. I didn't mean "left as TODO for others", but "for this to work, we need to think of http01 and tlsalpn01 as well".

challenge/dns01/nameserver.go
dmke: I'll put this on hold until we figure out the correct approach.

jsumners:
I know that that the code in the dns challenge is working as intended as I was able to test it in the situation where I need the patch via a local build. 

I didn't receive any further feedback on how we should avoid the enum, so I continued with that construct. I do not believe I am missing any places where these switches would apply, but I don't know the code base that well. Please point out if I have missed any.

dmke:
I'll try to review this on the weekend, I'm currently a bit constrained on time.

jsumners:
😕 This test runs cleanly locally https://github.com/go-acme/lego/actions/runs/3934066967/jobs/6757026356#step:5:278

jsumners:
Yeah, I think these tests are flaky. The same suite failed after my last push, but this time on a different domain name after passing the domain name that failed in the prior run.

https://github.com/go-acme/lego/actions/runs/3953224067/jobs/6769527173#step:5:279

ldez:
I don't think those tests are flaky, because I have the same problem locally.

Can you also lint the code? (`make checks`)

Can you run `make` locally?

jsumners:
> I don't think those tests are flaky, because I have the same problem locally.

The issue was running that subtest in isolation worked, but failed when run with all other subtests. In other words, my added test needed to have a `t.Cleanup` added to restore the expected state for that subtest.

> Can you also lint the code? (`make checks`)
> 
> Can you run `make` locally?

These should be passing now.

dmke:
@jsumners: I've added a bunch of review commits.

The commit messages contain some reasoning, but to reiterate my major changes:

1. Removal of your `go-getport` package. I do appreciate the effort, but that code is actually racy: After calling `getport.GetPort()` and before `server.ListenAndServe()` the OS is allowed and able to reassign the detected port to another process.
   
   We can remove that dependency and a bunch of code by directly using port 0, and retrieve the actual port directly from either the `server.Listener` (TCP) or `server.PacketConn` (UDP).

2. Keeping the API stable. Lego is used as library (notably in the Traefic project), so we can't easily change the exported (public) API. Doing so requires more than a new CLI option :)

   Luckily, we can *extend* the API without much problem, which is related to my third point.

3. Better hiding of the network stack options. @ldez doesn't particularly like iota-constants, and I've mentioned a problem with public types used for internal states (consumers could easily define their own, custom, defunct values, and cause panics in Lego this way).

   The solution I've come up with is to extend the interface and provide distinct functions to set the network stack (`SetIPv4Only`, `SetIPv6Only`, `SetDualStack`). While I welcome bike-shedding for the names, I will point out that all challenge packages use the *same* names.

dmke:
There's a remaining error on Windows:

```
=== RUN   TestFindZoneByFqdnCustom/only_non-existent_nameservers
    nameserver_test.go:265: 
        	Error Trace:	D:\a\lego\lego\challenge\dns01\nameserver_test.go:265
        	Error:      	"could not find the start of authority for mail.google.com.: dial tcp :9053: connectex: No connection could be made because the target machine actively refused it." does not contain "could not find the start of authority for mail.google.com.: dial tcp :9053: connect:"
        	Test:       	TestFindZoneByFqdnCustom/only_non-existent_nameservers
```

Seems like `net.DialContext` on Windows produces a different error message... I'll look into that tomorrow.

jsumners:
> 1. Removal of your `go-getport` package. I do appreciate the effort, but that code is actually racy: After calling `getport.GetPort()` and before `server.ListenAndServe()` the OS is allowed and able to reassign the detected port to another process.

That is explicitly noted in the docs: https://github.com/jsumners/go-getport/blob/ef4d1d0da783a1930d39b868161a94609ab08c71/getport.go#L55-L58. It also has no bearing in these tests. In CI, they are isolated. Running locally, it's extremely unlikely that anything will acquire a port in the nanoseconds between finding the open port and the using it for a test. Still, 🤷‍♂️.

> 3\. Better hiding of the network stack options. @ldez doesn't particularly like iota-constants, and I've mentioned a problem with public types used for internal states

I received no further feedback on which direction to take after asking for it. So I went with the only feedback that had been provided.

> 3\. The solution I've come up with is to extend the interface and provide distinct functions to set the network stack (`SetIPv4Only`, `SetIPv6Only`, `SetDualStack`).

This is fine by me.

> Seems like `net.DialContext` on Windows produces a different error message... I'll look into that tomorrow.

I don't have a Windows environment for testing. All I know is that my last commit was passing with the GitHub CI.

----

Regarding that statement about "my last commit". I'm rather put off by changes being made directly to the branch. I think they could have all been review comments with some suggestions (item 6 in https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/commenting-on-a-pull-request#adding-line-comments-to-a-pull-request) along the way.

I feel like at this point the PR is yours @dmke. I still want to see this added, as the functionality has solved a problem out of my control (my ISP having a bad IPv6 implementation) and likely solves several other issues as stated in the original proposal. So I can provide feedback if desired.

dmke:
>> Seems like net.DialContext on Windows produces a different error message... I'll look into that tomorrow.
> 
> I don't have a Windows environment for testing. All I know is that my last commit was passing with the GitHub CI.

The failure was [introduced here](https://github.com/go-acme/lego/pull/1802/commits/cdf3a6a7da513cc6004c577c602a961e372e78a4?diff=unified&w=1#diff-073ebda03d610a9a2dbcfe427329cff56b8c0bd76390e4d79800b92272488916R291-R304): 

```diff
 func sendDNSQuery(m *dns.Msg, ns string) (*dns.Msg, error) {
-	udp := &dns.Client{Net: "udp", Timeout: dnsTimeout}
+	network := getNetwork("udp")
+	udp := &dns.Client{Net: network, Timeout: dnsTimeout}
 	in, _, err := udp.Exchange(m, ns)
 
-	if in != nil && in.Truncated {
-		tcp := &dns.Client{Net: "tcp", Timeout: dnsTimeout}
+	network = getNetwork("tcp")
+	// We can encounter a net.OpError if the nameserver is not listening
+	// on UDP at all, i.e. net.Dial could not make a connection.
+	_, isOpErr := err.(*net.OpError)
+	if (in != nil && in.Truncated) || isOpErr {
+		tcp := &dns.Client{Net: network, Timeout: dnsTimeout}
 		// If the TCP request succeeds, the err will reset to nil
 		in, _, err = tcp.Exchange(m, ns)
 	}
```

(You've changed the tests for this [here](https://github.com/go-acme/lego/pull/1802/commits/d7da2731301912165c2338c5fe83126610ca8a46#diff-b869997ba97da31f491c5806e1ee04f19604b48e13be1a8e009325ee9cc92d31L264-R273); for that commit I forgot to approve a CI run though.)

The issue is that the TCP-retry-branch is now taken in more cases, as the previous error ("read udp") also is a `net.OpError`. Windows has an optimized syscall, [`ConnectEx`](https://learn.microsoft.com/en-us/windows/win32/api/mswsock/nc-mswsock-lpfn_connectex), for TCP connections, hence the different error message (the [fault location](https://cs.opensource.google/go/go/+/refs/tags/go1.19.4:src/net/fd_windows.go;l=112), for reference).

Tackling this is also a bit cumbersome, as a CI run takes *forever* to complete, and I haven't had a chance to revive my local Windows VM (I've also felt under the whether the last few days, so my capacity for patience to let Windows do its thing was much reduced).


> I think they could have all been review comments with some suggestions  
> [...]  
> I feel like at this point the PR is yours @dmke.

Sorry about that, I didn't want to scare you off. I've started a review the normal way, but it quickly turned into a wild mess of cross-comment-references. Also, review suggestions can only be applied to changed lines, not arbitrary places.

It turned out to be much cleaner to leave review commits and show the changes one-by-one, in separated contexts. In the end the commits will be squashed anyway, so the outcome would be the same.

dmke:
I've spent the last 7 hours in my Windows VM to get the tests green again. Turns out I hate Windows Defender with a passion.

I guess on the CI, Defender is active as well ([log from here](https://github.com/go-acme/lego/actions/runs/4010573872/jobs/6887225015)), and slows down operations considerably:

```
panic: test timed out after 10m0s
...
goroutine 232 [IO wait]:
internal/poll.runtime_pollWait(0x14f3a5a4ed8, 0x72)
	C:/hostedtoolcache/windows/go/1.18.10/x64/src/runtime/netpoll.go:302 +0x89
internal/poll.(*pollDesc).wait(0x10734ae?, 0x1153ab6?, 0x0)
	C:/hostedtoolcache/windows/go/1.18.10/x64/src/internal/poll/fd_poll_runtime.go:83 +0x32
internal/poll.execIO(0xc0003a3b98, 0x1228d10)
	C:/hostedtoolcache/windows/go/1.18.10/x64/src/internal/poll/fd_windows.go:175 +0xe5
internal/poll.(*FD).Read(0xc0003a3b80, {0xc0003b2000, 0x1000, 0x1000})
	C:/hostedtoolcache/windows/go/1.18.10/x64/src/internal/poll/fd_windows.go:441 +0x25f
net.(*netFD).Read(0xc0003a3b80, {0xc0003b2000?, 0xc0003a3c59?, 0x1292a98?})
	C:/hostedtoolcache/windows/go/1.18.10/x64/src/net/fd_posix.go:55 +0x29
net.(*conn).Read(0xc000006518, {0xc0003b2000?, 0xc0003b2000?, 0x0?})
	C:/hostedtoolcache/windows/go/1.18.10/x64/src/net/net.go:183 +0x45
github.com/miekg/dns.(*Conn).Read(0xc0002dbdc0, {0xc0003b2000, 0x1000, 0x1000})
	C:/Users/runneradmin/go/pkg/mod/github.com/miekg/dns@v1.1.50/client.go:340 +0x162
github.com/miekg/dns.(*Conn).ReadMsgHeader(0xc0002dbdc0, 0x0)
	C:/Users/runneradmin/go/pkg/mod/github.com/miekg/dns@v1.1.50/client.go:304 +0xb7
github.com/miekg/dns.(*Conn).ReadMsg(0xc0002dbdc0)
	C:/Users/runneradmin/go/pkg/mod/github.com/miekg/dns@v1.1.50/client.go:269 +0x27
github.com/miekg/dns.(*Client).exchangeContext(0xc0002e5f80, {0x128fdc8, 0xc00009a0b8}, 0xc000358900, 0xc0002dbdc0)
	C:/Users/runneradmin/go/pkg/mod/github.com/miekg/dns@v1.1.50/client.go:246 +0x445
github.com/miekg/dns.(*Client).exchangeWithConnContext(0x0?, {0x128fdc8?, 0xc00009a0b8?}, 0x0?, 0x0?)
	C:/Users/runneradmin/go/pkg/mod/github.com/miekg/dns@v1.1.50/client.go:195 +0x1d7
github.com/miekg/dns.(*Client).ExchangeWithConn(0xc0002e5f80?, 0x128fdc8?, 0xc00009a0b8?)
	C:/Users/runneradmin/go/pkg/mod/github.com/miekg/dns@v1.1.50/client.go:190 +0x30
github.com/miekg/dns.(*Client).Exchange(0x128d620?, 0xc00038ea00?, {0x11f218c?, 0xc000006510?})
	C:/Users/runneradmin/go/pkg/mod/github.com/miekg/dns@v1.1.50/client.go:170 +0x10e
github.com/go-acme/lego/v4/challenge/dns01.sendDNSQuery(0x11f6850?, {0x11f218c, 0x5})
	D:/a/lego/lego/challenge/dns01/nameserver.go:255 +0x136
github.com/go-acme/lego/v4/challenge/dns01.dnsQuery({0x11f6850?, 0x10?}, 0x20?, {0x14917c0, 0x3, 0xc00036e0f0?}, 0xf8?)
	D:/a/lego/lego/challenge/dns01/nameserver.go:232 +0xaa
github.com/go-acme/lego/v4/challenge/dns01.fetchSoaByFqdn({0x11f6844, 0x10}, {0x14917c0, 0x3, 0x3})
	D:/a/lego/lego/challenge/dns01/nameserver.go:177 +0x14e
github.com/go-acme/lego/v4/challenge/dns01.lookupSoaByFqdn({0x11f6844, 0x10}, {0x14917c0, 0x3, 0x3})
	D:/a/lego/lego/challenge/dns01/nameserver.go:160 +0x14f
github.com/go-acme/lego/v4/challenge/dns01.FindPrimaryNsByFqdnCustom({0x11f6844?, 0x13fdcf8?}, {0x14917c0?, 0xf?, 0x12ea146?})
	D:/a/lego/lego/challenge/dns01/nameserver.go:128 +0x2d
github.com/go-acme/lego/v4/challenge/dns01.TestFindPrimaryNsByFqdnCustom.func1(0x0?)
	D:/a/lego/lego/challenge/dns01/nameserver_test.go:282 +0x65
testing.tRunner(0xc000372000, 0xc00032b410)
	C:/hostedtoolcache/windows/go/1.18.10/x64/src/testing/testing.go:1439 +0x102
created by testing.(*T).Run
	C:/hostedtoolcache/windows/go/1.18.10/x64/src/testing/testing.go:1486 +0x35f
FAIL	github.com/go-acme/lego/v4/challenge/dns01	600.045s
```

In `nameserver.go:255`, we're calling `(*dns.Client).Exchange`, with `client.Net == "udp"` (i.e. before the TCP/ConnectEx retry-branch) and a timeout set to 20s. Clearly this timeout does not have an effect.

I'm baffled. This part of the code didn't change, how come it stopped working?

@ldez, do you have an idea?

(I'm frustrated enough to consider proposing to downgrade Windows support for Lego to "experimental", and drop Windows from the CI workflow altogether... Maybe I should sleep on this for a bit before doing so :/)

jsumners:
> I've spent the last 7 hours in my Windows VM to get the tests green again.

😨

jsumners:
What can I do to help get this to the finish line?

ldez:
I will spend some time on that in the following days

dmke:
I'm currently investigating as too why Windows performs abysmally poor, and seeking ways to improve that. However, that's going very slowly...

dmke:
I'm currently investigating as too why Windows performs abysmally poor, and seeking ways to improve that. However, that's going very slowly...

jsumners:
Closing in favor of https://github.com/go-acme/lego/pull/1984.

PR #1813, opened 2023-01-20
hetzner: fix missing pagination in getZones()

MaxKlaxxMiner:
Fix some unexpected "could not get zone"-errors if more than 100 DNS records exist in your account.


ldez:
I read the Hetzner API doc and GetZones can use a `name` so we don't need to get all zones.
https://dns.hetzner.com/api-docs#operation/GetZones

So I created another PR (#1815)

PR #1815, opened 2023-01-20
hetzner: improve zone ID detection

ldez:
Superseeds #1813

Fix unexpected "could not get zone"-errors when there are more than 100 zones.

### DIFF
--- a/providers/dns/hetzner/internal/client.go
+++ b/providers/dns/hetzner/internal/client.go
@@ -8,12 +8,16 @@ import (
 	"net/http"
 	"net/url"
 	"path"
+
+	querystring "github.com/google/go-querystring/query"
 )
 
 // defaultBaseURL represents the API endpoint to call.
 const defaultBaseURL = "https://dns.hetzner.com"
 
 const authHeader = "Auth-API-Token"
+
+const pageSize = 100
 
 // Client the Hetzner client.
 type Client struct {
@@ -129,7 +133,7 @@ func (c *Client) DeleteRecord(recordID string) error {
 
 // GetZoneID gets the zone ID for a domain.
 func (c *Client) GetZoneID(domain string) (string, error) {
-	zones, err := c.getZones()
+	zones, err := c.getZones(domain, &Pagination{Page: 1, PerPage: pageSize})
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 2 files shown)


MaxKlaxxMiner:
Thank you for the revision. I will test it right away on Monday as soon as I am back in the office (have 130 domains in my Hetzner account).

providers/dns/hetzner/internal/client.go
```diff
@@ -129,7 +133,7 @@ func (c *Client) DeleteRecord(recordID string) error {
 
 // GetZoneID gets the zone ID for a domain.
 func (c *Client) GetZoneID(domain string) (string, error) {
-	zones, err := c.getZones()
+	zones, err := c.getZones(domain, &Pagination{Page: 1, PerPage: pageSize})
```
dmke, line 136: This is a curious one. I was wondering, why the page number isn't iterated on.

The docs state for the `name` parameter (here given as `domain`), that it "will return an array with one or no results". Also, there's a curiously worded "400 Pagination selectors are mutually exclusive" response (without further explanation).

I'm interpreting that as applying pagination (`page=1&per_page=100`) would be either unnecessary (because there's only one page with 0-1 entries), or erroneous (because we've selected to paginate by name, not by cursor).

dmke (changes requested):
I'm not convinced this works...

providers/dns/hetzner/internal/client.go
```diff
@@ -144,17 +146,37 @@ func (c *Client) GetZoneID(domain string) (string, error) {
...
+	if pagination != nil {
+		values, err = querystring.Values(pagination)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse query parameters: %w", err)
+		}
+	}
+
+	if name != "" {
+		values.Set("name", name)
+	}
```
dmke, line 166: `url.Values` is a `map` type, and `value.Set` will crash if `pagination` is `nil`

dmke: https://go.dev/play/p/UIUn94iY8sK

ldez: yes I did that a bit too quickly :wink:

providers/dns/hetzner/internal/client.go
```diff
@@ -144,17 +146,37 @@ func (c *Client) GetZoneID(domain string) (string, error) {
 }
 
 // https://dns.hetzner.com/api-docs#operation/GetZones
-func (c *Client) getZones() (*Zones, error) {
+func (c *Client) getZones(name string, pagination *Pagination) (*Zones, error) {
```
dmke, line 149: It doesn't look like `getZones` is used outside of `GetZoneID`, hence `pagination` is always `nil`. Do we want to drop it (it is basically dead code now)?

ldez: I want to keep it for now, because I don't trust API documentation of DNS providers, so I want to wait for the feedback of @MaxKlaxxMiner

dmke: Sure, sounds good.

MaxKlaxxMiner: Ok, all versions of the fix currently work without issues.

I checked the name parameter more closely and found the following:
- If name parameter is sent empty or is omitted completely, then getZones always returns all domains. (as expected)
- If you search with name, the search is always Exact: If you search e.g. "vacando.co" (columbia), also only this one domain is returned and not the "vacando.co.uk" (united kingdom). If you search only "vacando", nothing will be found.

Searching by name can only return a single domain or a 404-error: Maybe a special single-request-function "getZoneByName" without pagnation would be enough here?

ldez: thank you for the feedback I will remove the pagination paramter

MaxKlaxxMiner: Yep and the request by name is also surprisingly much faster:
950-1120 ms (all = 2 pages = 130 domains)
580-660 ms (only 1 page = first 100 domains)
130-150 ms (search per name)

dmke (approved):

PR #1833, opened 2023-02-05
namesilo: add cleanup before add a DNS record

### LINKED ISSUE #1831: Namesilo DNS provider: fails if the record exists
### Welcome

If the challenge record exists for some reason (created manually, or by another software, or if I run the `run` command and then stop it with `ctrl+c` after it created the record, but before it finish, so the record remains), I expect to see the provider checks whether it exists, and if yes, it just replaces it with DnsUpdateRecord.

### What did you see instead?

In the described situation it just gives up with the error like 'duplicate record', and I should delete it manually.

### How do you use lego?

Binary

### Reproduction steps

1. create somehow the appropriate record for the domain with any value
2. run `lego [..] --dns namesilo [..] run`
4. it fails with 'duplicate record'

or

1. run `lego [..] --dns namesilo [..] run`
2. press `ctrl+c` after it created the record, but before it finish, so the record remains
3. rerun the command from point 1
4. it fails with 'duplicate record'

### Version of lego

```console
`lego version dev linux/amd64` built on 2023/01/31
```

### Logs

<details>
I looked at the code at `lego/providers/dns/namesilo/namesilo.go, at the `DNSProvider.Present` method, and in fact it executes only `DnsAddRecord`:

```go
	_, err = d.client.DnsAddRecord(&namesilo.DnsAddRecordParams{
		Domain: zoneName,
		Type:   "TXT",
		Host:   subdomain,
		Value:  value,
		TTL:    d.config.TTL,
	})
	if err != nil {
		return fmt.Errorf("namesilo: failed to add record %w", err)
	}
```

</details>

_No response_

e1fueg0:
fixes #1831

### DIFF
--- a/providers/dns/namesilo/namesilo.go
+++ b/providers/dns/namesilo/namesilo.go
@@ -98,6 +98,10 @@ func (d *DNSProvider) Present(domain, token, keyAuth string) error {
 		return fmt.Errorf("namesilo: %w", err)
 	}
 
+	err = d.CleanUp(domain, token, keyAuth)
+	if err != nil {
+		return err
+	}
 	_, err = d.client.DnsAddRecord(&namesilo.DnsAddRecordParams{
 		Domain: zoneName,
 		Type:   "TXT",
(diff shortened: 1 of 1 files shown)


ldez:
LGTM

But I want to add a note: lego cleans the TXT records that he creates.
I think that cleaning TXT records before adding a record is something out of the scope of lego.

ldez (approved):

e1fueg0:
@ldez sounds reasonable, but what if lego created one, and then it was stopped before it had time to clean it? then automation will be broken, because who then should clean it?

ldez:
> what if lego created one, and then it was stopped before it had time to clean it? then automation will be broken, because who then should clean it?

if a user kills the lego process, he has to assume the fact to clean the current state.

e1fueg0:
@ldez anyway, thank you very much for merging this.

PR #1835, opened 2023-02-09
chore: update to go1.19

ldez:
- updates CI to use go1.19 and go1.20
- uses either `URL.JoinPath` or `url.JoinPath` instead of `url.Parse`+ `path.Join` => removes more than 250 lines!
- update linter

https://tip.golang.org/doc/go1.19

Note:
`URL.JoinPath` (or `url.JoinPath`) is not the exact equivalent of `url.Parse`+ `path.Join` because the path elements in `URL.JoinPath` are path encoded which means that if an element contains a query string the query string will be path encoded (`%3F` instead of `?`).


dmke:
Will take a look later.

.github/workflows/go-cross.yml
```diff
@@ -16,7 +16,7 @@ jobs:
 
     strategy:
       matrix:
-        go-version: [ 1.18, 1.19, 1.x ]
+        go-version: [ 1.19, '1.20', 1.x ]
```
dmke, line 19: Let's generally quote things that could be mistaken as floating point numbers by YAML parsers:

```suggestion
        go-version: [ '1.19', '1.20', 1.x ]
```

providers/dns/hyperone/internal/client.go
```diff
@@ -77,9 +76,12 @@ func NewClient(apiEndpoint, locationID string, passport *Passport) (*Client, err
 // https://api.hyperone.com/v2/docs#operation/dns_project_zone_recordset_list
 func (c *Client) FindRecordset(zoneID, recordType, name string) (*Recordset, error) {
 	// https://api.hyperone.com/v2/dns/{locationId}/project/{projectId}/zone/{zoneId}/recordset
-	resourceURL := path.Join("dns", c.locationID, "project", c.projectID, "zone", zoneID, "recordset")
+	endpoint, err := url.JoinPath(c.apiEndpoint, "dns", c.locationID, "project", c.projectID, "zone", zoneID, "recordset")
+	if err != nil {
+		return nil, err
+	}
```
dmke, line 82: There's a lot of repetition here, in `CreateRecordset`, in `DeleteRecordset`, in `GetRecords`, and in `GetZones`.

Can we make `(*Client).apiEndpoint` a `*url.URL`, and use `c.apiEndoint.JoinPath(...)`?

dmke (changes requested):
Nice cleanup, `JoinPath` makes this much more readable than `path.Join`+`url.Parse`+error handling.

In the hyperone client, I'd like to see the URL parsing moved to `NewClient`. This also has the benefit of catching parse errors at initialization time, and not at a later `Present`/`Cleanup` call.

providers/dns/hyperone/internal/client.go
ldez: something like that? https://github.com/go-acme/lego/pull/1835/commits/b0b91728c49a6cb396fbfe8729db5ea474350e3e

dmke: That, plus

```diff
-endpoint, err := url.JoinPath(c.apiEndpoint, "dns", c.locationID, "project", c.projectID, "zone", zoneID, "recordset")
-if err != nil {
-	return nil, err
-}
+endpoint := c.apiEndpoint.JoinPath("dns", c.locationID, "project", c.projectID, "zone", zoneID, "recordset")
```

ldez: is it better like that? https://github.com/go-acme/lego/pull/1835/commits/5fd749110ab553bda056b1b81a4991c6e599c907

ldez: clean the constructor in https://github.com/go-acme/lego/pull/1835/commits/0d78446553de4c048ab3366cb1d8a4d9cc1b96fb

PR #1838, opened 2023-02-14
support for certificate with raw IP SAN #1649

### LINKED ISSUE #1649: support RFC8738

orangepizza:
have to create new PR because old PR was from master of fork. (why GitHub not allow create new source branch for PR 👎 )
will resolve #1649

----
a tiny edits on few functions that adapt type of identifier/ SAN type if input 'domain' can be parsed as IP address.
not bothered to change name of variable []domains to some better name: kind don't want to follow all those function calls to pass full identifier

test on pebble that this is enough for ipv4 and ipv6 in -d.


orangepizza:
why Mac OS ignore RFC3330 and only has 127.0.0.1/32, not 127.0.0.1/8  like everyone else?

orangepizza:
could you trigger ci run again? it errored on a dns api extension this doesn't touch.

certcrypto/crypto.go
```diff
@@ -277,9 +306,15 @@ func generateDerCert(privateKey *rsa.PrivateKey, expiration time.Time, domain st
 
 		KeyUsage:              x509.KeyUsageKeyEncipherment,
 		BasicConstraintsValid: true,
-		DNSNames:              []string{domain},
 		ExtraExtensions:       extensions,
 	}
+	// handling SAN filling as type suspected
+	ip := net.ParseIP(domain)
+	if ip != nil {
```
dmke, line 313:
```suggestion
	if ip := net.ParseIP(domain); ip != nil {
```

challenge/tlsalpn01/tls_alpn_challenge_test.go
```diff
@@ -116,3 +119,75 @@ func TestChallengeInvalidPort(t *testing.T) {
...
+		ext := remoteCert.Extensions[idx]
+		assert.True(t, ext.Critical, "Expected the challenge certificate id-pe-acmeIdentifier extension to be marked as critical")
+
+		zBytes := sha256.Sum256([]byte(chlng.KeyAuthorization))
+		value, err := asn1.Marshal(zBytes[:sha256.Size])
+		require.NoError(t, err, "Expected marshaling of the keyAuth to return no error")
+
+		if subtle.ConstantTimeCompare(value, ext.Value) != 1 {
+			t.Errorf("Expected the challenge certificate id-pe-acmeIdentifier extension to contain the SHA-256 digest of the keyAuth, %v, but was %v", zBytes[:], ext.Value)
+		}
```
dmke, line 164: `require.EqualValues` should suffice here:

```suggestion
		require.EqualValues(t, value, ext.Value, "Expected the challenge certificate id-pe-acmeIdentifier extension to contain the SHA-256 digest of the keyAuth")
```

(I may have the order of expected/actual value wrong)

challenge/tlsalpn01/tls_alpn_challenge_test.go
```diff
@@ -116,3 +119,75 @@ func TestChallengeInvalidPort(t *testing.T) {
...
+			Type:  "ip",
+			Value: domain,
+		},
+		Challenges: []acme.Challenge{
+			{Type: challenge.TLSALPN01.String(), Token: "tlsalpn1"},
+		},
+	}
+
+	err = solver.Solve(authz)
+	require.NoError(t, err)
```
dmke, line 192:
```suggestion
	require.NoError(t, solver.Solve(authz))
```

challenge/tlsalpn01/tls_alpn_challenge_test.go
```diff
@@ -116,3 +119,75 @@ func TestChallengeInvalidPort(t *testing.T) {
...
+		conn, err := tls.Dial("tcp", net.JoinHostPort(domain, port), &tls.Config{
+			InsecureSkipVerify: true,
+		})
+		require.NoError(t, err, "Expected to connect to challenge server without an error")
+
+		// Expect the server to only return one certificate
+		connState := conn.ConnectionState()
+		assert.Len(t, connState.PeerCertificates, 1, "Expected the challenge server to return exactly one certificate")
+
+		remoteCert := connState.PeerCertificates[0]
```
dmke, line 139: [RFC8738, section 6](https://datatracker.ietf.org/doc/html/rfc8738#section-6) requires the SNI extension to contain the `.in-addr.arpa` or `.ip6.arpa` reverse mapping of the IP address.

Can you add a test for this? [`(miekg/dns).ReverseAddr`](https://pkg.go.dev/github.com/miekg/dns#ReverseAddr) will be of help here.

challenge/tlsalpn01/tls_alpn_challenge_test.go
```diff
@@ -116,3 +119,75 @@ func TestChallengeInvalidPort(t *testing.T) {
...
+			if idPeAcmeIdentifierV1.Equal(ext.Id) {
+				idx = i
+				break
+			}
+		}
+
+		require.NotEqual(t, -1, idx, "Expected the challenge certificate to contain an extension with the id-pe-acmeIdentifier id,")
+
+		ext := remoteCert.Extensions[idx]
+		assert.True(t, ext.Critical, "Expected the challenge certificate id-pe-acmeIdentifier extension to be marked as critical")
```
dmke, line 156:
```suggestion
		var foundAcmeIdentifier bool
		for i, ext := range remoteCert.Extensions {
			if idPeAcmeIdentifierV1.Equal(ext.Id) {
				assert.True(t, ext.Critical, "Expected the challenge certificate id-pe-acmeIdentifier extension to be marked as critical")
				foundAcmeIdentifier = true
				break
			}
		}

		require.True(t, foundAcmeIdentifier, "Expected the challenge certificate to contain an extension with the id-pe-acmeIdentifier id,")
```

dmke (changes requested):
Overall this looks good! I've left some initial thoughts/comments.

challenge/tlsalpn01/tls_alpn_challenge_test.go
orangepizza: don't think we need that here, SNI extension is something ACME server's VA should sent, and out TLS-ALPN solver don't care about incomming traffics SNI and just give same certificate

challenge/tlsalpn01/tls_alpn_challenge_test.go
orangepizza: it's copied from DNS name version, do you want to change it in this commit too?

dmke: Oh, I didn't realize that.

It would be nice if you could cleanup the other test version as well, yes.

challenge/tlsalpn01/tls_alpn_challenge_test.go
dmke: Hm. My thinking is: if the server MUST send the IP address in the SNI extension, the client should validate it is correct. The `http-01` challenge solver does something similar with the `Host`/`Forwarded-*` headers (this is mostly to prevent DNS rebind attacks though).

@ldez: Do you have an opinion on this?

certcrypto/crypto.go
```diff
@@ -131,11 +132,22 @@ func GeneratePrivateKey(keyType KeyType) (crypto.PrivateKey, error) {
...
+		} else {
+			ips = append(ips, ip)
+		}
 	}
 
+	template := x509.CertificateRequest{
+		Subject: pkix.Name{CommonName: domain},
+		DNSNames: dnss,
+		IPAddresses: ips,
+	}	}
```
dmke, line 149: There's a stray `}` here. Sorry if my previous suggestion caused this.

orangepizza:
> Hm. My thinking is: if the server MUST send the IP address in the SNI extension, the client should validate it is correct. The `http-01` challenge solver does something similar with the `Host`/`Forwarded-*` headers (this is mostly to prevent DNS rebind attacks though).

as It current TLS-ALPN answerer didn't care or domain test send SNI by moke VA before this commit, and as this is acme client we would be attacker side of such attack, I don't think we need to verify ACME server do the right thithes, that'd be ACME server's duty to cornfirm they are strict enough.

dmke:
> we would be attacker side of such attack

This is not entirely true. With challenges like `tls-alpn-01`, ACME clients need to make the client host available to the outside world, by means of starting a web server. IIRC, the certificate order (be it a new certificate or renewal of an existing one) is publicly available, even before a certificate is issued. (Yes, the order URL is not easily guessable; no, there's no access control.)

This could allow potential 3rd-party attackers to meddle with the running Lego instance.

There's also random chance (e.g. after port scan or by sheer luck), that someone tries to access the TLS port.

Measures like checking the SNI field raise the bar for anyone accessing the running Lego server; any deviation from the standard should be rejected *immediately*.

challenge/tlsalpn01/tls_alpn_challenge_server.go
```diff
@@ -94,3 +94,25 @@ func (s *ProviderServer) CleanUp(domain, token, keyAuth string) error {
...
+	// Must be IPv6
+	buf := make([]string, 0, len(ip)+1)
+	// Add it, in reverse, to the buffer
+	for i := len(ip) - 1; i >= 0; i-- {
+		buf = append(buf, fmt.Sprintf("%x.%x", ip[i]&0x0F, ip[i]>>4))
+	}
+	// Append "ip6.arpa." and return (buf already has the final '.') see RFC3152 for how this address is constructed.
+	buf = append(buf, "ip6.arpa.")
+	return strings.Join(buf, ".")
+}
```
dmke, line 118: As mentioned before, use `(github.com/miegk/dns).ReverseAddr`.

challenge/tlsalpn01/tls_alpn_challenge_test.go
```diff
@@ -116,3 +120,72 @@ func TestChallengeInvalidPort(t *testing.T) {
...
+
+func TestChallengeIPaddress(t *testing.T) {
+	_, apiURL := tester.SetupFakeAPI(t)
+
+	domain := "127.0.0.1"
+	port := "23457"
+
+	mockValidate := func(_ *api.Core, _ string, chlng acme.Challenge) error {
+		conn, err := tls.Dial("tcp", net.JoinHostPort(domain, port), &tls.Config{
+			ServerName:         reverseaddr(domain),
```
dmke, line 132:
```suggestion
		arpa, _ := dns.ReversAddr(domain)
		conn, err := tls.Dial("tcp", net.JoinHostPort(domain, port), &tls.Config{
			ServerName:         arpa,
```

And add `github.com/miekg/dns` to the `import`s.

dmke (changes requested):
The `reverseaddr` implementation is redundant. The DNS library we're using already has such a function.

orangepizza:
> 

1. even if someone else visit our alpn server, connection will closed right after tls handshake anyway (effectively we reject everything), and if go TLS is hopelessly broken such that handshake alone enough to break it, we'd have much bigger problem.
2. if attacker saw to certificate we tried to issue they already know the right name to send. 
3. order url require POST-as-GET request signed by JWT signed by account key, so attacker can't read.
4. we didn't do that on domain name, doing that on this PR feels it will blur the scope of PR.

dmke:
You raise some valid points there, I'm glad you're on top of things.

Regarding (4), I agree. I was under the assumption that we're already checking the SNI field.

I'm currently a bit busy to finish the review in detail, but glancing over it, I don't see any deficits. If @ldez doesn't take over from here, you may need to wait until the weekend before I find some time to do it myself (sorry for that).

ldez (approved):
LGTM

Issue #1323, opened 2020-12-25
Namecheap - Waiting for DNS record propagation - Unable to obtain certificates


omares:
Hey, I am unable to obtain valid certificates for my Namecheap domains as the process is unable to validate the DNS entries. I tested this first with [udm-le](https://github.com/kchristensen/udm-le) which in its core is running lego inside docker/podman and also with a local docker setup.

Follow settings have been used:

```
NAMECHEAP_TTL=60
NAMECHEAP_PROPAGATION_TIMEOUT=300
NAMECHEAP_API_USER=xxx
NAMECHEAP_API_KEY=xxx
DNS_PROVIDER='namecheap'
```

The docker command executed:
```
docker run --env-file=lego.env goacme/lego -d pihole.mares.id -a -m xxx@xxx.TLD --dns.resolvers dns1.registrar-servers.com --dns namecheap run
```

To be 100% sure the Namecheap DNS host `dns1.registrar-servers.com` is provided as a resolver.

The log output:

```
2020/12/25 21:19:56 [INFO] [unifi.mares.id, pihole.mares.id] acme: Obtaining bundled SAN certificate
2020/12/25 21:19:57 [INFO] [pihole.mares.id] AuthURL: https://acme-v02.api.letsencrypt.org/acme/authz-xxxx
2020/12/25 21:19:57 [INFO] [unifi.mares.id] AuthURL: https://acme-v02.api.letsencrypt.org/acme/authz-xxxx
2020/12/25 21:19:57 [INFO] [pihole.mares.id] acme: Could not find solver for: tls-alpn-01
2020/12/25 21:19:57 [INFO] [pihole.mares.id] acme: Could not find solver for: http-01
2020/12/25 21:19:57 [INFO] [pihole.mares.id] acme: use dns-01 solver
2020/12/25 21:19:57 [INFO] [unifi.mares.id] acme: Could not find solver for: tls-alpn-01
2020/12/25 21:19:57 [INFO] [unifi.mares.id] acme: Could not find solver for: http-01
2020/12/25 21:19:57 [INFO] [unifi.mares.id] acme: use dns-01 solver
2020/12/25 21:19:57 [INFO] [pihole.mares.id] acme: Preparing to solve DNS-01
2020/12/25 21:19:59 [INFO] [unifi.mares.id] acme: Preparing to solve DNS-01
2020/12/25 21:20:00 [INFO] [pihole.mares.id] acme: Trying to solve DNS-01
[...]

PR #1841, opened 2023-02-17
Fallback to TCP DNS challenge if UDP fails

reinismu:
I'm currently in network where UDP DNS query doesn't work. Would be nice if library would fallback to TCP in this case. Or if we could configure it. Other people would appreciate this feature as well https://github.com/go-acme/lego/issues/1323#issuecomment-751387236


ldez:
Hello,

always fallback on TCP when there is an error seems overkill.

I will decline this PR, I suggest opening an issue instead to explain your use case.

reinismu:
> always fallback on TCP when there is an error seems overkill.

If I make a PR adding ENV variable to default to TCP you will accept it?

> I will decline this PR, I suggest opening an issue instead to explain your use case.

"I'm currently in network where UDP DNS query doesn't work". Idk why it doesn't work. Maybe ISP.

Issue #1672, opened 2022-07-04
Dynu dns-challenge fails creating a certificates for subdomains


daneget:
### Welcome

- [X] Yes, I'm using a binary release within 2 latest releases.
- [X] Yes, I've searched similar issues on GitHub and didn't find any.
- [X] Yes, I've included all information below (version, config, etc).

### What did you expect to see?

While dynu dns-challenge is working to generate a certificate, I expect to see an _acme-challenge.subdomain.example.com txt record in my dns manager

### What did you see instead?

Dynu dns-challenge does not create a certificate and I saw an _acme-challenge.example.com txt record that is not deleted in my dns manager

### How do you use lego?

Through Traefik

### Reproduction steps

1. Configure traefik with dynu dns-challenge including the needed environment variable with the api key
2. Add a website redirection througth traefik using subdomain.example.com as domain
3. Open the website in a navigator using the domain configured in traefik

### Version of lego

```console
Traefik: 2.8.0
Lego: v4.7.0
```

### Logs

<details>

```console
proxy-traefik-1  | time="2022-07-04T21:45:17Z" level=debug msg="legolog: [INFO] [subdomain.example.com] acme: Obtaining bundled SAN certificate"
proxy-traefik-1  | time="2022-07-04T21:45:18Z" level=debug msg="legolog: [INFO] [subdomain.example.com] AuthURL: https://acme-v02.api.letsencrypt.org/acme/authz-v3/126983907086"
proxy-traefik-1  | time="2022-07-04T21:45:18Z" level=debug msg="legolog: [INFO] [subdomain.example.com] acme: Could not find solver for: tls-alpn-01"
proxy-traefik-1  | time="2022-07-04T21:45:18Z" level=debug msg="legolog: [INFO] [subdomain.example.com] acme: Could not find solver for: http-01"
proxy-traefik-1  | time="2022-07-04T21:45:18Z" level=debug msg="legolog: [INFO] [subdomain.example.com] acme: use dns-01 solver"
proxy-traefik-1  | time="2022-07-04T21:45:18Z" level=debug msg="legolog: [INFO] [subdomain.example.com] acme: Preparing to solve DNS-01"
proxy-traefik-1  | time="2022-07-04T21:45:24Z" level=debug msg="legolog: [INFO] [subdomain.example.com] acme: Cleaning DNS-01 challenge"
[...]

PR #1842, opened 2023-02-19
dynu: fix subdomain support

SMHRambo:
Hi,

Fixes https://github.com/go-acme/lego/issues/1672

It should fix the subdomain issue.
Hope it helps.

### DIFF
--- a/providers/dns/dynu/dynu.go
+++ b/providers/dns/dynu/dynu.go
@@ -97,8 +97,7 @@ func (d *DNSProvider) Timeout() (timeout, interval time.Duration) {
 func (d *DNSProvider) Present(domain, token, keyAuth string) error {
 	fqdn, value := dns01.GetRecord(domain, keyAuth)
 
-	// TODO(ldez) replace domain by FQDN to follow CNAME.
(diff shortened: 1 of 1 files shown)


SMHRambo (commented):
I tested your code change but I get the error "dynu: could not find root domain".
I also tested it with a CNAME recorde, with the same result.
The old code accepted CNAME records without problems.
That being said, DYNU's API or the GetRootDomain function don't seem to like the dot at the end of the fqdn.
I'm not very good at reading go code, it's my second day with go.

ldez:
Thank you for the feedback :+1: 

The old code wasn't working with CNAME, I'm 100% sure.
I think I have to trim the `_acme-challenge` prefix from the FQDN.

ldez:
I changed the code, can you try it?

SMHRambo:
The problem is the dot at the end.
I tested the code and added the string "_acme-challenge" before the domain without a problem.

ldez:
So my latest change fixes the problem.

SMHRambo (commented):
This is working!

ldez (approved):
Thank you :+1:

PR #1843, opened 2023-02-20
Experimental option to force DNS queries to use TCP

reinismu:
Some networks block UDP packets. The ability to switch to TCP-only DNS queries will help a lot here.

`LEGO_EXPERIMENTAL_DNS_TCP_ONLY=true` to use TCP

https://community.traefik.io/t/docker-traefik-route53-dns-01-challenge-not-renewing-certificates/17463

### DIFF
--- a/challenge/dns01/nameserver.go
+++ b/challenge/dns01/nameserver.go
@@ -4,6 +4,8 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"os"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


reinismu:
Any updates?

ldez:
I have problems with this PR:
- we have no open issue about that, and no :+1: on the PR, it means there is no traction.
- the problem behind this PR is just network configuration (firewall, iptables, ...)

Adding an option to only one person with a problem that can be fixed by a network configuration doesn't seem the right thing.

So we have 2 possibilities:
1. Decline the PR
2. Convert this option to an experimental option (i.e. an option that we can remove)

dmke:
I have a similar opinion. The obvious solution is to fix the network first.

However, the change is small enough to hide behind a `LEGO_EXPERIMENTAL_DNS_TCP_ONLY` flag, and decline any support for it (if it works for you: great, otherwise: go fix your network).

reinismu:
Ohh I tried to fix it. From what I could tell it is in the internet provider's side. I tried to make those requests in multiple networks and I  didn't notice any options on the router that would block it from working. I don't see why you couldn't add `LEGO_EXPERIMENTAL_DNS_TCP_ONLY` flag for this case.  I'm surprised it wasn't there in the first  place

ldez (approved):

Issue #1854, opened 2023-03-01
LE cannot do rsa8192 - lego should not offer it


systemcrash:
### Welcome

- [X] Yes, I'm using a binary release within 2 latest releases.
- [X] Yes, I've searched similar issues on GitHub and didn't find any.
- [X] Yes, I've included all information below (version, config, etc).

### What did you expect to see?

Success

### What did you see instead?

```console
2023/03/01 14:49:50 Could not complete registration
	acme: error: 400 :: POST :: https://acme-v02.api.letsencrypt.org/acme/new-acct :: urn:ietf:params:acme:error:badPublicKey :: key size not supported: 8192
```

BTW: LE supports 2K, 3K and 4K RSA, if this info is still valid. The above error indicates this is the case.
https://letsencrypt.org/docs/integration-guide/#supported-key-algorithms

### How do you use lego?

Binary

### Reproduction steps

1. `/root/lego -a -k rsa8192 ...`

### Version of lego

```console
lego version 4.10.2 linux/amd64
```

### Logs

<details>

```console
# /root/lego -a -k rsa8192 ...
2023/03/01 14:49:49 [INFO] acme: Registering account for root@example.com
2023/03/01 14:49:50 Could not complete registration
	acme: error: 400 :: POST :: https://acme-v02.api.letsencrypt.org/acme/new-acct :: urn:ietf:params:acme:error:badPublicKey :: key size not supported: 8192
```

</details>

ldez:
Let's Encrypt != ACME

Every ACME server can implement or not this.

systemcrash:
ACME is a protocol - but I'd be interested to know if there are any public ACME CAs (non eAB) which offer RSA8192.

LE: no (max 4096)
ZeroSSL: no (max 3072)
Buypass: no (max 4096)
ssl.com: no (max 4096)
Google: no (max 4096)

PR #1855, opened 2023-03-01
Remove RSA 8192 - which LetsEncrypt does not support.

systemcrash:
Add RSA 3072 - which LetsEncrypt does support.

It seems safe to remove RSA8192 - even modified LEGO binaries for use with e.g. ZeroSSL cannot go above 3072. ZeroSSL supports max RSA3072.

Should fix issue #1854

### DIFF
--- a/certcrypto/crypto.go
+++ b/certcrypto/crypto.go
@@ -25,8 +25,8 @@ const (
 	EC256   = KeyType("P256")
 	EC384   = KeyType("P384")
 	RSA2048 = KeyType("2048")
+	RSA3072 = KeyType("3072")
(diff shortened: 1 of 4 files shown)


ldez:
https://github.com/go-acme/lego/issues/1854#issuecomment-1450227038

systemcrash:
Maybe just add 3072 then?

ldez:
yes just RSA 3072 is possible

ldez:
I cannot reopen the PR, I don't why, so you will have to create a new one.

ldez:
FYI I'm not a bot :wink:

systemcrash:
I figured 😉

PR #1864, opened 2023-03-07
Add flag --bupem to make a bottom-up PEM

systemcrash:
Outputs also a 'bottom-up' PEM (.bupem) file containing, in this order:

Private .key
Public .crt
Issuer .crt

Saves a manual `cat` step to build such a file. Useful for e.g. Postfix (+ >=OpenSSL 1.1.1) with its `*_chain_files` options:

```
smtpd_tls_chain_files =
  /etc/postfix/mail.example.com.rsa4096.bupem,
  /etc/postfix/mail.example.com.ec256.bupem
smtp_tls_chain_files =
  /etc/postfix/mail.example.com.rsa4096.bupem,
  /etc/postfix/mail.example.com.ec256.bupem
```

The 'bottom-up' convention expects the private key first.


systemcrash:
Any review for this?

cmd/flags.go
```diff
@@ -118,6 +118,10 @@ func CreateFlags(defaultPath string) []cli.Flag {
 			Usage: "Set the DNS timeout value to a specific value in seconds. Used only when performing authoritative name server queries.",
 			Value: 10,
 		},
+		&cli.BoolFlag{
+			Name:  "bupem",
+			Usage: "Generate an additional 'bottom-up' .bupem (PEM, base64) file by concatenating the .key, .crt and issuer .crt together, in that order.",
```
dmke, line 123: "PEM, base64" is redundant, as all PEM files encode X.509 DER serialization data as Base64.

```suggestion
			Usage: "Generate an additional 'bottom-up PEM' file (.bupem) by concatenating the private key, leaf certificate and issuer chain (in that order) into a single file.",
```

dmke (commented):
I'm not keen on that name. We should avoid making names up, as this leads to headaches. Searching for "buttom-up PEM" doesn't yield much related results, with one exception: this PR.

Postfix, confusingly, calls this format a "chain file". In terms of certificates, "chain" usually means a list of issue, intermediate and leaf certificates (in *that* order, and without private key). I'd advise using "chain PEM"/`.chainpem` as alternative.

I also don't have a useful name, at the moment.

<details><summary>Idea for another approach</summary>

Maybe the solution is not to provide hard-coded output formats, and instead allow the user to describe the format(s) they need, something akin to this:

    lego run ... \
      --cert.out bu.pem=pkey,cert,issuerchain \
      --cert.out chain.pem=issuerchain,cert \
      --cert.out key=pkey

which would yield three files:

- `domain.bu.pem`, containing the file described in this PR,
- `domain.chain.pem`, containing an ordinary bundle, and
- `domain.key`, containing just the private key.

(The general syntax being `--cert.out <extension>=<token-list>`, with predefined tokens like "pkey" for the private key).

</details>

systemcrash:
Having a customizable `cert.out` order as you suggest is nice. Although I currently cannot imagine those getting much more use since the desired constellations for certificates and their order in a file is limited. Still, more flexible than my approach.

dmke:
> Although I currently cannot imagine [`--cert.out`] getting much more use since the desired constellations for certificates and their order in a file is limited

Sure, I'd imagine it more as a general replacement for the `--pem`, `--pfx`, and `--no-bundle` CLI options (which then would be deprecated). Maybe this is something for the next major version (or an alternative CLI frontend), as there are more issues with the current CLI...

cmd/flags.go
systemcrash: Not strictly necessary - abstract is equally intelligible. First 'the' is acceptable in native English.

PR #1839, opened 2023-02-15
Add Google Domains as DNS provider

aaomidi:
Create the [Google Domains](https://domains.google) DNS provider to lego.

Google domains has launched a new product called "ACME DNS" that allows you to change TXT records that have `_acme-challenge` in the string. This PR enables the support of that to the wonderful and extremely valuable library of Lego.

Fixes #1865

PR #1866, opened 2023-03-09
gcloud: update google APIs to latest version

aaomidi:
Google Domains tests were a bit flacky and had broken between these versions. Doing this as part of preperation for #1839


ldez (approved):
LGTM

PR #1907, opened 2023-04-30
feat: add LEGO_CERT_PEM_PATH and LEGO_CERT_PFX_PATH to run hook

### LINKED ISSUE #1906: Provide PEM & PFX environmental variables to run hook as well
### Welcome

### How do you use lego?

Binary

### Detailed Description

#1609 added PEM & PFX environmental variables for the renewal hook, but not the run hook. I would find them useful in both contexts.

speshak:
Currently only the renew hook gets these variables set, this makes the behavior consistant for run hooks as well.

Fixes #1906


ldez (approved):
LGTM

PR #1908, opened 2023-05-01
doc: fix wrong env var name

cole-h:
I noticed that this env var doesn't exist anymore, as it was presumably renamed to `LEGO_CERT_KEY_PATH` as listed above.


ldez (approved):
LGTM

PR #1911, opened 2023-05-02
docs: fix `gdomains`→`googledomains`

alexzorin:
The example for the Google Domains provider refers to a `gdomains` code, but it should be `googledomains`.


ldez (approved):
LGTM

PR #1918, opened 2023-05-21
route53: pass ExternalID property to STS:AssumeRole API operation

### LINKED ISSUE #1919: route53: add ExternalID property
### Welcome

### How do you use lego?

Through Terraform ACME provider

### Detailed Description

First, I would like to say how great I find the work on the go-acme/lego and hashicorp /terraform-provider-acme projects.

I need to set the ExternalID property in addition to the AssumeRoleArn property to be allowed on the sts:AssumeRole API action.

Maybe something like #1918 could do the job ?

jvdbc:
External ID property is useful when granting access to AWS resources to a third party : https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_create_for-user_externalid.html

Fixes #1919


ldez (approved):
LGTM

PR #1937, opened 2023-06-15
fix: read status code from response

gsato-godaddy:
Currently, the badNonce error handling relies on the status field in the problem object .  
From my understanding of the Problem Object spec ([rfc7807](https://datatracker.ietf.org/doc/html/rfc7807#section-3.1)), the status field is not necessarily required.  

This change uses the response status code if the problem object does not contain the status field.

### DIFF
--- a/acme/api/internal/sender/sender.go
+++ b/acme/api/internal/sender/sender.go
@@ -133,6 +133,10 @@ func checkError(req *http.Request, resp *http.Response) error {
 		errorDetails.Method = req.Method
 		errorDetails.URL = req.URL.String()
 
+		if errorDetails.HTTPStatus == 0 {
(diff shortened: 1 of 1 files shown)


ldez:
Hello,

is your PR related to a real problem you encountered?

gsato-godaddy:
Hi, thanks for the fast reply!

Yes, I'm integrating with Godaddy's acme service. I was seeing this error:
```
acme: error: 0 :: POST :: https://acme.godaddy.com/v1/acme/new-order :: urn:ietf:params:acme:error:badNonce :: The supplied nonce has expired.
```
I figured that the `checkError` wasn't picking up the status code.

ldez (approved):
LGTM

PR #1912, opened 2023-05-03
Initial ACME Renewal Info (ARI) Implementation

beautifulentropy:
Additions to cmd/cmd_renew.go: 
- Optional flag `--ari-enable` to check the renewalInfo endpoint when evaluating whether a renewal is necessary and
- posting to the renewalInfo endpoint post-renewal to indicate that the certificate has been replaced.
- Optional flag `--ari-hash-name` to indicate the name of the hash expected by the renewalInfo endpoint (e.g. "SHA-256").
- Optional flag `--ari-wait-to-renew-duration` to indicate the duration you're willing to sleep for the renewal window suggested by the renewalInfo endpoint.

Additions to acme/api/certificate.go
- `CertificateService.GetRenewalInfo()` to support 4.1. Getting Renewal Information of draft-ietf-acme-ari-01
- `CertificateService.UpdateRenewalInfo()` to support 4.2. Updating Renewal Information of draft-ietf-acme-ari-01

Additions to acme/commons.go
- renewalInfo in Directory
- Various objects to support renewalInfo POST and GET

Additions to certificate/certificates.go
- Convenience methods to support renewalInfo Get and Update workflows in `cmd/cmd_renew.go`
- Function for deriving the CertID provided a certificate and issuer.

Fixes #1878

PR #1944, opened 2023-06-23
fix: ARI explanationURL

ldez:
I detected a variation from the RFC: the RFC says `explanationURL` instead of `explanationUrl`.

`explanationURL` follows the same naming rule (commons Go initialism) as `certID`.

https://www.ietf.org/archive/id/draft-ietf-acme-ari-01.html#section-4.1
https://www.ietf.org/archive/id/draft-ietf-acme-ari-01.html#section-4.2

Related to #1912


dmke (approved):
Good catch!

PR #1959, opened 2023-07-20
Allow to set EAB kid and hmac via environment variables

### LINKED ISSUE #1953: Allow passing EAB-Information as envvar
### Welcome

### How do you use lego?

Binary

### Detailed Description

For many usecases it is preferrable to not have secrets hardcoded e.g. in systemd-units and not have them visible in ps. Therefore I would propose to add a way to specify EAB-Information (`kid` and `hmac`) via an environment variable

(Kind of related to #1375, but different)

n0emis:
closes #1953


ldez (approved):
LGTM

Issue #1895, opened 2023-04-14
Disable date prefix in output


mattiasb:
### Welcome

- [X] Yes, I've searched similar issues on GitHub and didn't find any.

### How do you use lego?

Binary

### Detailed Description

Hi!

First I'd like to say *thanks* for lego. It's amazing and simple to use!

I'm running lego as a (systemd) service on my machines. Since the logs end up in the journal you get a timestamp for the output already legos own datetime gets redundant (and a bit noisy).

Example output from `systemctl status lego@myserver.example.com`:
```
Apr 14 16:46:30 myserver lego[19429]: 2023/04/14 16:46:30 [INFO] [myserver.example.com] acme: Could not find solver for: tls-alpn-01
Apr 14 16:46:30 myserver lego[19429]: 2023/04/14 16:46:30 [INFO] [myserver.example.com] acme: use http-01 solver
Apr 14 16:46:30 myserver lego[19429]: 2023/04/14 16:46:30 [INFO] [myserver.example.com] acme: Trying to solve HTTP-01
```

I'd prefer it to be:
```
Apr 14 16:46:30 myserver lego[19429]: [INFO] [myserver.example.com] acme: Could not find solver for: tls-alpn-01
Apr 14 16:46:30 myserver lego[19429]: [INFO] [myserver.example.com] acme: use http-01 solver
Apr 14 16:46:30 myserver lego[19429]: [INFO] [myserver.example.com] acme: Trying to solve HTTP-01
```

Thanks again for Lego! ♥

PR #1962, opened 2023-07-21
Add --log-format flag

bsd-department:
This allows logger flags to be specified through the command line.

Solves #1895


ldez:
Hello,

just to explain why your PR is stuck: I plan to rewrite the log system based on the new slog package, so I will not merge your PR because I will have to remove it when I will change the logger.

bsd-department:
I took a peek inside slog, and the default slog logger [uses the default `log` package logger to write logs](https://cs.opensource.google/go/x/exp/+/92128663:slog/logger.go;drc=06a737ee72cbcbbd5e2849214b802711d14c7a0f;l=20). `--log-format` implementation will not change much, `log.SetFlags` will be called instead.

jantytgat:
> Hello,
> 
> 
> 
> just to explain why your PR is stuck: I plan to rewrite the log system based on the new slog package, so I will not merge your PR because I will have to remove it when I will change the logger.
> 
> 

I'm working on two providers for use with NetScaler ADC, for which I'm already using slog.
Do you have any ETA on the rewrite?
Is there any way I might contribute to this to speed up an said ETA?

PR #1986, opened 2023-08-05
add warning about different fallback env var values

santschi:
A little QOL addition in showing a warning when fallback env vars are set to different values.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 .vscode/
 dist/
 builds/
+.devcontainer/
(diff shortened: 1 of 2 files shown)


ldez:
Hello,

Can you explain why do you think this is important?

Do you have a problem related to this function?

santschi:
Hey!

I ran into a situation where it turned out that there was a env var set with higher priority than the one I was changing and lingering on the system. I took a while to figure out why my configuration was not working since I kept getting error messages about api access etc.

I would have been glad about a short warning which would have saved me some time so I figured I could also just add it myself.

ldez:
I will decline because you will validate only if the main value is filled and because I think this is outside of the scope of what lego should check.

Thank you anyway for the time you've spent to contribute to lego.

PR #1988, opened 2023-08-11
azuredns: update docs

jlaundry:
- Clarify that specifying the target DNS zone's Tenant ID, Subscription ID, and Resource Group Name is required to use this provider with Managed Identities.
- Provide an example of how to set permissions on a DNS Zone so that a Managed Identity can request a certificate.
- Provide a hint for people trying to use Azure Arc's Managed Identity, that they may need to set special environment variables.

Pinging @pchanvallon for a review (and thanks for your efforts building this provider!)


ldez:
Hello,

this file is generated:
https://github.com/go-acme/lego/blob/07c4daeff3c5d5e228b2490e002295716083a5fa/docs/content/dns/zz_gen_azuredns.md?plain=1#L12-L14

Could you update the source file: https://github.com/go-acme/lego/blob/master/providers/dns/azuredns/azuredns.toml

jlaundry:
🤦🏻‍♂️ very sorry, updated now

pchanvallon (approved):
LGTM ✅

PR #1993, opened 2023-08-18
 chore: update to go1.20

### LINKED ISSUE #793: export resolver.obtainError?

ldez:
- updates CI to use go1.20 and go1.21: uses the aliases `stable` and `oldstable` to avoid updating the branch protection each time there is a Go update.
- uses `errors.Join` to improve `obtainError`

https://tip.golang.org/doc/go1.20

Fixes #793

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -11,7 +11,7 @@ jobs:
     name: Build and deploy documentation
     runs-on: ubuntu-latest
     env:
-      GO_VERSION: '1.20'
+      GO_VERSION: stable
       HUGO_VERSION: '0.117.0'
       CGO_ENABLED: 0
 
--- a/.github/workflows/go-cross.yml
+++ b/.github/workflows/go-cross.yml
@@ -20,33 +20,15 @@ jobs:
         os: [ubuntu-latest, macos-latest, windows-latest]
 
     steps:
-      # https://github.com/marketplace/actions/setup-go-environment
-      - name: Set up Go ${{ matrix.go-version }}
-        uses: actions/setup-go@v3
-        with:
-          go-version: ${{ matrix.go-version }}
-
       # https://github.com/marketplace/actions/checkout
       - name: Checkout code
         uses: actions/checkout@v3
 
-      # https://github.com/marketplace/actions/cache
-      - name: Cache Go modules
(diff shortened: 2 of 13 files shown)


dmke:
I currently don't have the mental capacity, so a full review will need to wait until tomorrow, but:

> uses the aliases `stable` and `oldstable`

Does this work with the caching mechanism?

```yaml
# https://github.com/marketplace/actions/cache
- name: Cache Go modules
  uses: actions/cache@v3
  with:
    # ...
    key: ${{ runner.os }}-${{ matrix.go-version }}-go-${{ hashFiles('**/go.sum') }}
    restore-keys: |
      ${{ runner.os }}-${{ matrix.go-version }}-go-
```

Now with `matrix.go-version` being stable, the cache key will only change if the `go.sum` files change. This would only be the case for dependency updates, since the [minimum Go version](https://go.dev/doc/modules/gomod-ref#go) is only present in the `go.mod` file...

ldez:
The modules don't depend on the Go version, so the cache can stay the same between two version of Go.

But in fact since `actions/setup-go@v4` the caching is handled by the action, so I will remove the caching step.

dmke (approved):

PR #1994, opened 2023-08-18
doc: retrieve hugo-theme-learn as hugo module

deining:
This PR changes: the way the documentation is created: `hugo-theme-learn` ist now pulled in as hugo module. This is the newer and more elegant way to use hugo theme components.

### DIFF
--- a/docs/go.mod
+++ b/docs/go.mod
@@ -1,0 +1,5 @@
+module github.com/go-acme/lego/docs
+
+go 1.20
+
+require github.com/matcornic/hugo-theme-learn v0.0.0-20211028190410-e817f53d690d // indirect
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -6,15 +6,9 @@ clean:
 	rm -rf public/
 
 
-hugo-build: clean hugo-themes
+hugo-build: clean
 	hugo --enableGitInfo --source .
 
 hugo:
 	hugo server --disableFastRender --enableGitInfo --watch --source .
 	# hugo server -D
-
-hugo-themes:
-	rm -rf themes
-	mkdir themes
-	git clone --depth=1 https://github.com/matcornic/hugo-theme-learn.git themes/hugo-theme-learn
-	rm -rf themes/hugo-theme-learn/.git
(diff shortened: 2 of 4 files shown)


ldez:
Hello,

it was a choice to not use the hugo module system.

deining:
> Hello,
> 
> it was a choice to not use the hugo module system, when I added the theme the hugo module system was already existing.

I see. I respect your decision.
Out of interest: may I ask why you deliberately didn't use the hugo module system?

ldez:
Even if I really appreciate Hugo, I encounter several problems.
The [zerover](https://0ver.org/) approach is a problem.
I was a long-time user and at some point, I was tired of breaking changes and the choice of standard or extended version.

The hugo module system is pretty much the same thing.
The git submodule was a bad solution for me, and I never use it.
When the hugo module appears I was thinking that was a solution but I "fought" against themes that have some Go files in them.
At this time, the Go modules were new, and hugo modules too.
It was not a good experience so I decided to use a simple `git clone`.

But maybe it's time to evaluate again the hugo modules.

---

I have a question completely off-topic: you feel like a "Hugo evangelist", and I need to replace the way to build the documentation for another opensource project (but I need to keep the style and several other things), are you interested to help me on that?

ldez:
There is nothing private, it is just a question to know if it's something possible.
I need to talk with the maintainers of this project before.

deining:
> There is nothing private 😄, it is just a question to know if it's something possible.

Yes it is. No promise, though.

> I need to talk with the maintainers of this project before.

Keep me informed.

deining:
> But maybe it's time to evaluate again the hugo modules.

After bumping to latest hugo version, checkers are green now :smile:.

docs/go.mod
```diff
@@ -0,0 +1,5 @@
+module github.com/go-acme/lego/docs
+
+go 1.21
```
ldez, line 3:
```suggestion
go 1.20
```
Since go1.21, it is a hard constraint (blocking minimum go version) so it's better to use go1.20.

ldez (changes requested):

docs/go.mod
deining: Changed as requested.

ldez (approved):
LGTM

Issue #2016, opened 2023-09-08
PowerDNS PDNS_API_URL API version issue


m0t1x:
### Welcome

- [X] Yes, I'm using a binary release within 2 latest releases.
- [X] Yes, I've searched similar issues on GitHub and didn't find any.
- [X] Yes, I've included all information below (version, config, etc).

### What did you expect to see?

Would lilke that lego accepts API version in PDNS_API_URL or to detect the version properly.

### What did you see instead?

When  'PDNS_API_URL' is set to 'http://server/', without 'api/v1' prefix, API version is not detected and all pdns requests are made to `/` instead to `/api/v1` so all PowerDNS requests fails.

When  'PDNS_API_URL' is set to 'http://server/api/v1' all pdns requests are made to /api/v1/api/v1. So version prefix is duplicated and again all PowerDNS request fails.

Why not just allow API version in URL?

Another project noticed the same issue - https://github.com/traefik/traefik/issues/9726

### How do you use lego?

Binary

### Reproduction steps

1. Install lego
2. Configure PDNS_API_KEY and PDNS_API_URL with http://server or with http://server/api/v1 
3. Run 
```
lego -a \                                             
-d <domain> \
-m <e-mail> \
--server <ACME server if used > \
--dns pdns \
renew
```

### Version of lego

```console
lego version 4.14.0 darwin/arm64
```

### Logs

<details>

`PDNS_API_URL=https://< server >/api/v1`

```console
...
<title>Error</title>
</head>
<body>
<pre>Cannot GET /api/v1/api</pre>
</body>
</html>
...
```

`PDNS_API_URL=https://< server >`

```console
...
<title>Error</title>
</head>
<body>
<pre>Cannot GET /api</pre>
</body>
</html>
...
```

</details>

ldez:
Hello,

I explained the behavior here: https://github.com/traefik/traefik/issues/9726#issuecomment-1432329197

> Why not just allow API version in URL?

Because there are several use-cases depending on users.

PR #2019, opened 2023-09-19
pdns: optional custom API version

m0t1x:
Some PowerDNS servers don't have exposed the [root API endpoints](https://doc.powerdns.com/recursor/common/api/endpoint-api.html) and PowerDNS API version auto-detection fails. Use the optional parameter `PDNS_CUSTOM_API_VERSION` to set the version number in such situations.

related to  #2016

### DIFF
--- a/cmd/zz_gen_cmd_dnshelp.go
+++ b/cmd/zz_gen_cmd_dnshelp.go
@@ -2078,7 +2078,7 @@ func displayDNSHelp(w io.Writer, name string) error {
 		ew.writeln()
 
 		ew.writeln(`Additional Configuration:`)
-		ew.writeln(`	- "PDNS_CUSTOM_API_VERSION":	Skip API version autodetection and use custom version number.`)
+		ew.writeln(`	- "PDNS_API_VERSION":	Skip API version autodetection and use the provided version number.`)
 		ew.writeln(`	- "PDNS_HTTP_TIMEOUT":	API request timeout`)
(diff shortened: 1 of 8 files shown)


ldez:
Do we agree that I close the issue and flag it as "question" and you still created a PR?
What exactly do you expect from this behavior?

m0t1x:
Hi,
As described in the PR, PowerDNS has root API endpoints to provide the version information (https://doc.powerdns.com/recursor/common/api/endpoint-api.html). Some PowerDNS deployment don't have these endpoints exposed and without them it is not possible to auto-detect PowerDNS API version and the whole process with lego fails. Using custom variable to set the API version, you can skip the API version auto-detection and the process will continue.
I found I am not the only one with this issue, but it's up to you to see whether you want to have this feature or not.
Best Regards,Marijan




    On Wednesday, October 25, 2023 at 01:15:58 AM GMT+2, Ludovic Fernandez ***@***.***> wrote:  
 
 


Do we agree that I close the issue and flag it as "question" and you still created a PR?
What exactly do you expect from this behavior?

7c:
I agree, powerdns 4.* does not have this endpoint, so i had to hack into the code in order to set the apiVersion to 1

ldez (approved):
Even if I disagree with the fact of opening a PR when an issue was closed, I will merge this PR anyway.
I hope this will not happen again and discussion will happen inside the issue BEFORE opening a PR.

PR #2022, opened 2023-09-20
ipv64: remove unused option

ldez:
`IPV64_SEQUENCE_INTERVAL` is unused.

### DIFF
--- a/providers/dns/ipv64/ipv64.go
+++ b/providers/dns/ipv64/ipv64.go
@@ -24,6 +24,7 @@ const (
 	EnvPropagationTimeout = envNamespace + "PROPAGATION_TIMEOUT"
 	EnvPollingInterval    = envNamespace + "POLLING_INTERVAL"
 	EnvHTTPTimeout        = envNamespace + "HTTP_TIMEOUT"
+	EnvSequenceInterval   = envNamespace + "SEQUENCE_INTERVAL" // Deprecated: unused, will be removed in v5.
 )
 
 // Config is used to configure the creation of the DNSProvider.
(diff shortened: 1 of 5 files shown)


providers/dns/ipv64/ipv64.go
```diff
@@ -24,15 +24,13 @@ const (
...
 	EnvHTTPTimeout        = envNamespace + "HTTP_TIMEOUT"
-	EnvSequenceInterval   = envNamespace + "SEQUENCE_INTERVAL"
 )
 
 // Config is used to configure the creation of the DNSProvider.
 type Config struct {
 	APIKey             string
 	PropagationTimeout time.Duration
 	PollingInterval    time.Duration
-	SequenceInterval   time.Duration
```
dmke, line 35: Doesn't this change make it a breaking change? `Config` is exported, so removing fields could affect Lego-as-library users.

For now, I'd just add a comment and mark it for removal in v5, e.g.

```diff
-	SequenceInterval   time.Duration
+ 	SequenceInterval   time.Duration // Deprecated and not used. Will be removed in Lego v5.
```

dmke (changes requested):

providers/dns/ipv64/ipv64.go
ldez: There is no breaking change because the option is not used and the option was introduced in v4.13.

ldez: If we keep and deprecate the field `SequenceInterval` we also have to keep and deprecate the env var name (constant).

dmke: Hm, true... So do we remove it (and deal with the fallout later, if there's any), or do we keep them with a deprecation note?

dmke (approved):

providers/dns/ipv64/ipv64.go
ldez: We can keep them, it's better from the point of view of the project.

PR #2026, opened 2023-09-27
azuredns: provide the ability to select authentication methods

pchanvallon:
Hello,

After using `azuredns` new DNS provider, I figured out it could be really useful to manage which authentication methods to use, in order to configure more accurately the expected behavior.

Therefore, I have splited the `DefaultAzureCredential` by using the different Azure credential methods and added the possibility to disable some of them.

Note that for `NewManagedIdentityCredential` I added a timeout management as described in the [azidentity package documentation](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#example-NewChainedTokenCredential-ManagedIdentityTimeout)

Let me know if you want me to add anything.
Thanks.

### DIFF
--- a/cmd/zz_gen_cmd_dnshelp.go
+++ b/cmd/zz_gen_cmd_dnshelp.go
@@ -307,6 +307,7 @@ func displayDNSHelp(w io.Writer, name string) error {
 		ew.writeln()
 
 		ew.writeln(`Credentials:`)
+		ew.writeln(`	- "AZURE_CLIENT_CERTIFICATE_PATH":	Client certificate path`)
 		ew.writeln(`	- "AZURE_CLIENT_ID":	Client ID`)
 		ew.writeln(`	- "AZURE_CLIENT_SECRET":	Client secret`)
 		ew.writeln(`	- "AZURE_RESOURCE_GROUP":	DNS zone resource group`)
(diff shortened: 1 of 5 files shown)


ldez:
Hi, what is the value of disabling some auth methods instead of choosing only one?

pchanvallon:
Hello, 
The purpose is to avoid breaking changes while introducing this feature.
The behavior of `DefaultAzureCredentials` is to provide a builtin fallback mechanism between the different authentication methods. By enabling all of them by default, we are keeping the behavior as is and provide the ability to configure which one should be used.
The other way round is also possible but it will introduce some changes in terms of configuration.

ldez:
We can keep the default and choose one with only one option.

pchanvallon:
Hello, I have updated the PR according to your suggestion.

ldez:
I think you did not understand my suggestion: I was speaking about 1 option instead of 5 options, and not changing the current lego default.

pchanvallon:
Hello,
The goal of using different variables was to mimic the [`azurerm` terraform provider configuration](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs#use_msi).
I have no cons to use only one variable as we still have the default credential method as fallback.
Therefore I updated the PR accordingly.

ldez:
You can mimic the terraform provider but I need to know if being able to use several auth methods is a real thing or not, it's not a pro/con.

I need more concrete arguments to justify if auth method fallback is a thing for Azure users.

pchanvallon:
The fallback mechanism can be useful during the development phase on a local machine. The `DefaultAzureCredentials` purpose was made for that purpose as described in the [Azure SDK documentation](https://learn.microsoft.com/en-us/azure/developer/go/azure-sdk-authentication?tabs=bash): 
> DefaultAzureCredential simplifies authentication by combining commonly used credential types. It chains credential types used to authenticate Azure-deployed applications with credential types used to authenticate in a development environment.

In a production environment for a workload or CI/CD usage, defining one and only one authentication method is viable because it is determined by the context of the workload or CI/CD environment.

ldez (approved):

PR #2007, opened 2023-09-02
gandiv5: add Personal Access Token support

ldez:
Closes #2006

https://api.gandi.net/docs/livedns/

PR #2029, opened 2023-10-06
gandiv5: Add "Bearer" prefix to the auth header

Christoph-D:
The "Bearer" prefix is required per
https://api.gandi.net/docs/authentication/ to declare the content of the Authorization header as a personal access token.

Without this change, the user has to specify `GANDIV5_PERSONAL_ACCESS_TOKEN='Bearer 12341234'` instead of `GANDIV5_PERSONAL_ACCESS_TOKEN=12341234` which disagrees with the documentation and the existing test data.

This change is not backwards-compatible. This should be OK because the commit 113648a36817a5d9667b8dd1f1c6c67eeb914916 (#2007) which added support for personal access tokens was not in any release yet.


ldez (approved):
LGTM, Thank you.

PR #2031, opened 2023-10-11
liquidweb: detect zone automatically

jakdept:
I work for LiquidWeb, and we'd like to get some updates into our provider in lego to simplify things a bit and change the API endpoint.

I've followed the guidelines at https://github.com/go-acme/lego/blob/2140e6befe2a46267c88c735a40f4791fe6036ab/CONTRIBUTING.md and will open an issue as well once I've got this open.

I'm skipping including `make test` cause of the long output, but it passes as well.

```
╰─> make fmt
make: Nothing to be done for `fmt'.
┬─{ jack:~/s/g/j/lego git:(refresh)
╰─> make checks
golangci-lint run
╰─> make build
BIN_OUTPUT: dist/lego
rm -rf dist/ builds/ cover.out
Version: a99af0b0e7dd2c9e01c62a129e21e1fad9732a56
go build -trimpath -ldflags '-X "main.version=a99af0b0e7dd2c9e01c62a129e21e1fad9732a56"' -o  dist/lego ./cmd/lego/
```

I'm guessing you'll squash my commits, which is fine if needed. Or I can submit another and squash them myself if needed.


ldez:
Hello,

Why did you change the prefix?

ldez:
All the PR should be rewritten to:
- follow the same design as the other providers.
- don't create regression (the exposed elements are a part of the lego API).

I will rewrite it.

jakdept:
I changed the prefix so it matches our terraform provider and upcoming prefix for the CLI and other tooling. However, I changed it in a way so that the old prefix still works. I believe I have tests showing that as well.

I did not see anything specifying a standard around the prefix. If that's an issue, we can leave it as `LIQUID_WEB_` and I can work around it, but it's going to differ from all other tooling we provide.

ldez:
I will merge the PR without the `LWAPI_`.

I will extend the system I created to handle fallback env vars for the cloudflare provider to be usable in this context.

I will create a PR with the support of `LWAPI_` with the right fallback system.

ldez (approved):
LGTM

PR #2051, opened 2023-11-10
regru: HTTP method changed to POST

boiler:
http method GET is deprecated
https://www.reg.ru/reseller/api2doc#common_connection

so, we need to change it to POST


ldez (approved):
LGTM

PR #2055, opened 2023-11-17
route53: Update IAM policy example to remove route53:ListHostedZonesByName

os11k:
Seems `route53:ListHostedZonesByName` is not needed if we specify zoneID for `ListResourceRecordSets`. I got this idea from here:

https://cert-manager.io/docs/configuration/acme/dns01/route53/#set-up-an-iam-role

And I tested today, and it does provides certificate without `route53:ListHostedZonesByName`.


os11k:
Actually I tried now on my other AWS account and it doesn't work without `route53:ListHostedZonesByName`. In that case I had separate hosted zone for domain... I will close this now.

Issue #277, opened 2016-08-23
register without email


streaps:
Can lego register an account without an email address?

It is possible with other tools (letsencrypt.sh) and I don't have any need to share my email address with letsencrypt. see https://community.letsencrypt.org/t/email-address-disclosures-june-11-2016/17025

mholt:
The underlying acme package can, but the command doesn't currently allow that.

hairyhenderson:
@mholt is there any reason why `lego` _shouldn't_ allow registering without an e-mail address?

mholt:
Other than it being strongly, strongly discouraged? Not that I know of, no. @xenolf would know better though.

hairyhenderson:
Thanks @mholt!

xenolf:
Yeah like @mholt said, there is no other reason then it being discouraged. I'm not sure why you would like to specifically not have a recoverable account.

hairyhenderson:
@xenolf It's not a _strong_ use-case, to be clear - more of a vague idea. My main use-case is for developers who set up temporary sites that usually live for 2-3 days (7 tops), and it'd be _slightly_ inconvenient for them to have to specify an e-mail address when there's never really a need to renew. Right now that scenario involves self-signed certificates, which are increasingly annoying to deal with.

Mostly I'm not 100% clear on exactly what LE/ACME _does_ with the e-mail address - I'll do some more reading and figure it out.

smtalk:
@hairyhenderson similar needs here, we just don't know real email addresses on the systems we're auto-deploying the certs. On the ones we 'known' them - some certs are just temporary ones or different sets, and they get flooded with expiring certificate notices then (they're confusing, as they include a different set of SANs). Having ability not to specify the email would be really great!

smtalk:
@ldez just wondering if there is a chance to see this included into the next release? We'd replace our own LE client with lego on all the installations of DirectAdmin, that's the only thing left.. :)

rafaelgieschke:
I tried to implement this in <https://github.com/go-acme/lego/issues/277>.

ldez:
I don't really understand the need:
- if you are a company, it's highly recommended to use an email, like that if you have a rate limit problem Let's Encrypt will be able to increase your limit for this email
- you can use an email alias to catch all those emails, and "hide" them
- Let's Encrypt discouraged to do that

I don't want to change the default behavior, that promotes the use of an email like Let's Encrypt recommends, without a strong use case.

[...]

PR #2059, opened 2023-12-05
when eab is used, use the userID of "eab"

pschou:
When using an External Account Binding for requesting certs, the email field does not apply.  In this case let's use a generic string of "eab" for the userID and allow -eab to be specified on the command line.


ldez:
Related to #277

I prefer to handle the problem behind that instead of creating a specific behavior.

PR #2061, opened 2023-12-05
feat(cli): add environment variable for specifying alternate directory URL

pschou:
When using a different corporate ACME endpoint, allow the environment variable to be set so it doesn't have to be an arg on each invocation.


ldez (approved):
LGTM

PR #2062, opened 2023-12-06
add option to register only for an eab setup

pschou:
When using lego in an EAB infrastructure, having the ability to test the registration first will help ensure that the eab is correct before going on with creating an ACME request for a certificate.


ldez:
This implementation is a little too naive to work effectively.
I will add working `register` command in lego v5.

Related to #2798

PR #2063, opened 2023-12-06
cli: add format option for PFX encoding

pschou:
The created pfx file uses an older encryption method, this adds a --pfx.format while keeping backward compatibility.


ldez (approved):
LGTM

PR #2065, opened 2023-12-06
Remove the requirement for a challenge when --eab is used

pschou:
This handles the case where EAB is the only method for verification.


ldez:
EAB is not a challenge

PR #2066, opened 2023-12-06
Update CertID format as per draft-ietf-acme-ari-02

beautifulentropy:
- Implemented the updated CertID format as specified in draft-ietf-acme-ari-02.
- Removed the `--ari-hash-name` flag, which is no longer necessary with the updated format.
- Enabled ARI usage for users with `--no-bundle` option. Previously, this was not possible as the draft-ietf-acme-ari-01 format required metadata available only in the issuer.

---

ATTENTION: this must NOT be merged and should remain in draft until https://github.com/letsencrypt/boulder/pull/7184 is merged and deployed.

UPDATE: this change has been merged, but it will not be deployed to Production until a few days into the new year. I'll update this issue when that's the case.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -236,4 +236,4 @@ issues:
     - path: providers/dns/hosttech/internal/client_test.go
       text: 'Duplicate words \(0\) found'
     - path: cmd/cmd_renew.go
-      text: 'cyclomatic complexity 16 of func `renewForDomains` is high'
+      text: 'cyclomatic complexity 15 of func `renewForDomains` is high'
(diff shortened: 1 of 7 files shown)


beautifulentropy:
Failing test:

```
Error: cmd/cmd_renew.go:128:1: cyclomatic complexity 15 of func `renewForDomains` is high (> 12) (gocyclo)
func renewForDomains(ctx *cli.Context, client *lego.Client, certsStorage *CertificatesStorage, bundle bool, meta map[string]string) error {
^
make: *** [Makefile:36: checks] Error 1
Error: Process completed with exit code 2.
```

ldez:
for now, you can ignore the linter about cyclomatic complexity, I will look at the best way to handle that.

ldez:
I fixed the problem of cyclomatic complexity.
Just waiting for the deployment to Production.

beautifulentropy:
> I fixed the problem of cyclomatic complexity. Just waiting for the deployment to Production.

Thanks for fixing this. Just checked our SRE team and support for this portion of draft-ietf-acme-ari-02 has been successfully deployed to both staging and production.

ldez:
Maybe you can set the PR as "Ready for review"?

ldez (approved):
LGTM

Note: this change is breaking but it's expected of a draft. So we will not create a major version only for this change.

PR #2076, opened 2024-01-09
cloudru: change default URLs

danilsolovyov:
Changes:
- APIBaseURL changed: The variable has been changed because I was getting the RBAC access denied error. Technical support cloud.ru gave me an answer: "Danil, good afternoon! There is an error in our documentation, use the host console.cloud.ru instead of console.sbercloud.ru to send a request to the API." With the host console.cloud.ru the error is gone.

- AuthBaseURL changed: The old variable works too. But I think that soon they will completely migrate the host API from sbercloud.ru on cloud.ru .

### DIFF
--- a/providers/dns/cloudru/internal/client.go
+++ b/providers/dns/cloudru/internal/client.go
@@ -16,8 +16,8 @@ import (
 
 // Default API endpoints.
 const (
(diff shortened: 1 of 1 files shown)


ldez (approved):
LGTM

PR #2084, opened 2024-01-16
inwx: wait before generating new TOTP TANs

### LINKED ISSUE #1608: INWX: Issue when using 2FA - Same 2FA code is used twice when Present returns non-error
### Welcome

i should've seen that lego is presenting letsencrypt the dns challenge and that the challenge gets removed.

The removal does not work though which results in not downloading the certificate.

### What did you see instead?

Output:
```
cleaning up failed: inwx: (2200) Authentication error. Reason: (INVALID) The confirmation code is invalid.
```
Unless you are lucky and your `Present` is called in one 30s timeframe and your `CleanUp` is called in the next 30s timeframe the same 2FA code gets used twice (which is not a valid case for 2FA and most systems seem to implement it correctly).

### How do you use lego?

Docker image

### Reproduction steps

* have a inwx account
* enable 2fa
* try to get a wildcard certificate:
```
docker run -v certs:/lego -w /lego -e INWX_USERNAME=a -e INWX_PASSWORD=b -e INWX_SHARED_SECRET=c  goacme/lego:v4.6.0 -m your-mail@gmail.com --dns inwx -d example.de -d '*.example.de' -a --path /lego --pem run
```

### Version of lego

```console
4.6.0
```

### Logs

<details>

```console
2022/03/20 17:34:21 [INFO] [example.de, *.example.de] acme: Obtaining bundled SAN certificate
2022/03/20 17:34:22 [INFO] [*.example.de] AuthURL: https://acme-v02.api.letsencrypt.org/acme/authz-v3/12345
2022/03/20 17:34:22 [INFO] [example.de] AuthURL: https://acme-v02.api.letsencrypt.org/acme/authz-v3/12345
2022/03/20 17:34:22 [INFO] [example.de] acme: authorization already valid; skipping challenge
2022/03/20 17:34:22 [INFO] [*.example.de] acme: use dns-01 solver
2022/03/20 17:34:22 [INFO] [*.example.de] acme: Preparing to solve DNS-01
2022/03/20 17:34:26 [INFO] [*.example.de] acme: Cleaning DNS-01 challenge
2022/03/20 17:34:29 [WARN] [*.example.de] acme: cleaning up failed: inwx: (2200) Authentication error. Reason: (INVALID) The confirmation code is invalid.
2022/03/20 17:34:30 [INFO] Skipping deactivating of valid auth: https://acme-v02.api.letsencrypt.org/acme/authz-v3/12345
2022/03/20 17:34:30 [INFO] Deactivating auth: https://acme-v02.api.letsencrypt.org/acme/authz-v3/12345
2022/03/20 17:34:30 Could not obtain certificates:
[*.example.de] [*.example.de] acme: error presenting token: inwx: (2200) Authentication error. Reason: (INVALID) The confirmation code is invalid.
```

</details>

_No response_

gnoack:
This is a workaround for #1608.  INWX forbids to re-use the same TOTP twice, but the INWX DNS provider tries to reauthenticate from scratch on each step.

I believe that this is not easily implementable with the existing Lego DNS provider interface, so to avoid refactoring that interaction, let's just make the INWX provider wait a bit until a new token is available.  A new token is available every 30 seconds.

The current workaround is to invoke Lego many more times.  Retrying at a higher level is worse than retrying here.

Fixes #1608

### DIFF
--- a/providers/dns/inwx/inwx.go
+++ b/providers/dns/inwx/inwx.go
@@ -51,8 +51,9 @@ func NewDefaultConfig() *Config {
 
 // DNSProvider implements the challenge.Provider interface.
 type DNSProvider struct {
-	config *Config
-	client *goinwx.Client
+	config         *Config
+	client         *goinwx.Client
+	prevTANCounter int64
 }
 
 // NewDNSProvider returns a DNSProvider instance configured for Dyn DNS.
@@ -202,10 +203,28 @@ func (d *DNSProvider) twoFactorAuth(info *goinwx.LoginResponse) error {
 		return errors.New("two-factor authentication but no shared secret is given")
 	}
 
+	// INWX forbids re-authentication with a previously used TAN.
+	// To avoid using the same TAN twice, we wait until the next
+	// TOTP period and retry.
+	const s = 30 // seconds per TOTP period
+	now := time.Now()
+	tanCounter := now.Unix() / s
+	if tanCounter == d.prevTANCounter {
+		nextPeriod := time.Unix(s*(now.Unix()/s)+s, 0)
+		d := nextPeriod.Sub(now)
+
+		log.Infof("inwx: waiting %v for next TOTP token", d)
+		time.Sleep(d)
+
+		now = time.Now()
+		tanCounter = now.Unix() / s
+	}
+
 	tan, err := totp.GenerateCode(d.config.SharedSecret, time.Now())
 	if err != nil {
 		return err
 	}
+	d.prevTANCounter = tanCounter
 
 	return d.client.Account.Unlock(tan)
 }
--- a/providers/dns/inwx/inwx.go
+++ b/providers/dns/inwx/inwx.go
@@ -51,9 +51,9 @@ func NewDefaultConfig() *Config {
 
 // DNSProvider implements the challenge.Provider interface.
 type DNSProvider struct {
-	config       *Config
-	client       *goinwx.Client
-	previousCall time.Time
+	config         *Config
+	client         *goinwx.Client
+	previousUnlock time.Time
 }
 
 // NewDNSProvider returns a DNSProvider instance configured for Dyn DNS.
@@ -204,31 +204,29 @@ func (d *DNSProvider) twoFactorAuth(info *goinwx.LoginResponse) error {
 	}
 
 	// INWX forbids re-authentication with a previously used TAN.
-	// To avoid using the same TAN twice, we wait until the next TOTP period and retry.
+	// To avoid using the same TAN twice, we wait until the next TOTP period.
 	sleep := d.computeSleep(time.Now())
 	if sleep != 0 {
 		log.Infof("inwx: waiting %s for next TOTP token", sleep)
 		time.Sleep(sleep)
 	}
 
-	now := time.Now()
-
-	tan, err := totp.GenerateCode(d.config.SharedSecret, now)
+	tan, err := totp.GenerateCode(d.config.SharedSecret, time.Now())
 	if err != nil {
 		return err
 	}
 
-	d.previousCall = now
+	d.previousUnlock = time.Now()
 
 	return d.client.Account.Unlock(tan)
 }
 
 func (d *DNSProvider) computeSleep(now time.Time) time.Duration {
-	if d.previousCall.IsZero() {
+	if d.previousUnlock.IsZero() {
 		return 0 * time.Second
 	}
 
-	endPeriod := d.previousCall.Add(30 * time.Second)
+	endPeriod := d.previousUnlock.Add(30 * time.Second)
 	if endPeriod.After(now) {
 		return endPeriod.Sub(now)
 	}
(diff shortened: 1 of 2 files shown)


ldez (commented):
INWX API is weird.

Maybe we can decrease the 30s by reducing the validity period of the TOTP.

```go
	tan, err := totp.GenerateCodeCustom(d.config.SharedSecret, time.Now(), totp.ValidateOpts{
		Period:    10, // <-- the default is 30s
		Skew:      1,
		Digits:    otp.DigitsSix,
		Algorithm: otp.AlgorithmSHA1,
	})
```

gnoack:
> INWX API is weird.
> 
> Maybe we can decrease the 30s by reducing the validity period of the TOTP.
> 
> ```go
> 	tan, err := totp.GenerateCodeCustom(d.config.SharedSecret, time.Now(), totp.ValidateOpts{
> 		Period:    10, // <-- the default is 30s
> 		Skew:      1,
> 		Digits:    otp.DigitsSix,
> 		Algorithm: otp.AlgorithmSHA1,
> 	})
> ```

This is a good idea, but unfortunately not possible.  The period is a system parameter, which the client ("prover") and server ("verifier") must agree on.  The TOTP QR code can specify the period, or it will be 30 seconds by default.

More mathematically speaking, the "counter" is derived by dividing the current Unix time in seconds by the period in seconds, rounding down.  If we made the period 10 instead of 30, we'd end up with counter values that are three times as high (and generate tokens which are valid roughly at epoch+3*(now-epoch), where epoch is Unix time 0).

So unfortunately, this idea does not work.

(Fun fact on the side, as I've been poking around in the RFC... it seems that by rejecting the second authentication with the same token value, INWX is actually following the RFC recommendation in the last paragraph of [RFC 6238, section 5.2](https://datatracker.ietf.org/doc/html/rfc6238#section-5.2).)

ldez (approved):
The fact of not being able to reuse the token is expected because it can be a major security issue.
This problem was my fault when I introduced the TOTP system, sometimes you were focusing on something and completely forgot the obvious :smile: 

I simplified the implementation and added tests to ease the maintenance.

Thank you for your work on this topic, your fix will benefit many users.

Offtopic: I closed (and reopened) the PR by error because I wanted to write something on another screen but I was still inside this page... (GitHub keyboard shortcuts + multiple screens = mistake)

providers/dns/inwx/inwx.go
```diff
@@ -204,27 +204,34 @@ func (d *DNSProvider) twoFactorAuth(info *goinwx.LoginResponse) error {
...
-	d.prevTANCounter = tanCounter
+
+	d.previousCall = now
 
 	return d.client.Account.Unlock(tan)
 }
+
+func (d *DNSProvider) computeSleep(now time.Time) time.Duration {
+	if d.previousCall.IsZero() {
+		return 0 * time.Second
```
gnoack, line 226: I believe you could just write `return 0` here.  Integer literals can stand for any integer type, IIRC.

providers/dns/inwx/inwx.go
```diff
@@ -204,27 +204,34 @@ func (d *DNSProvider) twoFactorAuth(info *goinwx.LoginResponse) error {
...
+	}
+
+	endPeriod := d.previousCall.Add(30 * time.Second)
+	if endPeriod.After(now) {
+		return endPeriod.Sub(now)
+	}
+
+	return 0 * time.Second
```
gnoack, line 234: ditto

providers/dns/inwx/inwx.go
ldez: Yes, I know it is useless, it was just to use an explicit value as a kind of documentation.

providers/dns/inwx/inwx.go
```diff
@@ -204,27 +204,34 @@ func (d *DNSProvider) twoFactorAuth(info *goinwx.LoginResponse) error {
...
 
 	return d.client.Account.Unlock(tan)
 }
+
+func (d *DNSProvider) computeSleep(now time.Time) time.Duration {
+	if d.previousCall.IsZero() {
+		return 0 * time.Second
+	}
+
+	endPeriod := d.previousCall.Add(30 * time.Second)
```
gnoack: While this works, I suspect we are wasting a few seconds here on average, because d.previousCall is never rooted to be a multiple of 30 seconds after the epoch (Unix time 0)?

It works though and how relevant this is depends a bit on the distance between ajacent TOTP attempts.

gnoack:
Thank you @ldez for merging this - much appreciated, also for writing the tests, which I missed :)

I had a look at your patches and I think they work.  I left some minor comments - I think it is slowing the wait time down a few seconds more than the previous patch did, but I'll happily stand corrected if I am wrong here.  But it works anyway, and is an improvement over what this DNS provider did before.

providers/dns/inwx/inwx.go
ldez: You are right, sorry I completely missed that.
Sorry, I will try to fix that.
I will truncate the previous time by 30 seconds.

PR #2089, opened 2024-01-24
desec: increase default propagation interval

### LINKED ISSUE #2072: desec: increase default `DESEC_PROPAGATION_TIMEOUT` by 60s
### Welcome

### How do you use lego?

Binary

### Detailed Description

Due to an upstream PowerDNS bug, deSEC appears to be polling propagations once a minute which results in the default configuration hitting the timeout almost every time which is a bad default experience.

See https://talk.desec.io/t/global-record-propagation-issues/332/ for context.

Would it be possible to increase the default timeout as a workaround? 120s would probably be a good value.

I'm not at all familiar with Go or else I would have opened a PR but perhaps something like

```Go
 // Propagation can currently take upwards of a minute: https://talk.desec.io/t/global-record-propagation-issues/332/
 PropagationTimeout: env.GetOrDefaultSecond(EnvPropagationTimeout, dns01.DefaultPropagationTimeout + 60 * time.second)
```

in here

https://github.com/go-acme/lego/blob/7186ebb6f194c55781432162a47ff62a0bb21023/providers/dns/desec/desec.go#L46-L46

would do it.

ldez:
Closes #2072

### DIFF
--- a/providers/dns/desec/desec.go
+++ b/providers/dns/desec/desec.go
@@ -43,8 +43,8 @@ type Config struct {
 func NewDefaultConfig() *Config {
 	return &Config{
 		TTL:                env.GetOrDefaultInt(EnvTTL, defaultTTL),
-		PropagationTimeout: env.GetOrDefaultSecond(EnvPropagationTimeout, dns01.DefaultPropagationTimeout),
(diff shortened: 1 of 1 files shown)


dmke (approved):

Atemu:
Thanks!

PR #2094, opened 2024-01-30
docs: add installation for Gentoo

rahilarious:


rahilarious:
Thank you for quick look. Replace "You can" with "User can"?

ldez (approved):

PR #2105, opened 2024-02-08
Removed Brandit

Gaardsholt:
[Brandit](https://brandit.com/) is now part of [Abion](https://abion.com/) which makes the Brandit dns provider useless.
A new pull request for Abion should be created.

@Andrei-Predoiu and @ChristianGroentved can confirm this.


ldez:
Hello,

but even if Brandit is now a part of another company, some current customers can still use this provider.

I think you should provide a link to something that says that Brandit API is not usable.

In all cases, the provider will not be removed but deprecated.

ldez:
Are you a customer or an employee of Brandit/Albion?
If yes, do you have information about the API?
If yes, are you able to create or test an implementation?

Gaardsholt:
I am a customer of Brandit/Abion.
I had a quick look at the API, and it is completely different from the Brandit API, so a re-write of the provider is necessary.

We at least can't use the old API anymore, and it is my understanding that every customer of Brandit has been moved to Abion.

ldez:
The Brandit will not be written, we will just deprecate it, and create a new implementation for Albion.

Can you provide a link to the Abion API?
Can you open a new [issue](https://github.com/go-acme/lego/issues/new?assignees=&labels=enhancement%2Cnew-provider&projects=&template=new_dns_provider.yml&title=Support+for+provider%3A+%3Cput+the+name+of+your+provider%3E)?

Gaardsholt:
I will see if I can find the API, and I can create the issue if you want me to.
But we are discussing internally if we want to continue using Abion or move it somewhere else.

ldez:
If you can find the API doc and open an issue it can be great.

If you find the API doc, maybe I can work with you to create an implementation before you leave Albion.

Gaardsholt:
Here you go: https://github.com/go-acme/lego/issues/2111

I guess we can go ahead a close this pull request then?

PR #2109, opened 2024-02-10
feat: improve errors and logs related to DNS call

### LINKED ISSUE #2097: Misleading log output
### Welcome

Would expect to see something like:

```
2024/01/31 16:15:28 [INFO] [*.example.com] acme: Checking DNS record propagation using the authoritive nameservers and 1.1.1.1...
```

as after printing this log line, we're calling the `preCheck.call()` func, which in turn calls the `checkDNSPropagation` where only the INITIAL dnsQuery is sent to the recursive nameservers. if that is good, we're checking the authoritive nameservers but this part is missing from the log line and may be misleading and lets the user believe only 1.1.1.1 is needed for a proper dns propagation..

### What did you see instead?

lego prints a line in the logs like this:

```
2024/01/31 16:15:28 [INFO] [*.example.com] acme: Checking DNS record propagation using [1.1.1.1:53 8.8.8.8:53 9.9.9.9:53]
```

### How do you use lego?

Library

### Reproduction steps

1. Use the Library example code
2. see the log

### Version of lego

```console
v4.15.0
```

### Logs

see expected output

_No response_

ldez:
I improved errors and logs related to DNS calls: better information related to the context (FQDN, zone, NS, DNS message, ...)

Also, with the new error type, every error returned by `sendDNSQuery` (and `dnsQuery`) will contain the questions (ex: `example.com. IN  SOA`), so it will be far easier to diagnose DNS call problems.

Otherwise, when we iterate nameservers to do DNS calls if the latest call returns an error then we report all the errors of the previous calls instead of only the latest error.

Also, I added the FDQN inside the error message of `FindZoneByFqdn`, `FindZoneByFqdnCustom`, and `FindPrimaryNsByFqdnCustom`, so we can remove that from the error handling of the callers.

The log message:
```
Checking DNS record propagation using [1.1.1.1:53 8.8.8.8:53 9.9.9.9:53]
```
is now:
```
Checking DNS record propagation. [nameservers=1.1.1.1:53 8.8.8.8:53 9.9.9.9:53]
```

The previous message was not exactly wrong but it could be misinterpreted.
Now the nameservers are just a piece of context information.

Fixes #2097

### DIFF
--- a/challenge/dns01/dns_challenge.go
+++ b/challenge/dns01/dns_challenge.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/go-acme/lego/v4/acme"
@@ -124,7 +125,7 @@ func (c *Challenge) Solve(authz acme.Authorization) error {
 		timeout, interval = DefaultPropagationTimeout, DefaultPollingInterval
 	}
 
-	log.Infof("[%s] acme: Checking DNS record propagation using %+v", domain, recursiveNameservers)
+	log.Infof("[%s] acme: Checking DNS record propagation. [nameservers=%s]", domain, strings.Join(recursiveNameservers, ","))
 
 	time.Sleep(interval)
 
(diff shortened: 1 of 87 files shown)


dmke (approved):
LGTM

PR #2113, opened 2024-02-13
Add flag to only include manual DNS provider

KT0O:
# General idea

When using lego as a library, having support of all the DNS providers is not always useful. 
It adds a huge amount of dependencies,  considerably increasing library size (50MB+ difference).
The idea is to be able to disable these third parties by setting a compile time flag to use when you do not need these implementations.


ldez:
Hello,

we will not do that.

Related to #992

PR #2112, opened 2024-02-13
Add DNS provider for Abion

ldez:
Closes #2111

Ping @Gaardsholt, can you run the command with your domain, credentials, etc. (and put the output of this command to a comment)?

```bash
make build
rm -rf .lego

ABION_API_KEY=yyyy \
./dist/lego  --dns abion -d '*.example.com' -d example.com -s https://acme-staging-v02.api.letsencrypt.org/directory run
```

Note the wildcard domain is important.

<details><summary>How to test this PR?</summary>

1. You need [Go](https://go.dev/doc/install)
2. Checkout the PR:
    ```bash
    git clone https://github.com/ldez/lego.git
    cd lego
    git checkout feat/abion
    ```
3. Compile lego:
    - if you have `make`: `make build`
    - if you don't make `make`: `go build -o dist/lego ./cmd/lego`
4. Run the following command with your own information (email, domain, credentials):
    ```bash
    ABION_API_KEY=yyyy \
    ./dist/lego -m --dns abion -d '*.example.com' -d example.com -s https://acme-staging-v02.api.letsencrypt.org/directory run
    ```
   **The wildcard domain is important**
5. Before each run of the command, you should clean your local environment:
    ```bash
    rm -rf .lego
    ```

</details>

PR #2116, opened 2024-02-17
brandit: provider deprecation

ldez:
Brandit has been acquired by Abion.

Related to #2105

---

Waiting for the merge of #2112

### DIFF
--- a/providers/dns/brandit/brandit.toml
+++ b/providers/dns/brandit/brandit.toml
@@ -1,5 +1,10 @@
-Name = "Brandit"
-Description = ''''''
+Name = "Brandit (deprecated)"
+Description = '''
+Brandit has been acquired by Abion.
+Abion has a different API.
+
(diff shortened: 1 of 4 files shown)


providers/dns/brandit/brandit.toml
```diff
@@ -1,5 +1,10 @@
-Name = "Brandit"
-Description = ''''''
+Name = "Brandit (deprecated)"
+Description = '''
+Brandit has been acquired by Abion.
+Abion has a different API.
```
dmke, line 4: Does this mean the API is already defunct? If so, we could prune the provider similar to what I suggested in #2324 (i.e. immediately return an `errors.New("brandit: provider defunct, migrate to albion")` or similar)

ldez: we don't know, maybe their API still works, so I prefer to not drop it for now.

dmke: I see. In this case, LGTM!

dmke (approved):

Issue #2137, opened 2024-03-10
ERROR acme: error: 500 :: POST renewing with v4.16.0 --- v4.15.0 is OK


stefanomarty:
### Welcome

- [X] Yes, I'm using a binary release within 2 latest releases.
- [X] Yes, I've searched similar issues on GitHub and didn't find any.
- [X] Yes, I've included all information below (version, config, etc).

### What did you expect to see?

Normal renewal of a certificate

### What did you see instead?

This error:
```
2024/03/10 21:36:14 [INFO] [redacted] acme: Trying renewal with 1316 hours remaining
2024/03/10 21:36:14 [INFO] [redacted] acme: Obtaining bundled SAN certificate
2024/03/10 21:36:14 acme: error: 500 :: POST :: https://acme-v02.api.letsencrypt.org/acme/new-order :: urn:ietf:params:acme:error:serverInternal :: While validating order as a replacement and error occurred
```

### How do you use lego?

Binary

### Reproduction steps

Install lego v4.16.0 and try to renew a certificate

### Version of lego

```
lego version 4.16.0 linux/amd64
```

Going back to lego v4.15.0 the certificate gets renewed without errors:

```console
$ wget https://github.com/go-acme/lego/releases/download/v4.15.0/lego_v4.15.0_linux_amd64.tar.gz
tar xf lego_v4.15.0_linux_amd64.tar.gz lego
$ ./lego --version
lego version 4.15.0 linux/amd64

$ ./lego --email="redacted" --http --http.webroot "/var/www/html/"  --domains="redacted" --path=/etc/lego renew --days 60
2024/03/10 21:46:34 [INFO] [redacted] acme: Trying renewal with 1316 hours remaining
2024/03/10 21:46:34 [INFO] [redacted acme: Obtaining bundled SAN certificate
2024/03/10 21:46:35 [INFO] [redacted] AuthURL: https://acme-v02.api.letsencrypt.org/acme/authz-v3/324772220427
2024/03/10 21:46:35 [INFO] [redacted] acme: Could not find solver for: tls-alpn-01
2024/03/10 21:46:35 [INFO] [redacted] acme: use http-01 solver
2024/03/10 21:46:35 [INFO] [redacted] acme: Trying to solve HTTP-01
2024/03/10 21:46:39 [INFO] [redacted] The server validated our request
2024/03/10 21:46:39 [INFO] [redacted] acme: Validations succeeded; requesting certificates
[...]

PR #2138, opened 2024-03-10
fix: don't generate ARI cert ID if ARI is not enable

ldez:
It doesn't fix the problem behind ARI but at least doesn't break the renewal for all users.

Related to #2137


ldez:
I consider this as a critical problem, then I will merge the PR.

dmke (approved):

ldez:
Thank you @dmke :heart: 

I will also create a new release.

PR #2141, opened 2024-03-15
pdns: reconstruct zone URLs to enable non-root folder API endpoints

### LINKED ISSUE #2128: pdns: API endpoint not at URL root resulting in incorrect URL queried and thus failing with error code 404
### Welcome

When requesting a first or renewing an existing certificate via DNS challenge and PowerDNS API with the API endpoint not being located at the URL root (e.g., `https://login.udmedia.de/dns/api` instead of `https://login.udmedia.de/api`), the command should run through without an error.

### What did you see instead?

When requesting a first or renewing an existing certificate via DNS challenge and PDNS API with the API endpoint not being located at the URL root (e.g., `https://login.udmedia.de/dns/api` instead of `https://login.udmedia.de/api`), the command fails with an error code 404, indicating that an unknown URL was queried.

After debugging, this was pinpointed to the URL generated as `endpoint` for updating within [`UpdateRecords` in `providers/dns/pdns/internal/client.go`](https://github.com/go-acme/lego/blob/82e9a5e2a917e72870781d80d2de515c27208304/providers/dns/pdns/internal/client.go#L119-L133) not being the expected `https://login.udmedia.de/dns/api/v1/servers/udmedia/zones/example.com.` but instead `https://login.udmedia.de/dns/api/v1/dns/api/v1/servers/udmedia/zones/example.com.` (i.e., with duplicated `dns/api/v1/`).

The reason for that is that the URL of the host to which the path is appended by `joinPath` is already containing the `/dns` but the URL of the zone that is returned by the provider's PDNS compatible API (determined via `GetHostedZone`) and AFAIU also by PowerDNS itself (tentative as I don't have a PowerDNS installation to try this out with but see, e.g., discussion [here](https://github.com/PowerDNS/pdns/pull/4526)) is an absolute URL path also containing the starting `/dns`. Therefore, `/dns` of the Host URL is extended by a second `/dns` from the Zone URL. Furthermore, in `joinPath` for API version != 0, any path to add that is not starting with `/api` (which is the case here as the path starts with `/dns/api/...`) is also prepended by `/api/v1`. This then results in the superfluous `/api/v1/dns` being added between the host URL and the Zone URL path.

This could be fixed by adding a second join function (e.g., `joinAbsolutePath`) that removes any remaining path from the Host URL before joining and (as this seems unnecessary in that case to me, but please correct me there if this is actually needed for, e.g., earlier version of PowerDNS or the API) not trying to guess whether `/api/v1` needs to be prepended to the URL path. The function could look somewhat like this:

```go
func (c *Client) joinAbsolutePath(elem ...string) *url.URL {
	p := path.Join(elem...)

	rawHost := *c.Host
	rawHost.Path = ""

	return rawHost.JoinPath(p)
}
```

and could replace `joinPath` [here](https://github.com/go-acme/lego/blob/82e9a5e2a917e72870781d80d2de515c27208304/providers/dns/pdns/internal/client.go#L120) and [here](https://github.com/go-acme/lego/blob/82e9a5e2a917e72870781d80d2de515c27208304/providers/dns/pdns/internal/client.go#L140).

I can confirm that this fixes the issue at least for my provider. If helpful, I can provide a corresponding PR also including some tests of the new function. However, the caveat here is that I have limited experience with go and PowerDNS in general and don't want to inadvertently break something for someone else.

### How do you use lego?

Binary

### Reproduction steps

Try requesting a first or renewing an existing certificate via DNS challenge and PDNS API (API version v1) with the API endpoint not being located at the URL root (`PDNS_API_URL` containing a non-empty path, e.g., `https://login.udmedia.de/dns` instead of `https://login.udmedia.de` for an API endpoint at `https://login.udmedia.de/dns/api` instead of `https://login.udmedia.de/api`).

The command used (and environment variables set (except for the API key), see Logs section.

(See also discussion #2122)

For my understanding of the origin of the error after debugging this, see section on "What did you see instead?".

### Version of lego

```console
lego version 4.15.0 linux/386
```

### Logs

<details>

Command executed to generate the log:

```shell
export PDNS_API_URL=https://login.udmedia.de/dns
export PDNS_SERVER_NAME=udmedia
export PDNS_API_KEY=<read_from_file>
export PDNS_API_VERSION=1  # Required as this provider returns not a list but just a single entry when querying the version
./lego --accept-tos --path '.' -d 'some_domain.de' --email 'some_email@example.xom' --key-type 'ec256' --dns 'pdns' --dns.resolvers '5.1.66.255:53' --server 'https://acme-staging-v02.api.letsencrypt.org/directory' run
```

```console
2024/02/28 21:46:31 [INFO] [some_domain.de] acme: Obtaining bundled SAN certificate
2024/02/28 21:46:31 [INFO] [some_domain.de AuthURL: https://acme-staging-v02.api.letsencrypt.org/acme/authz-v3/<snip>
2024/02/28 21:46:31 [INFO] [some_domain.de] acme: Could not find solver for: tls-alpn-01
2024/02/28 21:46:31 [INFO] [some_domain.de] acme: Could not find solver for: http-01
2024/02/28 21:46:31 [INFO] [some_domain.de] acme: use dns-01 solver
2024/02/28 21:46:31 [INFO] [some_domain.de] acme: Preparing to solve DNS-01
2024/02/28 21:46:32 [INFO] [some_domain.de] acme: Cleaning DNS-01 challenge
2024/02/28 21:46:32 [WARN] [some_domain.de] acme: cleaning up failed: pdns: unexpected status code: [status code: 404] body: {
    "error": "Not found"
}
2024/02/28 21:46:32 [INFO] Deactivating auth: https://acme-staging-v02.api.letsencrypt.org/acme/authz-v3/<snip>
2024/02/28 21:46:32 Could not obtain certificates:
[some_domain.de] [some_domain.de] acme: error presenting token: pdns: unexpected status code: [status code: 404] body: {
    "error": "Not found"
}
```

</details>

<details>

```console
$ go version && go env
# paste output here
```

</details>

jotasi:
As I was trying to both better understand the behavior and create a patch for myself that makes lego work for my DNS provider's API, I have implemented a fix for #2128 even as the discussion in the issue has not yet progressed any further. If you require further discussion before this PR should be considered, please let me know.

This corresponds to the first of the two solutions, that I've proposed in #2128. It replaces the usage of the (at least for my provider) absolute URL `zone.URL`, which resulted in incorrect URLs being used when the API endpoint is not located at the server's root, by relative URLs. AFAICS, these should now work in all cases (both if the API endpoint is located at the server's root and if it isn't).

The two added test cases for non-root API endpoints also demonstrate the issue (if you add them without the changes to the code, the tests fail).

The constructed URLs reflect the URLs as they are documented in the PDNS API documentation for [the update for a specific zone](https://doc.powerdns.com/authoritative/http-api/zone.html#put--servers-server_id-zones-zone_id) and for [notifying for a specific zone](https://doc.powerdns.com/authoritative/http-api/zone.html#put--servers-server_id-zones-zone_id-notify). The description in the API is not 100% unambiguous as `{zone_id}` is not explicitly defined as to be the same as the Zone's ID returned by the API for that zone (as `"id"`). But both conceptually (as the zone's ID's main feature is URL-embedability) and from looking at the PDNS code (at least to the level that I could dig through it without spending a lot of time on it), this should be the case. If this is true, I don't see a downside of constructing the URLs in this way for existing use cases and it enables mine.

Sadly, I don't have a PDNS installation to play with and test this further (and also no time to set one up for this). Therefore, I cannot verify myself that it properly works in all cases except for the automated tests and my considerations listed above.

Fixes #2128

### DIFF
--- a/providers/dns/pdns/internal/client.go
+++ b/providers/dns/pdns/internal/client.go
@@ -117,7 +117,7 @@ func (c *Client) GetHostedZone(ctx context.Context, authZone string) (*HostedZon
 }
 
 func (c *Client) UpdateRecords(ctx context.Context, zone *HostedZone, sets RRSets) error {
-	endpoint := c.joinPath("/", zone.URL)
+	endpoint := c.joinPath("/", "servers", c.serverName, "zones", zone.ID)
 
 	req, err := newJSONRequest(ctx, http.MethodPatch, endpoint, sets)
 	if err != nil {
@@ -137,7 +137,7 @@ func (c *Client) Notify(ctx context.Context, zone *HostedZone) error {
 		return nil
 	}
 
-	endpoint := c.joinPath("/", zone.URL, "/notify")
+	endpoint := c.joinPath("/", "servers", c.serverName, "zones", zone.ID, "notify")
 
 	req, err := newJSONRequest(ctx, http.MethodPut, endpoint, nil)
 	if err != nil {
--- a/providers/dns/pdns/internal/client_test.go
+++ b/providers/dns/pdns/internal/client_test.go
@@ -254,13 +254,43 @@ func TestClient_GetHostedZone_v0(t *testing.T) {
 }
 
 func TestClient_UpdateRecords(t *testing.T) {
-	client := setupTest(t, http.MethodPatch, "/api/v1/servers/localhost/zones/example.org.", http.StatusOK, "zone.json")
+	client := setupTest(t, http.MethodPatch, "/api/v1/servers/server/zones/example.org.", http.StatusOK, "zone.json")
 	client.apiVersion = 1
 
 	zone := &HostedZone{
 		ID:   "example.org.",
 		Name: "example.org.",
-		URL:  "api/v1/servers/localhost/zones/example.org.",
+		URL:  "api/v1/servers/server/zones/example.org.",
+		Kind: "Master",
+	}
+
+	rrSets := RRSets{
+		RRSets: []RRSet{{
+			Name:       "example.org.",
+			Type:       "NS",
+			ChangeType: "REPLACE",
+			Records: []Record{{
+				Content: "192.0.2.5",
+				Name:    "ns1.example.org.",
+				TTL:     86400,
+				Type:    "A",
+			}},
+		}},
+	}
+
+	err := client.UpdateRecords(context.Background(), zone, rrSets)
+	require.NoError(t, err)
+}
+
+func TestClient_UpdateRecords_NonRootApi(t *testing.T) {
+	client := setupTest(t, http.MethodPatch, "/some/path/api/v1/servers/server/zones/example.org.", http.StatusOK, "zone.json")
+	client.Host = client.Host.JoinPath("some", "path")
+	client.apiVersion = 1
+
+	zone := &HostedZone{
+		ID:   "example.org.",
+		Name: "example.org.",
+		URL:  "some/path/api/v1/servers/server/zones/example.org.",
 		Kind: "Master",
 	}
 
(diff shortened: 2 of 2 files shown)


jotasi:
I rebased onto the current master to keep the branch up to date. I'll do so every once in a while. If this is not helpful for your review, please let me know.

ldez:
You don't need to rebase unless there are conflicts.

ldez (approved):
LGTM

PR #2144, opened 2024-03-20
chore: update to go1.22

ldez:
- update linter to v1.57.0

https://tip.golang.org/doc/go1.22

### DIFF
--- a/.github/workflows/go-cross.yml
+++ b/.github/workflows/go-cross.yml
@@ -16,7 +16,7 @@ jobs:
 
     strategy:
       matrix:
-        go-version: [ stable, oldstable ]
+        go-version: [ stable ]
         os: [ubuntu-latest, macos-latest, windows-latest]
 
     steps:
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       GO_VERSION: stable
-      GOLANGCI_LINT_VERSION: v1.57.0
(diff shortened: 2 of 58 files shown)


dmke (approved):
LGTM

We should also update `docs/content/installation/_index.md` (which still mentions Go 1.17), as the new `range` semantics require 1.22+.

dmke (approved):

iwind:
Is it realy worth to do this? it's a expensive decision, and gain less benifits.

dmke:
Expensive in terms of... what exactly? I find the tests *much* easier to read and maintain without the line noise in `for`-loops.

Sure, we loose support for Go 1.21. But since Lego follows the Go support cycle (i.e. the only the current and previous Go releases are supported), and Go 1.23 is slated for release in August, we *could* have waited until then. Using 1.22 *now* however is not a huge ask.

iwind:
To improve some (maybe 4-5) for loops, you forced thousands projects to upgrade their Go version to 1.22. Okay, maybe you are right, and everyone should and MUST fork the repository and change the go version in their go.mod file. It seems not a big problem, but it's still EXPENSIVE in my own opinion. 

And, at last, i sincerely  thank you for your work.

dmke:
Again, not merging this PR now would have delayed the 1.22 upgrade only by a few months.

Also note that this commit is not yet part of any release. If you want to use the development version, you need to keep up with the project dependencies :)

ldez:
> you forced thousands projects to upgrade their Go version to 1.22

Can you provide a list of those thousands projects?

Those thousands projects that use unreleased version of lego should think to sponsor lego if it's critical for them.

PR #2166, opened 2024-04-20
exec: stream command output

### LINKED ISSUE #2165: '--dns exec' show output in real time?
### Welcome

### How do you use lego?

Binary

### Detailed Description

I'm working on a shell script to use with `--dns exec` (to manually add/remove DNS records), and running into issues that take significantly longer than they should to debug, because the part of `lego` which runs my script appears to be gathering the script's output in memory and then dumping it to the output stream all at once. The process of adding/removing DNS records can take up to two minutes per name, so if I'm testing with a certificate having multiple hostnames on it (i.e. `name1.domain.xyz`, `'name2.domain.xyz`, and `name3.domain.xyz`), I see no output at all for several minutes and then get a dump all at once at the end.

I'm looking at `providers/dns/exec/exec.go`, but I don't know enough about the exec module to know if this is even possible, so ... is there a way to make the code print each line of input, as soon as the external script prints it?

ldez:
The command outputs will be displayed in realtime.

Fixes #2165

### DIFF
--- a/providers/dns/exec/exec.go
+++ b/providers/dns/exec/exec.go
@@ -2,6 +2,7 @@
 package exec
 
 import (
+	"bufio"
 	"context"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


dmke (approved):

providers/dns/exec/exec.go
```diff
@@ -117,10 +118,27 @@ func (d *DNSProvider) run(ctx context.Context, command, domain, token, keyAuth s
...
 
-	output, err := cmd.CombinedOutput()
-	if len(output) > 0 {
-		log.Println(string(output))
+	stdout, err := cmd.StdoutPipe()
+	if err != nil {
+		return fmt.Errorf("create pipe: %w", err)
+	}
+
+	cmd.Stderr = cmd.Stdout
```
kg4zow, line 126: Oooh, that actually answers a question that I had in the back of my mind, namely how to "capture" both STDOUT and STDERR and "combine" them into a single stream. I'm glad I saw this.

Also, I didn't realize you were a maintainer. Thanks for merging this so quickly.

PR #2173, opened 2024-05-05
ovh: add OAuth2 authentication

Idix:
## Description

The latest go-ovh version allows authentication using OAuth2 client_id and client_secret. Which in turn can be provisioned using the OVH terraform provider thus enabling everything to be configured as code.

## Note

I'm not usually writing code in Go so I'm not sure if everything is idiomatic. Please advice if something needs to be changed to be more in line with the rest of the code.


ldez (approved):
LGTM

Idix:
I see you did all the hard work, thanks!

PR #2179, opened 2024-05-06
exoscale: simplify record creation

### LINKED ISSUE #2174: exoscale: fails when issuing certificate for "*.domain.example" and "domain.example"
### Welcome

ACME DNS-01 challenge succeeds for both `*.domain.example` and `domain.example` and the certificate is issued.

### What did you see instead?

ACME DNS-01 challenge for one of the domains fails because the Exoscale DNS provider does not create two TXT records (with the same name). Instead, the record for one of the domains is overwritten with the value of the other domain before validation starts. ~This is a limitation of the Exoscale DNS API: it does not allow creation of TXT records with the same name. The domains must be validated sequentially just like the `exec` provider does it.~

### How do you use lego?

Binary

### Reproduction steps

`EXOSCALE_API_KEY=<key> EXOSCALE_API_SECRET=<secret> lego --email <email> --dns exoscale --domains <domain> --domains *.<domain> run`

### Version of lego

```console
lego version 4.16.1 linux/amd64
```

### Logs

<details>

```console
# paste output here
```

</details>

<details>

```console
$ go version && go env
# paste output here
```

</details>

ldez:
Instead of updating a record, it always creates a new record.

To find the record to delete, the record value is used.

Fixes #2174

### DIFF
--- a/providers/dns/exoscale/exoscale.go
+++ b/providers/dns/exoscale/exoscale.go
@@ -127,38 +127,16 @@ func (d *DNSProvider) Present(domain, token, keyAuth string) error {
 		return fmt.Errorf("exoscale: zone %q not found", zoneName)
 	}
 
-	recordID, err := d.findExistingRecordID(deref(zone.ID), recordName)
-	if err != nil {
-		return fmt.Errorf("exoscale: %w", err)
-	}
-
-	if recordID == "" {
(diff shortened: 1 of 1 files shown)


dmke (approved):

pierre-emmanuelJ (commented):
LGTM

PR #2189, opened 2024-05-21
chore: remove remaining deprecated ARI call

### LINKED ISSUE #2187: Leftover deprecated update renewal info call
### Welcome

### How do you use lego?

Library

### Detailed Description

I see there’s still a call in core for update renewal info API (`acme/api/renewal.go`) but that was made obsolete in the latest ARI specification. Is there a reason for leaving it there or was it just an oversight when removing other pre v3 ARI deprecated calls? Should this one also be removed so we can avoid any further confusion?

mocheryl:
Closes #2187

### DIFF
--- a/acme/api/renewal.go
+++ b/acme/api/renewal.go
@@ -3,8 +3,6 @@ package api
 import (
 	"errors"
 	"net/http"
-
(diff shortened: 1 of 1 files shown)


ldez (approved):
LGTM

PR #2191, opened 2024-05-27
dode: update API URL

henrikalves:
Hi,
we are moving our API to a different subdomain (the old one will continue to work for now),
I also have updated the links to the documentation


ldez:
It's sad that the documentation is not public, it can be a maintenance problem.

ldez (approved):
LGTM

PR #2204, opened 2024-06-11
designate: allow manually overwriting DNS zone

jan-di:
This is basically the same feature as in #1433 for the OpenStack Designate provider.

This solves problems that occur in situations like SplitDNS/Private DNS Zones where the automatic SOA detection might not work to the full extend. When setting the Zone Name via environment, it specifically uses this Zone Name when communicating with the Designate API. Otherwise, nothing changes.


jan-di:
@ldez Thank you for refactoring! I'm not that fluent in golang :-)

Is there anything else I should do or what is needed from me?

PR #2216, opened 2024-06-21
ovh: allow to use ovh.conf file

### LINKED ISSUE #2215: ovh: allow to use ovh.conf
### Welcome

### How do you use lego?

Through Terraform ACME provider

### Detailed Description

Hello there !

We have some OVHcloud customers that are using the ability to provide the API credentials through ovh.conf file (see [README](https://github.com/ovh/go-ovh/blob/6817886d12a8c5650794b28da635af9fcdfd1162/README.md?plain=1#L152)) notably in OVHcloud terraform provider.

When they are also using acme provider in a plan, they need to change the way they provide the credentials since in acme, the only supported way seems to be through [environment variables](https://github.com/go-acme/lego/blob/master/providers/dns/ovh/ovh.go#L102)

This feature request would be to enable the usage of `ovh.conf` file in acme for the OVHcloud provider.

ldez:
Fixes #2215

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -187,3 +187,5 @@ require (
 	gopkg.in/ini.v1 v1.67.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+replace github.com/ovh/go-ovh => github.com/ldez/go-ovh v1.4.2-0.20240621144634-e57d43b43c1c
--- a/go.mod
+++ b/go.mod
@@ -187,5 +187,3 @@ require (
 	gopkg.in/ini.v1 v1.67.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
-
-replace github.com/ovh/go-ovh => github.com/ldez/go-ovh v1.4.2-0.20240621144634-e57d43b43c1c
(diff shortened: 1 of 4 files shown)


rbeuque74:
Hello,

Thanks for the patch. I suggested another way around here: https://github.com/ldez/lego/pull/2

What do you think ?

Romain

dmke:
I'm a little confused: Where does it read the `ovh.conf`, and why was ovh/go-ovh#84 closed without merging?

ldez:
The read of the file is inside the constructors of the client.

https://github.com/ovh/go-ovh/blob/9c5297024e0b551f222c66504a3ffdf7b0cd479b/ovh/configuration.go#L77-L91

https://github.com/ovh/go-ovh/blob/9c5297024e0b551f222c66504a3ffdf7b0cd479b/ovh/ovh.go#L103-L159

My PR was to be able to fill all the fields and then call `loadConfig` but by playing with constructors we can do the same.

It's not my preferred solution but it works as expected.

dmke (approved):
> It's not my preferred solution but it works as expected.

I concur. It works *now* as expected, but as soon as OVH changes their client implementation (as likely or unlikely this may be) and the behaviour diverges, we have a problem.

I **much** prefer ovh/go-ovh#84 (or something similar) to become available, but alas, this seems to be the next best thing...

PR #2217, opened 2024-06-29
fix: restore Solaris

ldez:
I removed Solaris during [v4.17.2](https://github.com/go-acme/lego/blob/master/CHANGELOG.md#v4172---2024-05-28) because of a problem with [flock](https://github.com/gofrs/flock/issues/60).

The problem with flock is now fixed https://github.com/gofrs/flock/releases/tag/v0.10.0

```bash
GOOS=solaris GOARCH=amd64 go build -o templego ./cmd/lego
```

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -17,7 +17,7 @@ builds:
       - linux
       - freebsd
       - openbsd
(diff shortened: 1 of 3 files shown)


dmke (approved):
Solaris... that's a name I haven't heard in a looong time :)

PR #2220, opened 2024-07-07
feat: add --quiet option

### LINKED ISSUE #866: Option to silence non-error log messages on successful CLI run
It'd be great if there were a `--quiet` or similar flag to silence non-error log messages when `lego` successfully issues or renews a certificate. I've implemented this in a wrapper script for running Lego via FreeBSD `periodic`, but it's a bit [clunky](https://gist.github.com/mhoran/1d8cbe8ccb61bef7a9fb2d93baf2fb85).

The use case is this: `peroidic` runs via `cron`, and can be quite noisy. Silencing output of successful jobs is generally desirable. There's even an option to do this by default. However, to provide a uniform experience for all users, I've added the above hack to the script.

I'd be happy to take a pass at this if it's agreed that this is a valuable feature. I also realize that the CLI is undergoing a refactor, so perhaps this work should wait until that's complete.

raylu:
Closes #866

### DIFF
--- a/cmd/cmd_before.go
+++ b/cmd/cmd_before.go
@@ -6,6 +6,8 @@ import (
 )
 
 func Before(ctx *cli.Context) error {
+	log.Quiet = ctx.Bool("quiet")
+
 	if ctx.String("path") == "" {
 		log.Fatal("Could not determine current working directory. Please pass --path.")
 	}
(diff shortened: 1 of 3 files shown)


ldez:
Hello,

I will decline the PR because the proposed option is not something that fits with standard log management.

lingfish:
> something that fits with standard log management

My 2 cents is that I don't agree.  It's quite common practice to have an argument like this, and technically, the output is console output, not log output.

ldez:
The issue is still open, I just don't agree with the implementation.

We have several issues and PRs about the topic, I want to handle all of them by a full rewrite of the log management.

Issue #970, opened 2019-09-17
Use context.Context to enforce cancellation


mholt:
It would be useful, when shutting down a server or even the lego command, that Ctrl+C can cancel any active & pending challenges in a way that their resources get cleaned up when shutting down.

In other words, an API that exposes [`context.Context`](https://golang.org/pkg/context) which can be used for cancellation would be helpful. What do you think?

mholt:
Implementing this in https://github.com/mholt/acme

systemcrash:
Just ran into this. Doing a certain amount of testing means if I CTRL+C while lego is waiting for record propagation, it does not clean up DNS records.

outofforest:
And when doing DNS-01 it's not possible to gracefully stop application until timeout happens after 1 minute.

PR #2221, opened 2024-07-10
support context.Context when use as library

hirosumee:
When used lego as a library we need a way to pass context then controlling  the timeout. 
So i make this PR

### DIFF
--- a/acme/api/account.go
+++ b/acme/api/account.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"context"
 	"encoding/base64"
 	"errors"
 	"fmt"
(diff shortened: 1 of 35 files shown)


ldez:
Hello,

First, I recommend looking for existing issues before opening a PR.
If there is no issue, you can create one and wait for feedback before opening a PR.
If there is an issue, you should add a comment on the issue and wait for feedback before opening a PR.

Currently, an issue already exists: #970

The major problem with this change is that it's breaking.
And, for now, I don't want to be breaking on that.

Also, you don't need context to handle the timeout:
- each provider has a specific option for that
- the DNS client has also an option for that
- etc.

So, for now, I will decline this PR.

PR #2235, opened 2024-07-25
chore: bump to github.com/vultr/govultr/v3

mikelolasagasti:
Version v3 of the API ~~adds http.response to calls, add a function to report response code.~~

https://github.com/vultr/govultr/compare/v2.17.2...v3.9.0


ldez:
The management of the response and the status code is already handled by the API client.

https://github.com/vultr/govultr/blob/d2142cf9d4b52038721b4fb452441acf28f49d11/govultr.go#L214-L223

ldez (approved):
LGTM

PR #2256, opened 2024-08-31
exoscale: update Egoscale to v3.1.1

mlec1:
Update Egoscale library for the Exoscale DNS provider to v3.

v1 and v2 are deprecated: https://github.com/exoscale/egoscale?tab=readme-ov-file#deprecated-use-v3

https://github.com/exoscale/egoscale/compare/v0.102.3...v3.1.1


pierre-emmanuelJ (commented):
👍 LGTM

Thanks, @mlec1, for the contribution! And thanks @ldez for the review :)

providers/dns/exoscale/exoscale.go
```diff
@@ -84,12 +81,8 @@ func NewDNSProviderConfig(config *Config) (*DNSProvider, error) {
 		return nil, errors.New("exoscale: credentials missing")
 	}
 
-	egoscale.UserAgent = "go-acme/lego " + egoscale.UserAgent
```
pierre-emmanuelJ: Any reason for dropping it?

ldez: Because it's a global variable, I don't want to produce side effects: if someone uses lego and also exocale as a lib, using the global variable will be a problem.

Generally speaking, the use of global variables should be avoided.

ldez: IMHO, the `User-Agent` should be an option of the `Client` like the other `ClientOpt` instead of a global variable.

ldez: I was hoping something like that: https://github.com/exoscale/egoscale/compare/master...ldez:egoscale:feat/user-agent

I will not open a PR because it's a breaking change, except if you want it.

ldez (approved):
LGTM

PR #2260, opened 2024-09-04
docs: update installation for FreeBSD

lcheylus:
- lego is in official FreeBSD ports tree: https://www.freshports.org/security/lego
- use `pkg install` to install package


ldez (approved):
LGTM

PR #2261, opened 2024-09-04
docs: add installation for OpenBSD

lcheylus:
- lego is in official OpenBSD ports tree https://openports.pl/path/security/lego
- install it with `pkg_add lego`


ldez (approved):
LGTM

PR #2284, opened 2024-09-21
feat: add dns.propagation-rns option

### LINKED ISSUE #2276: Provide an option to verify propagation against a custom list of DNS-servers
### Welcome

### How do you use lego?

Binary and Traefik.

### Detailed Description

Hi, we've hit an issue using DNS-01 challenge with EAB and CNAME delegation with [EJBCA](https://www.ejbca.org/) as the CA, but this deployment scenario is probably going to be more common as ACME has started to be adopted in enterprise.

I have included a diagram to try and explain our interpretation of the current flow, and where we see an issue.

lego

1. We issue a renewal request, all good so far
2. We create the TXT record with the challenge
3. LEGO uses our `--dns.resolvers` that points to the same DNS-server as the PKI will query for the challenge. These servers are used by lego to find the authoritative servers only as specified [here](https://github.com/go-acme/lego/blob/0da0942081a1666a73b4f2f3f35328e9ae28af0f/cmd/flags.go#L121-L123).
  It finds the authoritative servers and can successfully query to verify they serve the TXT record with the correct value.
4. LEGO gives the all clear to the PKI
5. PKI queries DNS against the company wide DNS-service. This company DNS-service has network communication to the authoritative DNS-servers over tcp/udp 53. Queries are cached and if forwarded.

So now to the problem:
* Step 5 we can see the PKI is NOT querying the authoritative DNS-server. This is by network security design.
* In Step 3 we would like to validate the same record as we added to the authoritative DNS-server has propagated to the DNS-servers used by the PKI before giving the "All good" message to the PKI.

Essentially we'd want something like this (behind a flag such as `--dns.propagation-servers` of `[]string` type) to specify what additional servers we'd want to append to the `checkAuthoritativeNss` function.

https://github.com/Jonher937/lego/blob/a152249a1a02146604936099d3bf1a9d13999280/challenge/dns01/precheck.go#L76-L88

Full commit can be found [here](https://github.com/go-acme/lego/compare/master...Jonher937:lego:verify-propagation-at-recursor) and acts as an example. I have not found a good way to pass down a flag this deep into the process, but this issue is created as a topic for discussing how and if this could be implemented.

In the end our issues comes down to the DNS topology/implementation and slow (60+ seconds) propagations to the company DNS-service which the PKI verifies against.

We have also tries with the newly added `--dns.propagation-wait` option and have successfully managed to obtain certificates if we tweak it high enough for the propagation to happen, but this might randomly fail if the zone update has not yet made it's way to the company DNS-service.

ldez:
- deprecation of `dns.disable-cp`, replaced by `dns.propagation-disable-ans`.
- add `dns.propagation-rns` option to check all the authoritative nameservers.

Fixes #2276

### DIFF
--- a/challenge/dns01/precheck.go
+++ b/challenge/dns01/precheck.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net"
 	"strings"
+	"time"
 
 	"github.com/miekg/dns"
 )
@@ -28,6 +29,13 @@ func DisableCompletePropagationRequirement() ChallengeOption {
 		chlg.preCheck.requireCompletePropagation = false
 		return nil
 	}
+}
+
+func PropagationWaitOnly(wait time.Duration) ChallengeOption {
+	return WrapPreCheck(func(domain, fqdn, value string, check PreCheckFunc) (bool, error) {
+		time.Sleep(wait)
+		return true, nil
+	})
 }
 
 type preCheck struct {
--- a/challenge/dns01/precheck.go
+++ b/challenge/dns01/precheck.go
@@ -24,9 +24,15 @@ func WrapPreCheck(wrap WrapPreCheckFunc) ChallengeOption {
 	}
 }
 
+// DisableCompletePropagationRequirement obsolete.
+// Deprecated: use DisableAuthoritativeNssPropagationRequirement instead.
 func DisableCompletePropagationRequirement() ChallengeOption {
+	return DisableAuthoritativeNssPropagationRequirement()
+}
+
+func DisableAuthoritativeNssPropagationRequirement() ChallengeOption {
 	return func(chlg *Challenge) error {
-		chlg.preCheck.requireCompletePropagation = false
+		chlg.preCheck.requireAuthoritativeNssPropagation = false
 		return nil
 	}
 }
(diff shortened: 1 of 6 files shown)


challenge/dns01/precheck.go
```diff
@@ -23,23 +24,47 @@ func WrapPreCheck(wrap WrapPreCheckFunc) ChallengeOption {
...
+	return func(chlg *Challenge) error {
+		chlg.preCheck.requireAuthoritativeNssPropagation = false
+		return nil
+	}
+}
+
+func RecursiveNSsPropagationRequirement() ChallengeOption {
 	return func(chlg *Challenge) error {
-		chlg.preCheck.requireCompletePropagation = false
+		chlg.preCheck.requireAuthoritativeNssPropagation = true
```
Jonher937, line 42:
```suggestion
		chlg.preCheck.requireRecursiveNssPropagation = true
```
Calling `RecursiveNSsPropagationRequirement()`should set `chlg.preCheck.requireRecursiveNssPropagation`. Not authoritative.

ldez: good catch

cmd/flags.go
```diff
@@ -147,11 +150,19 @@ func CreateFlags(defaultPath string) []cli.Flag {
 		},
 		&cli.BoolFlag{
 			Name:  flgDNSDisableCP,
+			Usage: fmt.Sprintf("(deprecated) use %s instead.", flgDNSPropagationDisableANS),
+		},
+		&cli.BoolFlag{
+			Name:  flgDNSPropagationDisableANS,
 			Usage: "By setting this flag to true, disables the need to await propagation of the TXT record to all authoritative name servers.",
```
Jonher937, line 157:
```suggestion
			Usage: "By setting this flag, disables the need to await propagation of the TXT record to all authoritative name servers.",
```
Sorry if this is a nitpick, but that matches the style of the other descriptions better.

ldez: I will do the opposite, just because I want to be really explicit on those flags, I'm afraid there is ambiguity.

dmke (approved):
LGTM

PR #2285, opened 2024-09-29
perf: reducing the lock strength of the soa cache entry

djx30103:
When a large number of applications are made, there will be a large number of lock seizures and lower performance

### DIFF
--- a/challenge/dns01/nameserver.go
+++ b/challenge/dns01/nameserver.go
@@ -12,16 +12,12 @@ import (
 	"time"
 
 	"github.com/miekg/dns"
-	"golang.org/x/sync/singleflight"
 )
 
 const defaultResolvConf = "/etc/resolv.conf"
 
 var (
-	fqdnSoaCache sync.Map
-
-	// fqdnSoaGroup is a singleflight group to ensure that concurrent DNS lookups for SOA records are only executed once.
-	fqdnSoaGroup singleflight.Group
+	fqdnSoaCache = &sync.Map{}
 )
 
 var defaultNameservers = []string{
@@ -54,7 +50,7 @@ func (cache *soaCacheEntry) isExpired() bool {
 
 // ClearFqdnCache clears the cache of fqdn to zone mappings. Primarily used in testing.
 func ClearFqdnCache() {
-	fqdnSoaCache = sync.Map{}
+	fqdnSoaCache = &sync.Map{}
 }
 
 func AddDNSTimeout(timeout time.Duration) ChallengeOption {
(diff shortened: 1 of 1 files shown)


ldez:
Hello,

`singleflight.Group` is a `map` with a lock.
So you replaced a `map` with a lock with a `map` with a lock.
I think there are no advantages.

Additionally, with your changes, the `ClearFqdnCache` function is no longer concurrently safe.

ldez:
Your commit is not the expected answer, I'm waiting for a comment (in English).

Otherwise, are you an employee of @fit2cloud?

icpd:
When and initiating a certificate request request and `fetchSoaByFqdn` causes a `goroutine` to occupy a lock for a long time due to a `nameserver` failure, `sync.Map` solves this problem.

If the same `fqdn` does not have a concurrent request problem, then `singleflight` is no longer required to.

ldez:
Is it possible to have an answer of my questions?

I able to read the code, I just need answers to my previous questions.

djx30103:
Hello,
  I am not a fit2cloud employee.

The lines of code may not match, I have modified some of the handling of `defaultNameservers`, but the logic of `lookupNameservers` I haven't changed.
image
image

This is before modification, 200 concurrent.

image

This is before the modification, 200 concurrent.

image


Blocking changed from lock to udp, and each blocks its own udp request, so the result will be returned quickly, reducing the number of coroutines

ldez (approved):
LGTM

Based on the information from this PR:
- 2 different users on the same PR
- 200 concurrent requests per second

I can be wrong, but I think there is a company behind this, and it's not a problem.

But, please consider asking your company to donate.
If there is no company behind this, please consider donating anyway.
This will be appreciated, thank you :heart:

challenge/dns01/nameserver.go
```diff
@@ -51,9 +48,7 @@ func (cache *soaCacheEntry) isExpired() bool {
 
 // ClearFqdnCache clears the cache of fqdn to zone mappings. Primarily used in testing.
 func ClearFqdnCache() {
-	muFqdnSoaCache.Lock()
-	fqdnSoaCache = map[string]*soaCacheEntry{}
-	muFqdnSoaCache.Unlock()
+	fqdnSoaCache.Clear()
```
drakkan, line 51: `Clear()` was added in Go 1.23, please update go.mod or use API compatible with Go 1.22. Thank you

ldez: We will not update to go1.23, but I will change this.

PR #2288, opened 2024-10-02
feat: add PropagationWait function

ldez:
Replace the function `PropagationWaitOnly` with `PropagationWait`.

It's not a breaking change because #2266 is not a part of a release.

### DIFF
--- a/challenge/dns01/precheck.go
+++ b/challenge/dns01/precheck.go
@@ -44,10 +44,15 @@ func RecursiveNSsPropagationRequirement() ChallengeOption {
 	}
 }
 
-func PropagationWaitOnly(wait time.Duration) ChallengeOption {
+func PropagationWait(wait time.Duration, only bool) ChallengeOption {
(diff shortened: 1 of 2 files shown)


challenge/dns01/precheck.go
```diff
@@ -44,10 +44,15 @@ func RecursiveNSsPropagationRequirement() ChallengeOption {
 	}
 }
 
-func PropagationWaitOnly(wait time.Duration) ChallengeOption {
+func PropagationWait(wait time.Duration, only bool) ChallengeOption {
```
dmke, line 47: The name `only` does not really convey a meaning (as did the previous function name, but I didn't catch that in #2266).

Something like `onlySleep` or `noCheck` would be more descriptive, but is maybe too close to the implementation? Maybe this function just needs a bit of documentation.

dmke (changes requested):

challenge/dns01/precheck.go
dmke: Sounds good :+1:

dmke (approved):

PR #2292, opened 2024-10-04
fix: follow Go convention about generated comment

ldez:
https://pkg.go.dev/cmd/go#hdr-Generate_Go_files_by_processing_source


dmke (approved):

dmke (changes requested):
Hang on,

> This line must appear before the first non-comment, non-blank text in the file.

i.e. before the `package` statement.

dmke (approved):

PR #2293, opened 2024-10-04
feat: generate User-Agent for DNS API clients

ldez:
Updates the user agent generator to create a shared user agent between DNS clients.

The header follows HTTP conventions: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent#syntax


acme/api/internal/sender/useragent.go
```diff
@@ -1,7 +1,6 @@
 package sender
 
-// CODE GENERATED AUTOMATICALLY
-// THIS FILE MUST NOT BE EDITED BY HAND
+// Code generated by 'internal/useragent'; DO NOT EDIT.
```
dmke, line 3: > This line must appear before the first non-comment, non-blank text in the file.

internal/useragent/data_dns.go
```diff
@@ -0,0 +1,36 @@
...
+
+const dnsTemplate = `package useragent
+
+import (
+	"fmt"
+	"net/http"
+	"runtime"
+)
+
+// Code generated by 'internal/useragent'; DO NOT EDIT.
```
dmke, line 15:
```suggestion
const dnsTemplate = `// Code generated by 'internal/useragent'; DO NOT EDIT.

package useragent

import (
	"fmt"
	"net/http"
	"runtime"
)
```

internal/useragent/data_sender.go
```diff
@@ -0,0 +1,21 @@
+package main
+
+const senderBaseUserAgent = "xenolf-acme/"
+
+const senderSourceFile = "./acme/api/internal/sender/useragent.go"
+
+const senderTemplate = `package sender
+
+// Code generated by 'internal/useragent'; DO NOT EDIT.
```
dmke, line 9:
```suggestion
const senderTemplate = `// Code generated by 'internal/useragent'; DO NOT EDIT.

package sender
```

dmke (changes requested):

acme/api/internal/sender/useragent.go
dmke: (The remark above ought to be part of the review. It becomes obsolete after changing the templates and regenerating the files.)

dmke (approved):

dmke (approved):

PR #2298, opened 2024-10-10
feat: attempt to check ARI unless explicitly disabled

beautifulentropy:
With [draft-ietf-acme-ari](https://datatracker.ietf.org/doc/draft-ietf-acme-ari/) now in IETF Working Group Last Call—the final step before becoming an RFC—it is reasonable to enable ARI checks by default.

### DIFF
--- a/cmd/cmd_renew.go
+++ b/cmd/cmd_renew.go
@@ -20,7 +20,7 @@ import (
 // Flag names.
 const (
 	flgDays                   = "days"
-	flgARIEnable              = "ari-enable"
+	flgARIDisable             = "ari-disable"
 	flgARIWaitToRenewDuration = "ari-wait-to-renew-duration"
 	flgReuseKey               = "reuse-key"
 	flgRenewHook              = "renew-hook"
@@ -61,8 +61,8 @@ func createRenew() *cli.Command {
 				Usage: "The number of days left on a certificate to renew it.",
 			},
 			&cli.BoolFlag{
-				Name:  flgARIEnable,
-				Usage: "Use the renewalInfo endpoint (draft-ietf-acme-ari) to check if a certificate should be renewed.",
+				Name:  flgARIDisable,
+				Usage: "Do not use the renewalInfo endpoint (draft-ietf-acme-ari) to check if a certificate should be renewed.",
 			},
 			&cli.DurationFlag{
 				Name:  flgARIWaitToRenewDuration,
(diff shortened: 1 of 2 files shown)


ldez:
I don't merge the PR because I have doubts about the impact of this PR: I don't see any problem with doing this with LE, but I wonder about the ACME servers. :thinking: 

I'm torn between having a better default value, but that can introduce a change that may upset a large number of users, or just keeping the current option.

beautifulentropy:
> I don't merge the PR because I have doubts about the impact of this PR: I don't see any problem with doing this with LE, but I wonder about the ACME servers. 🤔
> 
> I'm torn between having a better default value, but that can introduce a change that may upset a large number of users, or just keeping the current option.

Oh, interesting! I figured that this was reasonably safe given that use of ARI is actually gated on whether or not the CA includes a `renewalInfo` field in its `directory` object [draft-ietf-acme-ari-06#section-3](https://datatracker.ietf.org/doc/html/draft-ietf-acme-ari-06#section-3). I'd be curious to hear your concerns @ldez.

ldez:
yes `api.ErrNoARI` is managed but there are several elements related to the flag:

https://github.com/go-acme/lego/blob/19b535caac614ece950449b5d8fd3c4f93e457f5/cmd/cmd_renew.go#L212-L217

https://github.com/go-acme/lego/blob/19b535caac614ece950449b5d8fd3c4f93e457f5/cmd/cmd_renew.go#L282-L287

I don't want to have the same problem as before: https://github.com/go-acme/lego/pull/2138

beautifulentropy:
If we change these to log.Warnf and continue using the default renewal logic would that be sufficient?

ldez:
:thinking: maybe we can move the call to `MakeARICertID` inside the first `if`

ldez:
something like that:
```go
	var replacesCertID string
	var ariRenewalTime *time.Time
	if ctx.Bool(flgARIEnable) {
		ariRenewalTime = getARIRenewalTime(ctx, cert, domain, client)
		if ariRenewalTime != nil {
			now := time.Now().UTC()
			// Figure out if we need to sleep before renewing.
			if ariRenewalTime.After(now) {
				log.Infof("[%s] Sleeping %s until renewal time %s", domain, ariRenewalTime.Sub(now), ariRenewalTime)
				time.Sleep(ariRenewalTime.Sub(now))
			}
		}

		replacesCertID, err = certificate.MakeARICertID(cert)
		if err != nil {
			log.Fatalf("Error while construction the ARI CertID for domain %s\n\t%v", domain, err)
		}
	}

	if ariRenewalTime == nil && !needRenewal(cert, domain, ctx.Int(flgDays)) {
		return nil
	}
...

	if replacesCertID != "" {
		request.ReplacesCertID = replacesCertID
	}
```

ldez (approved):
LGTM

PR #2311, opened 2024-10-25
chore: update readme generator

dmke:
This changes the provider table generator to output an HTML table directly, instead of writing Markdown markup.

This is mostly an attempt of keeping future PRs introducing new providers a little less noisy. Since the provider entries are now strictly stacked vertically in the markup (i.e. at most one provider per line), adding another one in the middle will at worst shuffle the `</tr><td>` tags around.

As a nifty side-effect, this also eliminates the empty header row:

<details><summary>before</summary>

image

</details>
<details open><summary>after</summary>

image

</details>

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,218 +51,182 @@ Detailed documentation is available [here](https://go-acme.github.io/lego/dns).
 
 <!-- START DNS PROVIDERS LIST -->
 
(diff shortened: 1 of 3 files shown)


ldez (approved):
LGTM

PR #1967, opened 2023-07-23
Add DNS provider for hosting.nl

ldez:
- [x] add a description to your PR
- [x] have a homogeneous design with the other providers
- [x] add tests (units)
- [x] add tests ("live")
- [x] add a [provider descriptor](https://github.com/go-acme/lego/blob/master/providers/dns/cloudflare/cloudflare.toml)
- [x] generate CLI help, documentation, and readme.
- [ ] be able to do: _(and put the output of this command to a comment in your PR)_
  ```bash
  make build
  rm -rf .lego
  
  ./dist/lego -m your@email.com --dns hostingnl -d *.example.com -d example.com -s https://acme-staging-v02.api.letsencrypt.org/directory run
  ```
  Note the wildcard domain is important.
- [x] pass the linter
- [x] do `go mod tidy`

Ping @arnovdk, can you run the command (with your domain, email and credentials)?

Closes #1598

PR #2312, opened 2024-10-26
chore: fix readme generator

ldez:
If the number of providers is odd then it generates non-empty cells:
```
<td><a href="https://go-acme.github.io/lego/dns//"></a></td>
```

https://github.com/go-acme/lego/pull/1967/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5


dmke (approved):
Good catch!

PR #2317, opened 2024-11-01
fix: Parse printf verbs in log line output

fhemberger:
Current log output in `setup_challenges.go` looks like this:

```
Nov 01 01:05:07 xxx lego[1548788]: 2024/11/01 01:05:07 The flag '%s' is deprecated use '%s' instead. dns.disable-cp dns.propagation-disable-ans
```

This MR fixes the output.


ldez (approved):
LGTM

PR #2324, opened 2024-11-06
cloudxns: provider deprecation

### LINKED ISSUE #2323: Remove CloudXNS DNS provider
### Welcome

The CloudXNS DNS provider has shut down. They [ended all free service in 2019](https://github.com/certbot/certbot/issues/7276) and then shut down. Their website is now a domain name seller parking page.

lego should remove the DNS provider from the supported providers and the code that interacts with their API

### What did you see instead?

<details>

image

</details>

### How do you use lego?

Binary

### Reproduction steps

1. Browse to the lego doc page : https://go-acme.github.io/lego/dns/cloudxns/
2. Click API documentation

You should be taken to a working DNS provider but instead are taken to a "buy this domain" page

### Version of lego

```console
lego version 4.19.2 linux/amd64
```

### Logs

N/A

_No response_

ldez:
Fixes #2323


dmke (changes requested):
May I propose a more radical approach where `cloudxns` will be reduced to something like

```go
// keep consts, type Config, and type DNSProvider

var ErrDeadProvider = errors.New("cloudxns: provider has shut down")

    return nil, ErrDeadProvider
}

func NewDNSProviderConfig(*Config) (*DNSProvider, error) {
    return nil, ErrDeadProvider
}

func (d *DNSProvider) Present(string, string, string) error {
    return ErrDeadProvider
}

func (d *DNSProvider) CleanUp(string, string, string) error {
    return ErrDeadProvider
}
```

When their website/domain is already offline, I don't see why we need to keep any code for it.

This way, users get an early and explicit reason ("provider has shut down" vs. "Get <URL>: context deadline exceeded")

providers/dns/cloudxns/cloudxns.go
```diff
@@ -2,28 +2,11 @@
...
-const (
-	envNamespace = "CLOUDXNS_"
-
-	EnvAPIKey    = envNamespace + "API_KEY"
-	EnvSecretKey = envNamespace + "SECRET_KEY"
-
-	EnvTTL                = envNamespace + "TTL"
-	EnvPropagationTimeout = envNamespace + "PROPAGATION_TIMEOUT"
-	EnvPollingInterval    = envNamespace + "POLLING_INTERVAL"
-	EnvHTTPTimeout        = envNamespace + "HTTP_TIMEOUT"
```
dmke, line 26: Removing these would be a breaking change, let's keep them for now.

dmke (changes requested):

dmke (approved):

PR #2330, opened 2024-11-08
rfc2136: add support for tsig-keygen generated file

### LINKED ISSUE #2329: rfc2136: add support for a new environment variable `RFC2136_TSIG_FILE`
### Welcome

### How do you use lego?

Binary

### Detailed Description

For the DNS provider [rfc2136](https://go-acme.github.io/lego/dns/rfc2136/index.html), rather that having to parse a key ourselves to extract the key name (i.e. `RFC2136_TSIG_KEY`), algorithm (i.e. `RFC2136_TSIG_ALGORITHM`) and secret (i.e. `RFC2136_TSIG_SECRET`), would it be possible to provide a new environment variable such as `RFC2136_TSIG_FILE` which would be the path to the key file generated by `tsig-keygen`?

This  way, lego would parse the file itself.

ldez:
This is a naive parsing of the file because the format is specific to bind9 tooling, and I didn't find a valid parser.

Fixes #2329


providers/dns/rfc2136/internal/tsigkey.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	key := &Key{}
+	err = mapstructure.Decode(data, key)
+	if err != nil {
+		return nil, fmt.Errorf("decode key: %w", err)
+	}
+
+	return key, nil
+}
```
dmke, line 74: Using `mapstructure` is a bit overkill. Here's the same code without it:

```suggestion
type Key struct {
	Name      string
	Algorithm string
	Secret    string
}

// ReadTSIGFile reads TSIG key file generated with `tsig-keygen`.
func ReadTSIGFile(filename string) (*Key, error) {
	file, err := os.Open(filename)
	if err != nil {
		return nil, fmt.Errorf("open file: %w", err)
	}

	defer func() { _ = file.Close() }()

	data := make(map[string]string)

	var (
		key Key
		read bool
	)

	scanner := bufio.NewScanner(file)
	for scanner.Scan() {
		line := strings.TrimSpace(strings.TrimSuffix(scanner.Text(), ";"))
		if line == "" {
			continue
		}

		if read && line == "}" {
			break
		}

		fields := strings.Fields(line)

		switch {
		case fields[0] == "key":
			read = true

			if len(fields) != 3 {
				return nil, fmt.Errorf("invalid key line: %s", line)
			}

			key.Name = safeUnquote(fields[1])

		case !read:
			continue

		default:
			if len(fields) != 2 {
				continue
			}

			k, v := safeUnquote(fields[0]), safeUnquote(fields[1])
			switch k {
			case "algorithm":
				key.Algorithm = v
			case "secret":
				key.Secret = v
			default:
				return fmt.Errorf("unexpected keyword: %s", k)
			}
		}
	}

	return key, nil
}
```

providers/dns/rfc2136/internal/tsigkey.go
```diff
@@ -0,0 +1,86 @@
...
+
+func safeUnquote(v string) string {
+	if v == "" {
+		return v
+	}
+
+	if len(v)-1 != 0 && v[0] == '"' && v[len(v)-1] == '"' {
+		return v[1 : len(v)-1]
+	}
```
dmke, line 83: `if len(v)-1 != 0` is a head scratcher. This is easier to understand:

```suggestion
	if len(v) < 2 {
		// empty or single character string
		return v
	}
	
	if v[0] == '"' && v[len(v)-1] == '"' {
		// string wrapped in quotes
		return v[1 : len(v)-1]
	}
```

dmke (changes requested):
Some minor picks, but otherwise LGTM.

dmke (approved):

PR #2335, opened 2024-11-10
feat: skip the TLS verification of the ACME server

ldez:
Adds an option to skip the TLS verification of the ACME server.

Fixes https://github.com/go-acme/lego/discussions/2334


dmke (changes requested):
The flag should be named `--tls-skip-verify`, as HTTP has nothing to verify.

dmke (approved):

PR #2339, opened 2024-11-10
docs: update least privilege instructions for Cloudflare

### LINKED ISSUE #2048: Support passing a cloudflare zone id instead of a zone read key
### Welcome

### How do you use lego?

Through Traefik

### Detailed Description

The acme.sh configuration for Cloudflare takes a zone scoped API key and the zone id. Ref: https://github.com/acmesh-official/acme.sh/wiki/dnsapi#i-single-dns-zone

Lego takes a zone scoped API key and another key with read across all zones. It's unclear why adding a second API key is necessary and contrary to the lego docs, this is a bit shy of least-privilege. It would be nice to add a CF_ZONE_ID environment variable for this.

I'm not sure if there's something I'm missing in understanding how the lego implementation differs from ACME on this.

ref: https://github.com/go-acme/lego/issues/984#issuecomment-539188212

joshka:
The least privilege instructions for setting up Cloudflare API keys were over-granting permissions. This would allow lego to read all the zones in the account. This was unnecessary. It only requires read access to the individual zones.

Closes #2048

---
note - I'm unsure if this is the correct file to edit, or if it's the docs file (or both)

### DIFF
--- a/docs/content/dns/zz_gen_cloudflare.md
+++ b/docs/content/dns/zz_gen_cloudflare.md
@@ -98,12 +98,13 @@ Then pass the API token as `CF_DNS_API_TOKEN` to Lego.
 **Alternatively,** if you prefer a more strict set of privileges,
 you can split the access tokens:
 
-* Create one with *Zone / Zone / Read* permissions and scope it to all your zones.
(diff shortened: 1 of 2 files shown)


ldez (approved):
LGTM

joshka:
Thanks:)

PR #2134, opened 2024-03-09
chore: add snap to release packages

gruyaume:
# Description

The LEGO snap makes it easy to install and use LEGO on any Linux distribution that supports snaps. This PR adds the snap to go releaser to automatically build and publish the snap. 

## Usage

```bash
sudo snap install lego
sudo lego \
  --email="you@example.com" \
  --domains="example.com" \
  --server=https://acme-staging-v02.api.letsencrypt.org/directory \
  --http \
  --http.port :8080 \
  run
```

**Note**: The LEGO snap can only write to the `/var/snap/lego/common/.lego` directory.

## Confinement

The snap is strictly confined, meaning it only has access to what it requires to work and for LEGO that means [network binding](https://snapcraft.io/docs/network-bind-interface). 

## Snap publishing

For the snap publishing to succeed, it will be necessary to do the following before merging this PR:
1. Create a snapcraft account
2. Make a request in the [snapcraft forum](https://forum.snapcraft.io/c/store-requests/) to have the ownership of the `lego` snap name switched from myself to this new account. I can make the request myself once we have step 1 completed. 
3. Generate a snapcraft login token attached to this specific snap with a reasonable ttl
4. Place this token in a GitHub secret named `SNAPCRAFT_STORE_CREDENTIALS`

Once all of those are completed, I will move the PR from draft to live.

## Channel

With the current state of the change proposed here, the snap will be automatically published to the `edge` channel. This means that promotion from edge -> beta -> candidate -> stable would have to be done manually through the Snapcraft UI.

Related to https://github.com/go-acme/lego/discussions/2133

PR #2348, opened 2024-11-13
chore: restore snap packaging

ldez:
The problem was the command path.

Related to #2134


dmke (commented):
I'm a little confused. The commit message says "remove", but the change actually allows publishing a Snap package?

dmke (approved):

PR #2349, opened 2024-11-13
chore: fix snap name

ldez:
I feel (I didn't test it) like I should follow the same configuration as goreleaser:

https://github.com/goreleaser/goreleaser/blob/efe8b7089e55ac2de54238b699f2ba18cc339d5f/.goreleaser.yaml#L353C12-L354C85


dmke (approved):
I can't test this, but that looks reasonable.

PR #2355, opened 2024-11-18
feat: add --force-cert-domains flag to renew

### LINKED ISSUE #1532: Renew only specified domains
### Welcome

### How do you use lego?

Binary

### Detailed Description

When using lego renew, one should be able to tell lego to only renew the specified domains, and drop any domains not specified.

This would make managing certs much easier as one'd be able to just use systemd for it, instead of remembering the run command and all that. Just add/remove -d's and call it a day, then run the renewal unit.

m1cr0man:
Closes #1532

When changing the --domains values, the certificate will not be refreshed to update the SANs appropriately.

This change introduces a simple flag to check + enforce that the domains specified match exactly those in the certificate during renewal.

I couldn't see a straightforward way to contribute a unit test for this, but I performed a full integration test using the NixOS ACME integration test suite. You can run that test locally with this command:

`nix run github:m1cr0man/nixpkgs/lego-domain-testing#nixosTests.acme`

### DIFF
--- a/cmd/cmd_renew.go
+++ b/cmd/cmd_renew.go
@@ -26,6 +26,7 @@ const (
 	flgReuseKey               = "reuse-key"
 	flgRenewHook              = "renew-hook"
 	flgNoRandomSleep          = "no-random-sleep"
+	flgForceCertDomains       = "force-cert-domains"
 )
 
 const (
@@ -52,6 +53,9 @@ func createRenew() *cli.Command {
 			}
 			if !hasDomains && !hasCsr {
 				log.Fatal("Please specify --%s/-d (or --%s/-c if you already have a CSR)", flgDomains, flgCSR)
+			}
+			if ctx.Bool(flgForceCertDomains) && hasCsr {
+				log.Fatal("--%s only works with --%s/-d, --%s/-c doesn't support this option.", flgForceCertDomains, flgDomains, flgCSR)
 			}
 			return nil
 		},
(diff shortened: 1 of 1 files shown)


m1cr0man:
It was suggested to me to name this option `--replace-cert-domains` rather than `--overwrite-domains`. I think it makes the purpose + functionality of the flag more obvious. If there's no disagreement I will update this PR tonight.

m1cr0man:
Done but I didn't squash it, as I would have to squash Idez' changes into mine. Can we squash merge on submit?

ldez:
The changes should be addressed through commits to ease the review, so no squash.

I merge the PRs with the "squash and merge" button, so in the end, everything will be squashed, no need to "fixup".

ldez:
My nickname is LDEZ.

m1cr0man:
> My nickname is LDEZ.

Oh sorry, I should have double checked!

ldez:
I feel like `replace` is more confusing than `override`.

Because it will not replace something inside the certificate.

:thinking: maybe something `force-cert-domain`

m1cr0man:
> I feel like `replace` is more confusing than `override`.
> 
> Because it will not replace something inside the certificate.
> 
> 🤔 maybe something `force-cert-domain`

I get you, because it's not an in-place change, it does require a renew.

I'll switch it to force.

ldez:
Not a major problem but you squashed all the commits, I cannot see the changes since the previous commits, so I should review all the code once again.

https://github.com/go-acme/lego/pull/2355#issuecomment-2489655147

m1cr0man:
I figured it would make it easier to review than having 3 commits for renames and 1 for indentation.

ldez (approved):
LGTM

PR #2358, opened 2024-11-21
Prepare release v4.20.3

ldez:
It's mainly to test the publication of the snap and the AUR.


ldez:
Good news, everything works as expected :tada:

ldez:
:thinking: It seems there several channels (I was not aware of that), and the snap has been published inside the `edge` channel.

ldez:
so I will add the `stable` channel to the configuration and create a new release.

PR #2359, opened 2024-11-21
Prepare release v4.20.4

ldez:
Related to https://github.com/go-acme/lego/pull/2358#issuecomment-2491426419

After checking the [goreleaser doc](https://goreleaser.com/customization/snapcraft/), There is not need for an explicit definition of the channels.


ldez:
I will merge the PR because this is just a minor change.

Issue #1546, opened 2021-12-17
Check if certificates needs renewal before create client


oott123:
### Welcome

- [X] Yes, I've searched similar issues on GitHub and didn't find any.

### How do you use lego?

Binary

### Detailed Description

lego will create a client and log in first, and then check if there are certificates to renew.

If there aren't any certificates renewal required, then no network requests required, so the created client can lead to unnecessary requests, which is not friendly to both users and ACME servers.

If we can check certificates expiry date before any network requests made, we can reduce the resources used by lego.

PR #2366, opened 2024-12-01
feat: Change renew rc when cert is still valid

m1cr0man:
Sometimes the lego Client may fail to contact the CA due to some temporary networking issue on the host. The certificate may otherwise still be valid, and a user may want to ignore this issue.

This change modifies the renew service so that if the lego client is nil, but the certificate is not expired, the exit code of the program will be 2 instead of 1. Crucially it does not change _if_ lego exits, only the exit code itself.

In order to implement this, I had to change the setup function to remove a usage of `log.Fatalf` which in turn required me to introduce explicit `os.Exit` calls in some commands. Although this was the lowest patch delta solution, I also considered refactoring the setup entirely to remove `log.Fatalf`, and return setup errors instead.

As with my previous PR, I performed a full integration test using the NixOS ACME integration test suite. You can run that test locally with this command:

`nix run github:m1cr0man/nixpkgs/lego-offline-renewal-test#nixosTests.acme`

Here's some example log output with this change implemented, sampled from a system with the network disabled:

```
$ lego --accept-tos --path . -d http.example.test --email admin@example.test --key-type ec256 --http --http.port :80 --server https://acme.test/dir renew --no-random-sleep --force-cert-domains --days 30
2024/11/30 23:33:00 [WARN] Could not create client: get directory at 'https://acme.test/dir': Get "https://acme.test/dir": dial tcp: lookup acme.test: Temporary failure in name resolution
2024/11/30 23:33:00 [http.example.test] The certificate expires in 1825 days, the number of days defined to perform the renewal is 30: no renewal.
2024/11/30 23:33:00 [WARN] [http.example.test] Failed to check ARI as there is no client. Check previous error messages for possible initialization issues.
$ echo $?
2
```

### DIFF
--- a/cmd/cmd_renew.go
+++ b/cmd/cmd_renew.go
@@ -124,7 +124,9 @@ func createRenew() *cli.Command {
 
 func renew(ctx *cli.Context) error {
 	account, client := setup(ctx, NewAccountsStorage(ctx))
-	setupChallenges(ctx, client)
+	if client != nil {
+		setupChallenges(ctx, client)
+	}
 
 	if account.Registration == nil {
 		log.Fatalf("Account %s is not registered. Use 'run' to register a new account.\n", account.Email)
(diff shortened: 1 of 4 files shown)


m1cr0man:
To note, this at least allows users to resolve #1280 (which stems from https://github.com/NixOS/nixpkgs/issues/85794#issuecomment-721408110) and #1546 by checking the return code in their specific situations.

ldez:
Hello,

I understand your problem but your proposal is not something I want to follow.
Can you open an issue?

m1cr0man:
Sure ok

PR #2373, opened 2024-12-03
inwx: delete only the TXT record related to the DNS challenge

jhbruhn:
The original implementation deleted all TXT records on the domain. That was not very nice. I took the way the IONOS Implementation is done to only delete the newly generated DNS Challenge records.

### DIFF
--- a/providers/dns/inwx/inwx.go
+++ b/providers/dns/inwx/inwx.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"time"
-	"strconv"
 
 	"github.com/go-acme/lego/v4/challenge"
 	"github.com/go-acme/lego/v4/challenge/dns01"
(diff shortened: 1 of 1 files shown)


providers/dns/inwx/inwx.go
```diff
@@ -182,9 +183,11 @@ func (d *DNSProvider) CleanUp(domain, token, keyAuth string) error {
 
 	var lastErr error
 	for _, record := range response.Records {
-		err = d.client.Nameservers.DeleteRecord(record.ID)
-		if err != nil {
-			lastErr = fmt.Errorf("inwx: %w", err)
+		if record.Content == strconv.Quote(challengeInfo.Value) {
```
ldez, line 186: are you sure the record content has quotes?

ldez (changes requested):

providers/dns/inwx/inwx.go
jhbruhn: You're right, I checked the API Docs and it is noted in double-quoted strings

ldez: OK, can you run lego with your PR and check that the TXT records are deleted?

ldez:
FYI, the records are already filtered on `_acme-challenge.example.com`.
I think you are using a CNAME, so the effective name should be something "unexpected".

providers/dns/inwx/inwx.go
jhbruhn: I have just tested the implementation successfully. The reply did indeed not contain quotes.

ldez (approved):
LGTM

jhbruhn:
Thanks for your improvements!

PR #2376, opened 2024-12-05
chore: shared deref and pointer functions

ldez:
I created a package `ptr` to handle `pointer` and `deref` functions instead of having several implementations.


dmke (approved):
:heart:

PR #2389, opened 2025-01-03
feat: add hook-timeout to run and renew commands

bossm8:
Add the option `--run-hook-timeout` and `--renewal-hook-timeout` which lets you configure how long lego should wait for the hook to finish. The default of `120` seconds may be too short in some cases.

Background:

We have a setup in which we run lego periodically with a systemd service/timer combo.
When we execute the run/renewal hook we have multiple tasks which we would execute with the hook script, those tasks may in some cases exceed the 120s timeout. The option added in this PR lets us configure this timeout and successfully finish all tasks with the script.

### DIFF
--- a/cmd/cmd_renew.go
+++ b/cmd/cmd_renew.go
@@ -110,10 +110,10 @@ func createRenew() *cli.Command {
 				Name:  flgRenewHook,
 				Usage: "Define a hook. The hook is executed only when the certificates are effectively renewed.",
 			},
-			&cli.UintFlag{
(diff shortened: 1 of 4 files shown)


ldez (approved):
Thank you :+1: 

Thank you for adding your use case, this is  appreciated :heart: 

I changed the type to use a duration: the syntax is more accurate because the need is more related to minutes (ex: `3m`) than seconds (ex: `180`).

Note: when you create a PR, it is better to use a dedicated branch instead of the main branch of the fork.

PR #2392, opened 2025-01-08
acme-dns: add option to use an existing account

rbicker:
While using acme-dns, add the option to use an existing account in addition to registering one per-domain.

Background:
We currently have a use case where we are required to have one acme-dns account per customer instead of having one per domain.

Thanks!

### DIFF
--- a/providers/dns/acmedns/acmedns.go
+++ b/providers/dns/acmedns/acmedns.go
@@ -22,6 +22,18 @@ const (
 	// EnvStoragePath is the environment variable name for the ACME-DNS JSON account data file.
 	// A per-domain account will be registered/persisted to this file and used for TXT updates.
 	EnvStoragePath = envNamespace + "STORAGE_PATH"
+	// EnvSubdomain is the environment variable name for the ACME-DNS subdomain
+	// associated with an already registered account.
+	// (e.g. 8e5700ea-a4bf-41c7-8a77-e990661dcc6a)
+	EnvSubdomain = envNamespace + "SUBDOMAIN"
+	// EnvUsername is the environment variable name for the ACME-DNS username
+	// associated with an already registered account.
+	// (e.g. c36f50e8-4632-44f0-83fe-e070fef28a10)
+	EnvUsername = envNamespace + "USERNAME"
+	// EnvPassword is the environment variable name for the ACME-DNS password
+	// associated with an already registered account.
+	// (e.g. htB9mR9DYgcu9bX_afHF62erXaH2TS7bg9KW3F7Z)
+	EnvPassword = envNamespace + "PASSWORD"
 )
 
 var _ challenge.Provider = (*DNSProvider)(nil)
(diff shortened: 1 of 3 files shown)


ldez:
Hello,

you cannot change the public function signatures, this is a breaking change of the API.

What is the problem with having one account per domain?

rbicker:
Hi

Thank you for your feedback. 

To provide you with some extra background information: We are running our own custom implementation of an acme-dns server, which does not provide the /register endpoint, as registration requires some extra steps that also lead to some automations for our customers. For the other acme clients that our customers currently use, it is possible to use pre-registered acme-dns accounts. 2 examples:

- posh-ACME: https://poshac.me/docs/latest/Plugins/AcmeDns/#advanced-pre-registration-and-cname-creation
- acme.sh: https://github.com/acmesh-official/acme.sh/wiki/dnsapi#45-use-acme-dns-api

Unfortunately, for traefik, this is currently not possible. We can setup some cronjobs that handle the certificates outside of traefik using acme.sh for example but this does feel like a hack.

Could please point out to me where I went wrong with my implementation? 
I was able to run `make test` and `make build` without running into any issues. I did not change the signature of `NewDNSProvider()`, I have only changed the signature of NewDNSProviderClient:

```
// NewDNSProviderClient creates an ACME-DNS DNSProvider with the given acmeDNSClient
// and either a goacmedns.Account or a goacmedns.Storage.
func NewDNSProviderClient(client acmeDNSClient, account *goacmedns.Account, storage goacmedns.Storage) (*DNSProvider, error)
```

 I was under the impression that this function was provider-specific, because the setup looks different for most providers - the `NewDNSProviderClient` mostly does not exist, the providers that I had looked at often use something like `NewDNSProviderConfig(config *Config) (*DNSProvider, error) `, with a provider-specific function?

 I am sorry if I missunderstood. I could use the initial signature for NewDNSProviderClient and just parse the environment variables within NewDNSProviderClient if you would prefere it this way. What do you think?

Thanks

ldez:
> I have only changed the signature of NewDNSProviderClient
> ...
> I was under the impression that this function was provider-specific

This function is exposed so you can't change the signature.

---

The usage of only one account information doesn't seem right to me because it requires one run per account.
`posh-ACME` and `acme.sh` are CLI only, but lego is a CLI and a lib, so being able to handle only one account is not something that fits with the lego approach.

I think a better approach is to use another implementation of the `Storage` or something with account mapping.
I will see what I can do.

rbicker:
I understand, thanks for the insight. I have added another commit that reverts the change to the signature and uses a custom storage implementation to store one single account. Is this what you had in mind? Thanks

ldez:
This is a good idea but this is still focused on one account only.

I created locally an HTTP implementation of the Storage: instead of reading a file, it calls an HTTP server.

- `Fetch` -> `GET <BaseURL>/<domain>`
- `FetchAll` -> `GET <BaseURL>`
- `Put` -> `POST <BaseURL>/<domain>` with the account as payload

WDYT?

ldez:
I created another implementation, a "lazy" file storage: instead of reading the account mapping one time, it read the file at each call.

WDYT?

rbicker:
Ah that is a nice way of decoupling account storage! 

Yes I think that would work for our use case (and for others!), we could just return the same account all the time. I assume it will turn non-2xx status into errors? This way, we can also return an error on "POST <BaseURL>/<domain>" for all calls. How do you handle "Save"?

ldez:
Take a look to #2393, this is a draft

rbicker:
I am closing this in favor of #2393

PR #2393, opened 2025-01-08
acme-dns: HTTP storage

ldez:
- `Fetch` -> `GET <BaseURL>/<domain>` and returns an account.
- `FetchAll` -> `GET <BaseURL>` and returns the domain/account mapping.
- `Put` -> `POST <BaseURL>/<domain>` with the account as payload.
- `Save` -> Noop

Related to #2392


rbicker:
Thanks, looks good to me! 

Just one minor thing, when I look at the `goacmedns.Storage` interface, it specifies that only a call to `Save` should persist the data.

```
// Storage is an interface describing the required functions for an ACME DNS
// Account storage mechanism.
type Storage interface {
	// Save will persist the `Account` data that has been `Put` so far
	Save() error
	// Put will add an `Account` for the given domain to the storage. It may not
	// be persisted until `Save` is called.
	Put(string, Account) error
	// ...
}
```

Your implementation conflicts with this. 
I am not sure how to handle this, maybe the http_storage should just send an empty POST request to a /save url?

ldez:
I will just remove the comment :smile_cat:

rbicker:
works for me :-)

BTW: Thank you very much for your work, it is much appreciated!

ldez:
@rbicker FYI you can use this implementation with Traefik now, because you just need to implement `Fetch` on your server to return the right account based on domain.

rbicker:
> > Unfortunately, for traefik, this is currently not possible. We can setup some cronjobs that handle the certificates outside of traefik using acme.sh for example but this does feel like a hack.
> > [#2392 (comment)](https://github.com/go-acme/lego/pull/2392#issuecomment-2577609691)
> 
> @rbicker FYI you can use this implementation with Traefik now because you just need to implement `Fetch` on your server to return the right account based on domain. You can create something based on customers (domain-> customer -> account).

Yes that is perfect. We can even implement some additional business logic on our account storage server, do basic monitoring, ...

Thank you so much!

providers/dns/acmedns/internal/readme.md
```diff
@@ -0,0 +1,70 @@
...
+    "server_url": "https://example.com"
+  },
+  "bar.example.com": {
+    "fulldomain": "bar.example.com",
+    "subdomain": "bar",
+    "username": "user",
+    "password": "secret",
+    "server_url": "https://example.com"
+  }
+}
```
dmke, line 53:
```suggestion
[
  "foo.example.com": {
    "fulldomain": "foo.example.com",
    "subdomain": "foo",
    "username": "user",
    "password": "secret",
    "server_url": "https://example.com"
  },
  "bar.example.com": {
    "fulldomain": "bar.example.com",
    "subdomain": "bar",
    "username": "user",
    "password": "secret",
    "server_url": "https://example.com"
  }
]
```

dmke (changes requested):
LGTM, just one minor oversight.

providers/dns/acmedns/internal/readme.md
ldez: it's a map not a slice

dmke (approved):

providers/dns/acmedns/internal/readme.md
dmke: Indeed, my bad :see_no_evil:

ldez:
@rbicker I hope you enjoyed my work, please consider donating or asking your company to do so.
My only source of revenue is donations.
This will be appreciated, thank you :heart:

rbicker:
@ldez: Thanks! I have a meeting with my company this Friday so I will ask for a donation.

PR #2396, opened 2025-01-10
netcup: remove TTL option

### LINKED ISSUE #2395: netcup: improve documentation
### Welcome

### How do you use lego?

Docker image

### Detailed Description

Hello, I would like to suggest improvements for the documentation of the Netcup provider:
- The Netcup API does not allow to set the TTL for a single DNS entry. It looks to me as if `NETCUP_TTL` does not have any effect. If that is correct, it should be removed from the docs (and probably from the source es well).
- For the environment variables `NETCUP_HTTP_TIMEOUT`, `NETCUP_POLLING_INTERVAL` and `NETCUP_PROPAGATION_TIMEOUT`, what are the units? Are these all in seconds?

ldez:
Fixes #2395

### DIFF
--- a/providers/dns/netcup/internal/types.go
+++ b/providers/dns/netcup/internal/types.go
@@ -72,7 +72,6 @@ type DNSRecord struct {
 	Destination  string `json:"destination"`
 	DeleteRecord bool   `json:"deleterecord,omitempty"`
 	State        string `json:"state,omitempty"`
-	TTL          int    `json:"ttl,omitempty"`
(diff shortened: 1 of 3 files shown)


dmke (changes requested):
Not sure if that preview is generally available, or if only I am seeing this (the "preview" badge irritates me), but there are some leftovers:

image

dmke (approved):
I've been too slow :stuck_out_tongue:

PR #2398, opened 2025-01-13
cli: add environment variable for specifying email

dmayle:
Extend the use of LEGO_EMAIL in the renew command to all uses of the --email flag


ldez:
`LEGO_ACCOUNT_EMAIL` is already used by the hook, so I renamed it to `LEGO_EMAIL`.

ldez (approved):
LGTM

PR #2422, opened 2025-02-05
feat: add delay option for HTTP challenge

### LINKED ISSUE #2419: Add delay in between when local web server starts and request is sent to CA
### Welcome

### How do you use lego?

Binary

### Detailed Description

Hi,

I’m using Lego to request certificates for a container in OpenShift. In front of the container, I have a service (load balancer) with a monitor that checks if the pod is ready to accept connections, checking readiness every 5 seconds. This setup is causing an issue with Lego. When the "Lego container" sends the ACME request to the CA, it starts the local web server and sends the request to the CA within a very short time frame. The ACME CA attempts to validate the request before the load balancer's monitor has marked the pod as "listening on port 80," causing the initial request to fail.

For some reason, subsequent retries from the CA don’t reach the load balancer at all. I’m not sure if the CA is even retrying, but I’m investigating this separately with the CA vendor.

If I disable the monitor (i.e., allow traffic to be sent to the pod regardless of whether it’s listening), the ACME request works perfectly every time.

A potential "fix" for these kinds of situations would be to introduce an initial delay between when the web server starts listening and when the request is sent to the CA, ensuring the pod is fully marked as ready before validation begins.

Best regards,
Jim

ldez:
Fixes #2419

### DIFF
--- a/challenge/http01/http_challenge.go
+++ b/challenge/http01/http_challenge.go
@@ -2,6 +2,7 @@ package http01
 
 import (
 	"fmt"
+	"time"
 
 	"github.com/go-acme/lego/v4/acme"
 	"github.com/go-acme/lego/v4/acme/api"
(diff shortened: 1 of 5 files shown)


dmke (approved):
LGTM

PR #2423, opened 2025-02-06
feat: option to set CSR emails

### LINKED ISSUE #2417: Add option to add custom attributes to ACME CSR
### Welcome

### How do you use lego?

Through Traefik

### Detailed Description

Hi there,

I ran into the following issue: the company CA has a policy, that any certificate requests need to contain the attribute emailAddress. The attempts to use traefik's ACME mechanism failed, because this attribute is missing in the CSR it sent to the CA via ACME. Debugging the CSR I found that there is no attribute at all except the SAN:
```
Certificate Request:
    Data:
        Version: 1 (0x0)
        Subject: CN = whoami.local.domain
        Subject Public Key Info:
            Public Key Algorithm: rsaEncryption
                Public-Key: (4096 bit)
                Modulus:
                    00:b2:37:c9:1a:73:16:6f:b3:44:4c:27:40:63:b1:
                    ...
                    58:be:60:27:21:38:2b:08:90:4c:08:69:d0:f1:09:
                    d8:51:f1
                Exponent: 65537 (0x10001)
        Attributes:
            Requested Extensions:
                X509v3 Subject Alternative Name: 
                    DNS:whoami.local.domain
    Signature Algorithm: sha256WithRSAEncryption
    Signature Value:
        1f:42:af:c8:7b:5e:57:2f:ea:e4:4d:0b:18:76:8c:81:6e:21:
        ...
        96:75:da:32:c4:ce:bc:c8
```

It would be great If there could be more attributes added. 

I just copied this issue over from https://github.com/traefik/traefik/issues/11432. I will be more than happy to contribute a PR if you think this is a good idea. Thanks!

ldez:
Fixes #2417

### DIFF
--- a/certcrypto/crypto.go
+++ b/certcrypto/crypto.go
@@ -135,10 +135,26 @@ func GeneratePrivateKey(keyType KeyType) (crypto.PrivateKey, error) {
 	return nil, fmt.Errorf("invalid KeyType: %s", keyType)
 }
 
+// Deprecated: uses [CreateCSR] instead.
 func GenerateCSR(privateKey crypto.PrivateKey, domain string, san []string, mustStaple bool) ([]byte, error) {
+	return CreateCSR(privateKey, CSROptions{
+		Domain:     domain,
+		SAN:        san,
+		MustStaple: mustStaple,
+	})
+}
+
+type CSROptions struct {
+	Domain         string
+	SAN            []string
+	MustStaple     bool
+	EmailAddresses []string
+}
+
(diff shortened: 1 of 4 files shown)


dmke (approved):
LGTM.

It would be nice if we had an ACME server which preserves the email address in the subject, so that the e2e test could verify that.

I guess Pebble not outright rejecting the CSR ought to be enough for now...

ldez:
Yes, I also checked the log, and there is nothing.

PR #2426, opened 2025-02-06
Avoid nil pointer in error interface in checkOrderStatus

andreyrusanov:
I caught a panic in the affected piece of code:

I had `acme.StatusInvalid` along with `order.Error` is nil (due to rate limitting error). Despite `order.Error` is of type `*ProblemDetails`, when it's returned as an error, it gets wrapped in an error interface. The key is that a nil pointer of a concrete type (*ProblemDetails) becomes a non-nil error interface when returned.
Then, when `Error()` method is called later on in scope of this `For`: https://github.com/go-acme/lego/blob/e644196bfc422de8590c4cd54a2ddf99bb46df0e/certificate/certificates.go#L353, it causes panic.

### DIFF
--- a/certificate/certificates.go
+++ b/certificate/certificates.go
@@ -687,6 +687,9 @@ func checkOrderStatus(order acme.ExtendedOrder) (bool, error) {
 	case acme.StatusValid:
 		return true, nil
 	case acme.StatusInvalid:
+		if order.Error == nil {
+			return false, nil
+		}
 		return false, order.Error
 	default:
 		return false, nil


ldez:
can you provide the stack trace?

ldez:
In fact your explanation is wrong, so your PR cannot be right.

https://go.dev/play/p/SM2E-co2M_A

I recommend opening an issue with the stack trace.

andreyrusanov:
@ldez

In your go.dev snippet you have a mistake: 
it should be `data := `{"status":"invalid","identifiers":null, "error":null}`  instead of `data := `{"status":"valid","identifiers":null, "error":null}` at line 62. If you change it, you will see
```
panic: value method main.ProblemDetails.Error called using nil *ProblemDetails pointer

goroutine 1 [running]:
main.(*ProblemDetails).Error(0x4e169f?)
	<autogenerated>:1 +0x26
main.main()
	/tmp/sandbox3868371876/prog.go:70 +0x131
```

 unfortunately I lost original stack trace because it was in my IDE. I will try to reproduce it tomorrow

ldez:
yes, but your explanation is still wrong, the problem is related to `Error:(*acme.ProblemDetails)(nil)` and not the `error` interface.

andreyrusanov:
Here is corrected snippet - https://go.dev/play/p/jRYyxHX6KOS

PR #2428, opened 2025-02-06
fix: rewrite status management

### LINKED ISSUE #2427: Nil pointer in error interface in checkOrderStatus
### Welcome

- [x] Yes, I'm using a binary release within 2 latest releases.
- [x] Yes, I've included all information below (version, config, etc).

Rate limit error should not rise an exception

### What did you see instead?

Panic, nil pointer dereference. I will try to reproduce it later on, so far it can be emulated (based on discussion here: https://github.com/go-acme/lego/pull/2426)

### How do you use lego?

Library

### Reproduction steps

See https://go.dev/play/p/WQWmhF8yF8N

### Version of lego

```console
v4.21.0
```

### Logs

<details>

```console
panic: value method main.ProblemDetails.Error called using nil *ProblemDetails pointer

goroutine 1 [running]:
main.(*ProblemDetails).Error(0x4e169f?)
	<autogenerated>:1 +0x26
main.main()
	/tmp/sandbox1881138421/prog.go:70 +0x131
```

</details>

<details>

```console
$ go version && go env
# paste output here
```

</details>

ldez:
Fixes #2427

### DIFF
--- a/acme/commons.go
+++ b/acme/commons.go
@@ -200,6 +200,14 @@ type Order struct {
 	Replaces string `json:"replaces,omitempty"`
 }
 
+func (r *Order) Err() error {
+	if r.Error != nil {
+		return r.Error
+	}
+
+	return nil
+}
+
 // Authorization the ACME authorization object.
 // - https://www.rfc-editor.org/rfc/rfc8555.html#section-7.1.4
 type Authorization struct {
@@ -283,6 +291,14 @@ type Challenge struct {
 
 	// https://www.rfc-editor.org/rfc/rfc8555.html#section-8.1
 	KeyAuthorization string `json:"keyAuthorization"`
+}
+
+func (c *Challenge) Err() error {
+	if c.Error != nil {
+		return c.Error
+	}
+
+	return nil
 }
 
 // Identifier the ACME identifier object.
--- a/acme/api/order.go
+++ b/acme/api/order.go
@@ -3,6 +3,7 @@ package api
 import (
 	"encoding/base64"
 	"errors"
+	"fmt"
 	"net"
 	"time"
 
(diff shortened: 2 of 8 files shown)


acme/commons.go
```diff
@@ -200,6 +200,14 @@ type Order struct {
 	Replaces string `json:"replaces,omitempty"`
 }
 
+func (r *Order) Err() error {
+	if r.Error != nil {
```
dmke, line 204: This method doesn't do anything useful. The expression `r.Error != nil ? r.Error : nil` can be condensed to just `r.Error`.

Do you want to add a nil check on `r`, i.e.

```suggestion
	if r != nil {
}
```

which would avoid a panic? https://go.dev/play/p/e9z35S2ydbu

acme/commons.go
```diff
@@ -285,6 +293,14 @@ type Challenge struct {
 	KeyAuthorization string `json:"keyAuthorization"`
 }
 
+func (c *Challenge) Err() error {
+	if c.Error != nil {
```
dmke, line 297: Same here :)

```suggestion
	if c != nil {
```

dmke (changes requested):

acme/commons.go
ldez: You missed the point: the panic doesn't happen because the order is nil, but because `Error` is nil.

> This method doesn't do anything useful.

This method is doing something useful.

acme/commons.go
ldez: Same here too.

ldez:
I rewrote the status management because the root of the problem is not a nil pointer, but the fact that `invalid` status can be associated with a nil error.

Also, I updated the PR with tests.

acme/commons.go
dmke: >> This method doesn't do anything useful.
>
> This method is doing something useful.

It's not being used though, how can it be useful?

I have locally removed the `(*Order) Err()` and `(*Challenge) Err()` methods, and tests still pass.

> https://go.dev/play/p/9X_m55Un5fI

That's surprising :thinking:

dmke: >>  I have locally removed the (*Order) Err() and (*Challenge) Err() methods, and tests still pass.
>
> I added a new commit yesterday, so yes, now they are unused but this is not the question.

Ah, I didn't catch that! That explains a lot :)

dmke (approved):

PR #2431, opened 2025-02-09
feat(cli): add an option to set the private key

ldez:
Closes #2429


cmd/cmd_run.go
```diff
@@ -64,6 +65,10 @@ func createRun() *cli.Command {
 				Usage:  "Set the notAfter field in the certificate (RFC3339 format)",
 				Layout: time.RFC3339,
 			},
+			&cli.StringFlag{
+				Name:  flgPrivateKey,
+				Usage: "The certificate private key. By default, the private key is generated.",
```
dmke, line 70:
```suggestion
				Usage: "The private key file for the certificate. By default, the private key is generated.",
```

or

```suggestion
				Usage: "Path to private key (in PEM encoding) for the certificate. By default, the private key is generated.",
```

dmke (changes requested):
Just a small addendum to the CLI flag, otherwise LGTM.

dmke (approved):

PR #2436, opened 2025-02-15
acme-dns: add CNAME provisioned flag for HTTP storage

skippyprime:
With the new acme-dns HTTP storage feature, it is possible to complete the challenge if the HTTP storage server will also create the CNAME records.  Currently, the acme-dns provider will always return an error indicating that the CNAME must be created.

This adds an environment configuration to tell the acme-dns provider to presume the CNAME record has been added to the DNS zone.  This allows the challenge to complete successfully when the CNAME record is created when a new DNS challenge is registered with acme-dns.  This is mainly useful when using the HTTP storage backend if the server will also create the CNAME record.

Related to (#2393)

### DIFF
--- a/providers/dns/acmedns/acmedns.go
+++ b/providers/dns/acmedns/acmedns.go
@@ -30,6 +30,11 @@ const (
 	// EnvStorageBaseURL  is the environment variable name for the ACME-DNS JSON account data.
 	// The URL to the storage server.
 	EnvStorageBaseURL = envNamespace + "STORAGE_BASE_URL"
+
+	// EnvCNAMEProvisioned is the environment variable name for the CNAME created flag.
+	// Override that the provider has provisioned the CNAME in the DNS zone. (Default: false)
+	// This is useful when using a HTTP storage backend that can update the DNS zone.
+	EnvCNAMEProvisioned = envNamespace + "CNAME_PROVISIONED"
 )
 
 var _ challenge.Provider = (*DNSProvider)(nil)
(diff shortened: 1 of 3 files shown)


ldez:
I think you can implement that inside the HTTP storage.

skippyprime:
I did, but the the acme-dns provider always returns `ErrCNAMERequired`.  This breaks the flow for many services using lego as they don't understand the context of the error for each provider, but only that the provider returned an error.

The use case I tested was with `traefik` + `acme-dns` with the HTTP storage server.  It required a restart of traefik to have it retry the DNS challenge because the `ErrCNAMERequired` error.  It doesn't understand all of the errors for each provider, just that the provider you are using returned an error.

ldez:
If the `Fetch` always returns something, the `register` will never be called.

ldez:
:thinking: I understand your context, the `Fetch` implementation is a solution but if you want to use the `register` I think a better approach is to delete this behavior the `Save` implementation.
I need to think about that.

skippyprime:
The Fetch doesn’t contain all of the information required.  It will only contain the domain attempting to be verified as a GET path parameter.  To create the CNAME, we need the acme-dns data returned from registration.

Since the Put and Save are both in the `register` function, it would not make a difference since that function will always return `ErrCNAMERequired`.

I like that the HTTP client doesn’t implement Save because then we would need to track state across requests.  I think Put is sufficient.

ldez:
I have a proposal, I will open a PR in a few seconds

ldez:
@skippyprime Could you take a look at https://github.com/go-acme/lego/pull/2437?

ldez:
Closed in favor of #2437

PR #2437, opened 2025-02-15
acme-dns: allow the HTTP storage server to create the CNAME

ldez:
This allows the HTTP storage server to control the CNAME creation.

if the `Put` returns a 201, lego expects the CNAME to be created automatically by the server.

Closes #2436

### DIFF
--- a/providers/dns/acmedns/internal/http_storage.go
+++ b/providers/dns/acmedns/internal/http_storage.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -16,6 +17,8 @@ import (
 )
 
 var _ goacmedns.Storage = (*HTTPStorage)(nil)
+
+var ErrCNAMECreated = errors.New("the CNAME has already been created")
 
 // HTTPStorage is an implementation of [acmedns.Storage] over HTTP.
 type HTTPStorage struct {
@@ -36,8 +39,13 @@ func NewHTTPStorage(baseURL string) (*HTTPStorage, error) {
 	}, nil
 }
 
-func (s *HTTPStorage) Save(_ context.Context) error {
-	return nil
+func (s *HTTPStorage) Save(ctx context.Context) error {
+	req, err := newJSONRequest(ctx, http.MethodPost, s.baseURL, nil)
+	if err != nil {
+		return fmt.Errorf("unable to create request: %w", err)
+	}
+
+	return s.do(req, nil)
 }
 
 func (s *HTTPStorage) Put(ctx context.Context, domain string, account goacmedns.Account) error {
@@ -98,6 +106,11 @@ func (s *HTTPStorage) do(req *http.Request, result any) error {
 	}
 
 	if result == nil {
+		// Hack related to `Save`.
+		if resp.StatusCode == http.StatusCreated {
+			return ErrCNAMECreated
+		}
+
 		return nil
 	}
 
--- a/providers/dns/acmedns/internal/http_storage.go
+++ b/providers/dns/acmedns/internal/http_storage.go
@@ -18,7 +18,7 @@ import (
 
 var _ goacmedns.Storage = (*HTTPStorage)(nil)
 
-var ErrCNAMECreated = errors.New("the CNAME has already been created")
+var ErrCNAMEAlreadyCreated = errors.New("the CNAME has already been created")
 
 // HTTPStorage is an implementation of [acmedns.Storage] over HTTP.
 type HTTPStorage struct {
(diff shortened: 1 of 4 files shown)


skippyprime:
I think the HTTP 201 Created logic needs to be moved to the Put function.  The save function won’t have the context of which domain it is responding to.

ldez:
The Save will be always called because of the file implementation, so an implementation of the `Save` is required.

ldez:
:thinking: I can skip the `Save` but it will be weird.

skippyprime:
The reason I suggested the check in Put vs Save is because when we have multiple domain challenges.  One the first domain challenge, everything works as intended 1. Put the domain + acme dns data and then 2. Save would create the single CNAME record.

When 2+ domains are registered, the Save would have to iterate all domain challenges stored and attempt to create the CNAME because it doesn’t have the context on Save to know which CNAME needs created.  If the Save had the domain attempting a minimum this wouldn’t be an issue.

ldez:
Can you check the new implementation?

ldez:
When I created the initial HTTP storage implementation the `Save` I wondered if I should skip the step.

My thinking was that if a DNS provider doesn't provide an API usable by lego, there is no possibility to automate the CNAME process.

So, I'm interested in more details about your context/use case, could you give me more details?

skippyprime:
I think the logic in HTTP response is way better than the environment variable approach.

While we know that there are only 2 storage implementations and the HTTP Save is a no-op, we could still call save to keep the client logic clean.  You could do:

```go
cnameCreated := errors.Is(internal.ErrCNAMECreated)

// call save and check for error

if !cnameCreated {
  // return error
}

return nil
```

Sorry for any errors in code above, on a phone.

ldez:
I cannot do that because no `ErrCNAMECreated` doesn't mean there is no error or an error.
I should handle the error first.

ldez:
I think I understand your suggestion, your snippet was a bit confusing, sorry.

Could you answer my previous question? https://github.com/go-acme/lego/pull/2437#issuecomment-2661083787

skippyprime:
The use case is for a generic ingress service for local development.

I end up working on a number of different projects (typically a mix of related micro-services or isolated projects) that either I need to access from the host browser or the services need to communicate directly.  Since most of these are separate git repos, it is ideal for them to be loosely coupled so they can run independently or together.  Rather than juggling lots of different exposed ports, it is preferred to use standard ports (80, 443) and depend of SNI or Host matching to route traffic to the correct services.

I am releasing an open source service kit (traefik + coredns + stepca + acme-dns) that developers can run locally to dynamically add routes to traefik, allow traefik to automatically obtain certificates (HTTP-01 and DNS-01 challenges).  The DNS challenges were required to support wildcard certificates.

The acme-dns HTTP storage layer was the piece that is making this possible.  I wrote a simple HTTP storage server that will save the acme-dns accounts in the same format as though you are using the file storage backend.  It will take care of registering the CNAME with coredns before it returns the HTTP 201 Created to the Put request.

ldez:
Your stack is interesting.

>  It will take care of registering the CNAME with coredns before it returns the HTTP 201 Created to the Put request.

How do you create the CNAMEs? (coredns has no API to add/remove records)
Is it a global forward?

providers/dns/acmedns/internal/http_storage.go
```diff
@@ -17,6 +18,8 @@ import (
 
 var _ goacmedns.Storage = (*HTTPStorage)(nil)
 
+var ErrCNAMECreated = errors.New("the CNAME has already been created")
```
dmke, line 21: The naming suggests that is an error *to create* a CNAME.

How about `ErrCNAMEAlreadyCreated` or `ErrDuplicateCNAME`?

dmke (changes requested):
LGTM, just a little bike shedding :)

dmke (approved):

skippyprime:
I found one other missed expectation that acme-dns would always return `ErrCNAMERequired`.  https://github.com/go-acme/lego/blob/master/providers/dns/acmedns/acmedns.go#L178

The result of `register(...)` is returned without checking if the error is set.  This causes the `UpdateTXTRecord` to never be called when the CNAME is created by the HTTP storage server.  Do you want me to open a PR to propose a fix?

ldez:
I will fix that.

I just created a release :cry:

ldez:
https://github.com/go-acme/lego/pull/2443

skippyprime:
So close.  Sorry, I was testing this last night and couldn't figure out what was preventing the solve.  Realized this morning after looking through the debug logs.

Sorry I missed this on the first pass.

ldez:
I'm still interested in your stack: https://github.com/go-acme/lego/pull/2437#issuecomment-2661244117

skippyprime:
@ldez I released the stack that uses the acme-dns http storage layer.  If you are interested, you can find the source and docs on github.

Source: https://github.com/skippyware/local-ingress
Docs: https://skippyware.github.io/local-ingress

PR #2443, opened 2025-02-17
acme-dns: continue the process when the CNAME is handled by the storage

ldez:
https://github.com/go-acme/lego/pull/2437#issuecomment-2663309250

### DIFF
--- a/providers/dns/acmedns/acmedns.go
+++ b/providers/dns/acmedns/acmedns.go
@@ -172,14 +172,14 @@ func (d *DNSProvider) Present(domain, _, keyAuth string) error {
 	// Check if credentials were previously saved for this domain.
 	account, err := d.storage.Fetch(ctx, domain)
 	if err != nil {
-		if errors.Is(err, storage.ErrDomainNotFound) {
-			// The account did not exist.
-			// Create a new one and return an error indicating the required one-time manual CNAME setup.
-			err = d.register(ctx, domain, info.FQDN)
-			if !errors.Is(err, internal.ErrCNAMEAlreadyCreated) {
-				return err
-			}
-		} else {
+		if !errors.Is(err, storage.ErrDomainNotFound) {
+			return err
+		}
+
+		// The account did not exist.
+		// Create a new one and return an error indicating the required one-time manual CNAME setup.
+		err = d.register(ctx, domain, info.FQDN)
+		if !errors.Is(err, internal.ErrCNAMEAlreadyCreated) {
 			return err
 		}
 	}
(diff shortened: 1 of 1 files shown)


providers/dns/acmedns/acmedns.go
```diff
@@ -171,14 +171,13 @@ func (d *DNSProvider) Present(domain, _, keyAuth string) error {
 
 	// Check if credentials were previously saved for this domain.
 	account, err := d.storage.Fetch(ctx, domain)
-	if err != nil {
+	if err != nil && !errors.Is(err, internal.ErrCNAMEAlreadyCreated) {
```
skippyprime, line 174: `internal.ErrCNAMEAlreadyCreated` won't get returned from `Fetch`.  The check need to be on the result of the call to `d.register(ctx, domain, info.FQDN)`.  If the result of that call is `nil`, we need to get to the `UpdateTXTReord`.

ldez: https://github.com/go-acme/lego/pull/2443#issuecomment-2663328197

skippyprime:
Well, I found one more issue (and I think this is the last) in the flow.

Since the call to `Fetch()` will return an error for a new challenge, account will be `nil`.  The `register` function will create the new account and must return that to the caller so it can be used in the `UpdateTXTRecord` call.  It is currently always `nil` for new challenges.

The `Present` function should probably look like this:

```go
func (d *DNSProvider) Present(domain, _, keyAuth string) error {
	ctx := context.Background()

	// Compute the challenge response FQDN and TXT value for the domain based on the keyAuth.
	info := dns01.GetChallengeInfo(domain, keyAuth)

	// Check if credentials were previously saved for this domain.
	account, err := d.storage.Fetch(ctx, domain)
	if err != nil {
		if !errors.Is(err, storage.ErrDomainNotFound) {
			return err
		}

		// The account did not exist.
		// Create a new one and return an error indicating the required one-time manual CNAME setup.
		account, err = d.register(ctx, domain, info.FQDN)
		if err != nil {
			return err
		}
	}

	// Update the acme-dns TXT record.
	return d.client.UpdateTXTRecord(ctx, account, info.Value)
}
```

And `register` function needs to return the new account:

```go
func (d *DNSProvider) register(ctx context.Context, domain, fqdn string) (goacmedns.Account, error) {
	newAcct, err := d.client.RegisterAccount(ctx, d.config.AllowList)
	if err != nil {
		return newAcct, err
	}

	var cnameCreated bool

	// Store the new account in the storage and call save to persist the data.
	err = d.storage.Put(ctx, domain, newAcct)
	if err != nil {
		cnameCreated = errors.Is(err, internal.ErrCNAMEAlreadyCreated)
		if !cnameCreated {
			return newAcct, err
		}
	}

	err = d.storage.Save(ctx)
	if err != nil {
		return newAcct, err
	}

	if cnameCreated {
		return newAcct, nil
	}

	// Stop issuance by returning an error.
	// The user needs to perform a manual one-time CNAME setup in their DNS zone
	// to complete the setup of the new account we created.
	return newAcct, ErrCNAMERequired{
		Domain: domain,
		FQDN:   fqdn,
		Target: newAcct.FullDomain,
	}
}
```

PR #2445, opened 2025-02-17
acme-dns: use new registred account

ldez:
Related to https://github.com/go-acme/lego/pull/2443#issuecomment-2663470866


dmke:
I think this would need a proper test case, no? This is the third change related to account creation in short time due to missed edge cases...

skippyprime:
This tests well locally.  Thank you for all of your support!

ldez:
> I think this would need a proper test case, no?

I'm currently writing tests, but I wanted to open the PR quickly.

dmke (approved):

PR #2299, opened 2024-10-11
Update Porkbun DNS Provider Configuration to New Hostname

imyourmanzi:
Porkbun will be transitioning its API hostname from `porkbun.com` to `api.porkbun.com`.  I was informed of this via email (see below) but have confirmed it on their [documentation site](https://porkbun.com/api/json/v3/documentation#apiHost).

The new hostname is live now and the old one will be disabled after November 30, 2024.

---

Screenshot source for Porkbun API hostname change email

PR #2457, opened 2025-02-26
fix(provider): Update porkbun.toml

acederberg:
porkbun.com -> api.porkbun.com.

I noticed that my requests in traefik (for the acme certresolver) were failing. After some experimentation I found that when I updated from 'porkbun.com/api/json/v3/ping' to 'api.porkbun.com/api/json/v3/ping' my requests work again.

Notice this chage in the [API documentation](https://porkbun.com/api/json/v3/documentation).

This reasoning above is verified in the following `python` snippet: 

```python
req = httpx.Request(
    method="POST",
    url="https://api.porkbun.com/api/json/v3/ping", 
    json=dict(
        apikey="...",
        secretapikey="...",
    ),
)


with httpx.Client() as client:
    res = client.send(req)
    print("Status Code:", res.status_code)
    print("Response Body:", res.content.decode())
```

### DIFF
--- a/providers/dns/porkbun/porkbun.toml
+++ b/providers/dns/porkbun/porkbun.toml
@@ -1,6 +1,6 @@
 Name = "Porkbun"
 Description = ''''''
-URL = "https://porkbun.com/"
+URL = "https://api.porkbun.com/"
 Code = "porkbun"
 Since = "v4.4.0"
 


ldez:
https://github.com/go-acme/lego/pull/2299#issuecomment-2408241087

acederberg:
I see! Thank you

ldez:
FYI Traefik uses lego v4.22, so your problem is that you are using an old version of Traefik.

PR #2460, opened 2025-02-28
edgedns: add account switch key option

hokamoto:
An Akamai EdgeDNS API client may have permission to manage multiple accounts. In such cases, specifying an account switch key allows the API client to edit DNS zones belonging to other Akamai accounts. Support for this account switch key has been implemented.
ref. https://techdocs.akamai.com/developer/docs/manage-many-accounts-with-one-api-client


ldez (approved):
LGTM

PR #2469, opened 2025-03-04
fix: kill hook when the command is stuck

ldez:
The `context.WithTimeout` doesn't kill a command when the command is stuck because `CombinedOutput` is waiting for a "pause".

The previous implementation was able to stop the hook if the script was too long, but not stuck on one element (cf test: `sleepy.sh`).

But it was not able to stop the hook if the script was waiting for a long (never-ending) process (cf test: `sleeping_beauty.sh`).

I modified the code to add a timer that kills the current process if the timeout is reached.

This implementation has another advantage: the output of the hook will be live streamed to the console.

Fixes #2468

PR #2471, opened 2025-03-05
chore: related timer with context.Done

ldez:
In fact, we don't need a timer, a go routine on `context.DOne()` is enough.

Related to #2469

### DIFF
--- a/cmd/hook.go
+++ b/cmd/hook.go
@@ -11,7 +11,6 @@ import (
 	"time"
 
 	"github.com/go-acme/lego/v4/certificate"
-	"github.com/go-acme/lego/v4/log"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


cmd/hook.go
```diff
@@ -49,13 +48,15 @@ func launchHook(hook string, timeout time.Duration, meta map[string]string) erro
...
-	defer timer.Stop()
+	go func() {
+		select {
+		case <-ctxCmd.Done():
+			if ctxCmd.Err() != nil {
+				_ = cmd.Process.Kill()
+				_ = stdout.Close()
+			}
+		}
+	}()
```
dmke, line 59: `gosimple` suggests something like

```suggestion
	go func() {
		<-ctxCmd.Done():
		if ctxCmd.Err() != nil {
			_ = cmd.Process.Kill()
			_ = stdout.Close()
		}
	}()
```

dmke (changes requested):

dmke (approved):

cmd/hook.go
```diff
@@ -49,13 +48,13 @@ func launchHook(hook string, timeout time.Duration, meta map[string]string) erro
...
 
-	timer := time.AfterFunc(timeout, func() {
-		log.Println("hook timed out: killing command")
-		_ = cmd.Process.Kill()
-		_ = stdout.Close()
-	})
-
-	defer timer.Stop()
+	go func() {
+		<-ctxCmd.Done():
```
dmke, line 52: my bad, this ought to be

```suggestion
		<-ctxCmd.Done()
```

dmke (changes requested):

cmd/hook.go
ldez: fixes :wink:

PR #2484, opened 2025-03-15
cloudflare: make base URL configurable

huncrys:
I'm trying to integrate lego with a DNS provider that implements the same API as Cloudflare. This makes it possible to integrate it with regular Cloudflare API clients by overriding the base URL.

`cloudflare-go` supports this, so I propose exposing this functionality from lego.


ldez:
No description, no issue related -> close.

huncrys:
@ldez sorry about that, added some context to the description

ldez:
What is this DNS provider with the same API as Cloudflare?

huncrys:
This is a private service right now, there are is no public-facing "branding" for it. So I understand if support this functionality falls outside the purview of this project.

ldez:
I need to think about that.

ldez:
I have several questions:
- why are you using the same API as Cloudflare? (because there are simpler APIs)
- in which context are you creating this private services? (is it commercial usage?)

ldez:
but why not using a open source DNS service?

huncrys:
The server is open source, PowerDNS, but it's API doesn't work for the backend that I'm using. We built this system for tinydns way before, so during the switch to the interface could remain the same as PowerDNS supports the tinydns format as well.

You mentioned other simpler APIs, I see lego already supports overriding the base URL for others, so I could instead build something for one of them. At first this seemed like a good idea to get support for this, but on further thought I have to admit it might be overkill just for one marginal use case :)

ldez (approved):
LGTM

Issue #2489, opened 2025-03-19
allinkl: cert renewal seems broken


fortysix2ahead:
### Welcome

- [x] Yes, I'm using a binary release within 2 latest releases.
- [x] Yes, I've searched similar issues on GitHub and didn't find any.
- [x] Yes, I've included all information below (version, config, etc).

### What did you expect to see?

I am using all-inkl.com as DNS provider. Renewing certs usually fails for the first time. When retrying for 5 or 10 times at some point it works. Don't know if this is a timeout problem, but at least there seems to be something wrong with the response (unconvertible type in the log, see below).

### What did you see instead?

An error message, see below.

### How do you use lego?

Binary

### Reproduction steps

```
lego --path path_to_config -a -m admin@example.com --dns allinkl --dns-timeout 120 --dns.resolvers ns6.kasserver.com --domains "*.example.com" renew
```

### Effective version of lego

```console
lego version 4.22.2 darwin/arm64
```

### Logs

<details>

```
2025/03/19 07:49:59 [INFO] [*.example.com] acme: renewalInfo endpoint indicates that renewal is needed
2025/03/19 07:49:59 [INFO] [*.example.com] acme: Trying renewal with 386 hours remaining
2025/03/19 07:49:59 [INFO] [*.example.com] acme: Obtaining bundled SAN certificate
2025/03/19 07:49:59 [INFO] [*.example.com] AuthURL: https://acme-v02.api.letsencrypt.org/acme/authz/XXXXXXXXX/XXXXXXXXXX
2025/03/19 07:49:59 [INFO] [*.example.com] acme: use dns-01 solver
2025/03/19 07:49:59 [INFO] [*.example.com] acme: Preparing to solve DNS-01
2025/03/19 07:50:00 [INFO] [*.example.com] acme: Trying to solve DNS-01
2025/03/19 07:50:00 [INFO] [*.example.com] acme: Checking DNS record propagation. [nameservers=ns6.kasserver.com:53]
2025/03/19 07:50:02 [INFO] Wait for propagation [timeout: 1m0s, interval: 2s]
2025/03/19 07:50:02 [INFO] [*.example.com] acme: Waiting for DNS record propagation.
2025/03/19 07:50:04 [INFO] [*.example.com] acme: Waiting for DNS record propagation.
2025/03/19 07:50:06 [INFO] [*.example.com] acme: Waiting for DNS record propagation.
[...]

PR #2490, opened 2025-03-20
allinkl: remove ReturnInfo

ldez:
The type of `ReturnInfo` doesn't seem "stable".

allinkl doesn't have a real API schema: https://kasapi.kasserver.com/soap/wsdl/KasApi.wsdl

```xml
<message name="KasApiAntwort">
<part name="Result" type="xsd:anyType"/>
</message>
```

```
2025/03/20 12:54:56 [WARN] [*.example.com] acme: cleaning up failed: allinkl: response struct decode: decoding failed due to the following error(s):

'Response.ReturnInfo' expected type '%!s(bool=false)', got unconvertible type '""', value: '""'
```

Related to #2489

### DIFF
--- a/providers/dns/allinkl/internal/client.go
+++ b/providers/dns/allinkl/internal/client.go
@@ -86,23 +86,23 @@ func (c *Client) AddDNSSettings(ctx context.Context, record DNSRequest) (string,
 }
 
 // DeleteDNSSettings Deleting a DNS Resource Record.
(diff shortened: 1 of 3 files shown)


dmke (approved):
> allinkl doesn't have a real API schema

Oh, the schema is real. It's just utterly useless.

ldez:
Yes it's a real decorative API schema :smile:

PR #2498, opened 2025-03-26
designate: speed up API requests by using filters

sebageek:
When searching for an object in OpenStack we can let the API do the heavy lifting by applying filters (i.e. ListOpts). This helps speeding up the API requests for projects that have a large number of zones or zones that have a large number of records.

---
In a larger test project this speeds up the API calls from about 30s to 1.5s for filtered vs unfiltered API calls.

### DIFF
--- a/providers/dns/designate/designate.go
+++ b/providers/dns/designate/designate.go
@@ -248,10 +248,12 @@ func (d *DNSProvider) getZoneID(wanted string) (string, error) {
 	listOpts := zones.ListOpts{
 		Name: wanted,
 	}
+
 	allPages, err := zones.List(d.client, listOpts).AllPages()
 	if err != nil {
 		return "", err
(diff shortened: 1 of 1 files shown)


sebageek:
Heyjo,

I don't think this creates site effects, as we're always filtering for the name of what we're looking for.

case 1: `getZoneID()` gets called with the zone name as `wanted`. Here we list all zones and then look for a zone that has the zone name we're searching for.

case 2: `getRecord()` gets called with the FQDN as `wanted`. Here we list all recordsets belonging to a given zone and filter for one with the wanted name + a type of TXT.

So, as we're always getting `wanted` as the name of the type of object we're looking for in the API we should be fine there.

Did I understand the question correctly? :)

ldez (approved):
LGTM

PR #2499, opened 2025-03-27
feat: add delay option for TLSALPN challenge

ldez:
Similar to #2422 but for TLSALPN challenge.

### DIFF
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -167,7 +167,7 @@ func CreateFlags(defaultPath string) []cli.Flag {
 		},
 		&cli.DurationFlag{
 			Name:  flgTLSDelay,
-			Usage: "Delay between the starts of the TLS listener (use for TLSALPN-01 based challenges) and the validation of the challenge.",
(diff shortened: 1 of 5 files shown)


cmd/flags.go
```diff
@@ -164,6 +165,11 @@ func CreateFlags(defaultPath string) []cli.Flag {
 			Usage: "Set the port and interface to use for TLS-ALPN-01 based challenges to listen on. Supported: interface:port or :port.",
 			Value: ":443",
 		},
+		&cli.DurationFlag{
+			Name:  flgTLSDelay,
+			Usage: "Delay between the starts of the TLS listener (use for TLSALPN-01 based challenges) and the validation of the challenge.",
```
dmke, line 170:
```suggestion
			Usage: "Delay between the start of the TLS listener (use for TLSALPN-01 based challenges) and the validation of the challenge.",
```

dmke (changes requested):
Found a typo :)

Otherwise LGTM

dmke (approved):

PR #2500, opened 2025-03-30
pdns: fix TXT record cleanup for wildcard domains

fries1234:
When using the pdns backend, using a domain and a wildcard domain at the same time (example.com, *.example.com) means cleanup wasn't successful because it was trying to DELETE all the TXT records even though its only supposed to delete 1 challenge at a time.


ldez (approved):
LGTM

PR #2501, opened 2025-04-03
chore: migrate to urfave v3

ldez:
The migration guide is sparse https://cli.urfave.org/migrate-v2-to-v3/, but the migration is not complex.


ldez:
There is problem with doc generator and the `dnshelp` command, I'm investigating

dmke (approved):
LGTM

dmke:
(Btw. I got a new graphics card, but Debian doesn't recognize it. I need distro hop a bit, before I have a working home PC again. For now, I'm limited to after-hours in the office).

ldez:
I have the feeling after playing a bit with urfave/cli v3 that not all things are stable, so we will wait a bit before the merge of this PR.

ldez:
v3.2 of urfave/cli breaks things... I always appreciated the simplicity of this lib, but the current release flow is bad, very bad.

dmke (approved):
I mean... if upgrading `urfave/cli` to v3 is this tedious, we can stay on v2 for a while longer.

Have you looked into alternatives, yet? Personally, I found `spf13/pflag` sufficient most of the time, and `spf13/cobra` helpful for anything beyond.

PR #2511, opened 2025-04-22
Control timestamp in log output using LEGO_LOG_TIMESTAMP

n12c:
Fixes https://github.com/go-acme/lego/issues/1895

### DIFF
--- a/log/logger.go
+++ b/log/logger.go
@@ -3,10 +3,23 @@ package log
 import (
 	"log"
 	"os"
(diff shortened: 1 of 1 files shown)


ldez:
Hello,

Issue #1895 is not something I wanted to implement for now: as you can see, I didn't flag this issue as an enhancement, and the issue has no traction.

Thank you, but I will decline this PR.

PR #2530, opened 2025-05-29
Increase the default response timeout from 30 to 45 seconds

fln:
This commit increases the amount of time we can wait for the ACME server response from 30 sec to 45 sec before retrying.

The reason for this change is the increasing amount of certificate issuance failures at the order finalization stage.

Example error:

```
2025/05/25 15:30:20 [INFO] [example.com] acme: Validations succeeded; requesting certificates
2025/05/25 15:30:51 Could not obtain certificates:
example.com: acme: error: 400 :: POST :: https://acme-v02.api.letsencrypt.org/acme/finalize/1234567890/123456789012 :: urn:ietf:params:acme:error:badNonce :: Unable to validate JWS :: JWS has an invalid anti-replay nonce: "AAAAAAAAAAAAAAAAAAAAAAAAAAAA-AAAAAAAAAAAAAAAAAAAAAA"
```

The error message reveals that the request was actually processed by the LE servers, but the `lego` client was impatient and retried the request after 30 seconds with the same JWS nonce.

The request timeout value from the Certbot project is used as a reference:

https://github.com/certbot/certbot/blob/v4.0.0/acme/acme/client.py#L33

The increased timeout value helps to successfully issue a certificate when LE servers are slow to respond.

### DIFF
--- a/lego/client_config.go
+++ b/lego/client_config.go
@@ -79,7 +79,7 @@ func createDefaultHTTPClient() *http.Client {
 				KeepAlive: 30 * time.Second,
 			}).DialContext,
 			TLSHandshakeTimeout:   30 * time.Second,
-			ResponseHeaderTimeout: 30 * time.Second,
+			ResponseHeaderTimeout: 45 * time.Second,
 			TLSClientConfig: &tls.Config{
 				ServerName: os.Getenv(caServerNameEnvVar),
 				RootCAs:    initCertPool(),


fln:
The root cause for the `JWS has an invalid anti-replay nonce` error is actually more intricate. The `lego` client has the code to retry with a new good nonce, but this code is almost dead due to misconfiguration:

https://github.com/go-acme/lego/blob/master/acme/api/api.go#L78

```go
func (a *Core) retrievablePost(uri string, content []byte, response interface{}) (*http.Response, error) {
	// during tests, allow to support ~90% of bad nonce with a minimum of attempts.
	bo := backoff.NewExponentialBackOff()
	bo.InitialInterval = 200 * time.Millisecond
	bo.MaxInterval = 5 * time.Second
	bo.MaxElapsedTime = 20 * time.Second    # <---- HERE IS THE PROBLEM

	var resp *http.Response
	operation := func() error {
		var err error
		resp, err = a.signedPost(uri, content, response)
		if err != nil {
			// Retry if the nonce was invalidated
			var e *acme.NonceError
			if errors.As(err, &e) {
				return err
			}

			return backoff.Permanent(err)
		}

		return nil
	}
	...
```

The HTTP client used for requests uses the `github.com/hashicorp/go-retryablehttp` package. It silently performs a retry with the same request (and same JWS nonce) in case of network error. The request header timeout is treated as a network error. So when LE servers are slow to respond, but actually do process the request, the `lego` client retries after 30 seconds with the same nonce.

The second request is immediately rejected by LE servers because of JWS nonce reuse. This error is returned to the `retrievablePost()` which has the logic to retry with a new (good) nonce. It uses the `github.com/cenkalti/backoff/v4` package for request retries. However, the retry logic is configured to give up after 20 seconds.

The main request timeout is 30 seconds. In this scenario the retry logic in `retrievablePost()` never actually gets used. We know 30 seconds have already passed until we get an error here, and it is already past the allowed retry time.

So the root problem is misalignment between default HTTP client configuration and `github.com/cenkalti/backoff/v4` configuration.

However, the errors we see on production servers hint to us that LE is actually processing the request. So instead of retrying with the same nonce, then getting an error and then retrying with a good nonce, it would be better to just be a little bit more patient and wait a bit more for LE to respond.

ldez:
I need to think about that.

ldez:
Instead of changing the timeout, I changed the way to handle the error: now the retryable client doesn't retry on badnonce error.

PR #2564, opened 2025-07-07
azion: improve zone lookup

marcusgrando:
also add tests.

### DIFF
--- a/providers/dns/azion/azion.go
+++ b/providers/dns/azion/azion.go
@@ -12,6 +12,7 @@ import (
 	"github.com/aziontech/azionapi-go-sdk/idns"
 	"github.com/go-acme/lego/v4/challenge/dns01"
 	"github.com/go-acme/lego/v4/platform/config/env"
+	"github.com/miekg/dns"
 )
 
 // Environment variables names.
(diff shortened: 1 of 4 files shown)


marcusgrando:
@ldez I would appreciate it if you could look at this. 😉

ldez:
@marcusgrando I have some doubts, based on your API documentation:
https://api.azion.com/#ba265bc9-82a9-4ec9-b5ca-2199a1ea0f40

<details>

```json
{
  "count": 1,
  "links": {
    "previous": null,
    "next": null
  },
  "total_pages": 1,
  "results": [
    {
      "domain": "youdomain.com",
      "is_active": true,
      "name": "New Hosted Zone via API POST method",
      "id": 1234
    }
  ],
  "schema_version": 3
}
```

</details>

The field `name` is a random string, and the field `domain` is the domain.

Can you confirm that?

marcusgrando:
Yes, you are right. A name is just any given name that describes the domain, and the domain is exactly the FQDN.

ldez:
ok so the existing implementation was wrong. I fixed that.

ldez (approved):
LGTM

PR #2566, opened 2025-07-07
chore: update to go1.24

ldez:
Go1.25 will happen in less than 2 months, so I think it's safe to update to go1.24.

https://go.dev/wiki/Go-Release-Cycle

- Updates the minimum Go version
- Uses go1.24 new features to simplify the code

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -6,6 +6,11 @@ formatters:
     - gofmt
     - gofumpt
     - goimports
+  settings:
+    gofmt:
+      rewrite-rules:
+        - pattern: 'interface{}'
+          replacement: 'any'
 
 linters:
   default: all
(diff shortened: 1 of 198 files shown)


dmke:
That's a large diff, I'll need some time to work through it (although most looks like a simple `interface{}` → `any` transformation).

ldez:
I created isolated commits, one change per commit, so it's easier to review each one.

You can skip the commits:
- [chore: replace interface{} with any](https://github.com/go-acme/lego/pull/2566/commits/e0a34a4b8e1ad7ddef8069ac38b76505a9ffc6ed)
- [chore: simplify function parameters](https://github.com/go-acme/lego/pull/2566/commits/e1f756c042961ad11d48767d932c0a8e57a465aa)

because this is only the result of golangci-lint autofix.

PR #1868, opened 2023-03-11
chore: refactor clients

ldez:
TLDR: rewrite all the internal clients to be more uniform, have better errors, and improve debugging. 

---

My journey on this PR started in December (yes December) when I was reviewing a PR about a new DNS provider.
I was thinking that the tests are really not homogeneous, and they are painful to write.
Then I started to think about something to simplify the way to write tests.
So I started but at this point, I was struck by the non-uniformity of the clients themselves.

I started to think about how to have something more uniform without using or creating a framework because each provider has its own small variation.
I rewrite all clients multiple times, and I tried multiple approaches for authentication, error management, etc.
I rewrote each client "layer" by "layer" to try to have something more uniform, then I increase the "duplication" on purpose.
I think it's easier to contribute if the "shape" of the clients is homogeneous even if they contain small variations.

Also, during this review/rewrite, I saw that the errors are never the same and didn't provide the expected information.
Then I created some shared error types.

In the end, I didn't rewrite the tests in this PR, because the PR is already too large, so the tests will be rewritten in another PR, but I added missing tests to help during the refactor.

This PR is large, and I spent a lot of time on it, so @dmke, I hope the review of the PR will be less complex than the analysis of all API clients.

I hope this PR will increase the maintainability, and facilitate the creation and review of new DNS providers. 

Now I will be available for reviewing the other PRs (even if I already prepared 3 new branches :smile:)

PR #2571, opened 2025-07-08
test: server client mock

ldez:
This PR contains what was my first goal when creating #1868.

For the history, I started to work on this idea in 2022.
I created multiple branches during this period on the topic.

This implementation uses multiple "new" Go elements (generics, `slices` package, new router syntax, ...); those elements were added after 2022, without them, this type of mock system cannot exist without important complexity.

I created a naive, heavily opinionated, but robust mock server/router, dedicated to tests, and only for lego.

---

I know this PR is massive.

The first commit is the mock system; other commits are updates of the tests.

You don't need to review all the tests and all the commits.
You can review the first commit and a few tests.

### DIFF
--- a/providers/dns/acmedns/internal/http_storage_test.go
+++ b/providers/dns/acmedns/internal/http_storage_test.go
@@ -5,14 +5,14 @@ import (
 	"net/http/httptest"
 	"testing"
 
-	"github.com/go-acme/lego/v4/platform/tester/stubrouter"
+	"github.com/go-acme/lego/v4/platform/tester/servermock"
 	"github.com/nrdcg/goacmedns"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
-func mockBuilder() *stubrouter.Builder[*HTTPStorage] {
-	return stubrouter.NewBuilder[*HTTPStorage](
+func mockBuilder() *servermock.Builder[*HTTPStorage] {
+	return servermock.NewBuilder[*HTTPStorage](
 		func(server *httptest.Server) (*HTTPStorage, error) {
 			storage, err := NewHTTPStorage(server.URL)
 			if err != nil {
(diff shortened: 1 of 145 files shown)


dmke:
Neat! Give me a few days to wrap my head around the details :)

providers/dns/acmedns/internal/http_storage_test.go
```diff
@@ -1,57 +1,35 @@
...
-
-	storage, err := NewHTTPStorage(server.URL)
-	require.NoError(t, err)
-
-	storage.client = server.Client()
-
-	return storage
+			return storage, nil
+		},
+		clientmock.CheckHeader().WithJSONHeaders())
```
dmke, line 26:
```suggestion
		clientmock.CheckHeader().WithJSONHeaders(),
	)
```

dmke (approved):
Fantastic job!

The name `clientmock` however implies you're mocking (i.e. faking or pretenting to be) a client, while in fact you're mocking the client's endpoint (i.e. server). Hence, something like `servermock` or `stubserver` would be more appropriate.

I *can* arrange myself with the current name though, so I'll leave it up to you whether you want to change it or not.

ldez:
At the beginning, I called it `servermock`, but as the goal of the `Build` method is mainly here to build a client that uses a mocked server, I renamed it to `clientmock`, the naming is not easy.

The `Build` method is also used to build other elements, such as `DNSProvider`, making the naming more complex.

ldez:
Maybe `stubrouter` :thinking:

ldez:
I renamed it to `stubrouter`, but we can change that later. I don't consider this package as a part of the lego API.

dmke:
Hm... `stubrouter` is something I'd use to test `gorilla/mux`.

Maybe something in the direction of *test double builder*, like `endpointdouble` or `acmedouble`?

ldez:
The implementation is a router, as it creates routes and a server, uses chains of handlers.

It's not related to `acme`, and the word `double` seems not relevant either.

:thinking:

dmke:
Again, I can arrange myself with `clientmock`. I just thought to bring it up ;-)

ldez:
I don't have a better name for now, but as I said, we can change that later.

ldez:
:thinking: I will rename it to `servermock`, it seems the more accurate name.

PR #2580, opened 2025-07-11
vinyldns: add an option to add quotes around the TXT record value

eusouovh:
**Description**
This PR adds quotes to DNS records to ensure Route53 compatibility during the DNS-01 ACME/Let's Encrypt validation process.

**Problem**
The current code didn't include quotes in DNS record values, causing DNS-01 challenge failures when used with Route53. Route53 has specific formatting requirements that require certain values to be quoted.


ldez:
Use the Route53 provider instead of the Vinyl one.

Maybe I'm missing something?

eusouovh:
Hi, @ldez !
I use VinylDNS to manage Route53. With this setup, I cannot go directly to Route53 - all DNS changes must be made through the VinylDNS provider.

ldez:
I found nothing in the API documentation about quotes around the TXT value.
- https://www.vinyldns.io/api/recordset-model.html
- https://www.vinyldns.io/api/recordset-model.html#record-data

I have no proof that it will not break the other users of Vinyl DNS.

ldez:
I added an option to add quotes (`VINYLDNS_QUOTE_VALUE=true`).

The default behavior is unchanged, so no possible regressions.

eusouovh:
Perfect, my friend. :)
I really appreciate your help.

ldez (approved):
I recommend opening an issue to explain your problem before opening this type of PR.

PR #2586, opened 2025-07-15
tests: improve function naming

ldez:
Use the same naming pattern with handlers that read files:
- `XXXFromFile`: reads a file from a path
- `XXXFromFixture`: reads a file from a `./fixtures` directory
- `XXXFromInternal`: reads a file from a `./internal/fixtures` directory

### DIFF
--- a/platform/tester/servermock/handler_file.go
+++ b/platform/tester/servermock/handler_file.go
@@ -15,6 +15,7 @@ type ResponseFromFileHandler struct {
 	filename   string
 }
 
+// ResponseFromFile creates a [ResponseFromFileHandler] using a filename.
 func ResponseFromFile(filename string) *ResponseFromFileHandler {
 	return &ResponseFromFileHandler{
 		statusCode: http.StatusOK,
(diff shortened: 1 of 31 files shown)


dmke (approved):
LGTM.

Just a curious question: Is there a particular reason to use a `fixtures` directory over the conventional `testdata` directory? The docs explicitly state only the latter:

> The go tool will ignore a directory named "testdata", making it available to hold ancillary data needed by the tests.
>
> https://pkg.go.dev/cmd/go#hdr-Test_packages

(see also: https://dave.cheney.net/2016/05/10/test-fixtures-in-go)

The context here is `go test ./...`, so *if* there were any `*.go` files beneath a `testdata` directory, they would never be compiled/executed in a test run. For other fixture files this wouldn't matter. I'm not entirely sure whether `testdata` and/or `fixtures` are excluded from the module cache as well (or whether that includes only `.go` files + a few selected like `README.md`).

ldez:
`testdata` and `fixtures` are not excluded from the module zip because test files are not excluded.

The Go Team's idea behind this is that the tests of a module can be run.

In the context of lego, there is no difference between `testdata` and `fixtures`.

That said, the naming is mainly related to historical reasons.

Now that the test code is homogeneous, the usage of explicit paths to `fixtures` is limited, so maybe I can rename the directories without spending too much time on this.

dmke:
> so maybe I can rename the directories without spending too much time on this

Don't bother, your justification is fine as is :)

ldez:
Just for fun and to share knowledge: directories that contain a `go.mod` are excluded from the module zip (because monorepo with multiple modules).

In Go 1.25 (not released yet, but soon, in August), there is a new directive inside the `go.mod` called `ignore` where we will be able to define paths we want to exclude from compiler scanning (same behavior as `testdata` directories).

> The new go.mod ignore [directive](https://tip.golang.org/ref/mod#go-mod-file-ignore) can be used to specify directories the go command should ignore. Files in these directories and their subdirectories will be ignored by the go command when matching package patterns, such as all or ./..., but will still be included in module zip files.
> https://tip.golang.org/doc/go1.25#go-command


This is interesting for mixed content projects (ex: Go + Node.js), because the Go compiler browses all directories (to find Go files) even if there are no Go files in them.
This browsing, in the context of Node.js, was very costly because "`npm` downloads the world" :smile_cat:

Issue #1613, opened 2022-03-30
OCI Instance Principal Based Auth Implementation


JanisOrlovs:
### Welcome

- [X] Yes, I've searched similar issues on GitHub and didn't find any.

### How do you use lego?

Library

### Detailed Description

Hello,

ACME OCI provider only provides the option to use service keys, in order to use LEGO. However, if you run within an instance of OCI there is a possibility to authenticate with the instance principal. https://docs.oracle.com/en-us/iaas/Content/Identity/Tasks/callingservicesfrominstances.htm

Currently, ACME LEGO don't support it. Please add support for OCI instance principal auth.

cleitinif:
I did this change on #2599, but I couldn't create tests for it, as OCI SDK expects to run this auth mode just on OCI Compute Instances.

PR #2599, opened 2025-07-22
oraclecloud: handle instance_principal authentication

cleitinif:
This pull request enhances the Oracle Cloud DNS provider by introducing support for InstancePrincipal authentication type and update required environment variables handling.

Fixes https://github.com/go-acme/lego/issues/1613


ldez:
Why are you opening several PRs?

Why is the PR #2598 closed?

cleitinif:
> Why are you opening several PRs?
> 
> Why is the PR #2598 closed?

It's my fault. I accidentally did a wrong merge commit and deleted that branch

ldez:
The branch is not deleted; you have to recreate it with the same name, and reopen the PR.

Well, it doesn't matter now because it's done.

ldez:
I have several problems with this PR:
1. The change is breaking because the env vars are not checked anymore.
2. Some changes make no sense.
3. Based on the implementation inside the oci-go-sdk of `instancePrincipalConfigurationProvider`, this should not be done like that.

cleitinif:
What do you mean by this should not be done like that? I checked the `InstancePrincipalConfigurationProvider` but I couldn't understand

Also, about the env vars, do you have some suggestions where to validate them? Because each authentication method requires different variables

ldez:
- Wrong usage of git and GitHub
- The tests are not run
- Usage of AI
- ...

This is not the right way to interact with me.

---

I updated your PR, so no need for AI slop.

Patience, patience, patience.

PR #2602, opened 2025-07-25
chore: clean up tests and code

ldez:
- Replaces test domains with `.example.` domains
- Uses the servermock system outside DNS providers

When I was updating tests with the new servermock system, I found a weird element:
```go
	// Get issuerCert from bundled response from Let's Encrypt
	// See https://community.letsencrypt.org/t/acme-v2-no-up-link-in-response/64962
	_, issuer := pem.Decode(cert)
	if issuer != nil {
```

The `if` is always `true` because `issuer` is a `[]byte`.

I wanted to fix that, so I read the reference inside the comment: https://community.letsencrypt.org/t/acme-v2-no-up-link-in-response/64962

And in ACME v2, there is no `up` link for the issuer.
- https://www.rfc-editor.org/rfc/rfc8555.html#section-7.4.2

### DIFF
--- a/acme/api/certificate.go
+++ b/acme/api/certificate.go
@@ -96,6 +96,8 @@ func (c *CertificateService) getCertificateChain(cert []byte, headers http.Heade
 		return &acme.RawCertificate{Cert: cert, Issuer: issuer}
 	}
 
+	// TODO(ldez) remove all this section.
+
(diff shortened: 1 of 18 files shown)


dmke (approved):
LGTM

PR #2610, opened 2025-08-01
chore: clean up

ldez:
- Removes useless call to `DumpRequest`
- Replaces `dns01.ToFqdn` with `dns.Fqdn`

### DIFF
--- a/challenge/dns01/fqdn.go
+++ b/challenge/dns01/fqdn.go
@@ -6,6 +6,11 @@ import (
 	"github.com/miekg/dns"
 )
 
+// ToFqdn converts the name into a fqdn appending a trailing dot.
+//
+// Deprecated: Use [github.com/miekg/dns.Fqdn] directly.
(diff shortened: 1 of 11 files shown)


challenge/dns01/fqdn.go
```diff
@@ -6,15 +6,6 @@ import (
 	"github.com/miekg/dns"
 )
 
-// ToFqdn converts the name into a fqdn appending a trailing dot.
-func ToFqdn(name string) string {
```
dmke, line 10: Removing an export is technically a breaking change.

Better leave it in place and deprecate it:

```go
// ToFqdn converts the name into a fqdn appending a trailing dot.
//
// Deprecated: Use [github.com/miekg/dns.Fqdn] directly.
var ToFqdn = dns.Fqdn
```

dmke (changes requested):

dmke (approved):

PR #2613, opened 2025-08-05
tests: new DNS router/server/mock

ldez:
I created a naive router for DNS server.
The goal is only to fit the lego needs.

I rewrote the rfc2136 tests with the new system.
I drop tests related to the previous fake DNS server.
I renamed and rewrote the tests to make them more relevant.

I added tests on the `dns01` package (the DNS01 challenge core).
Tests based on external things have been replaced with "local" elements.
Sadly, I was forced to add a non-exposed global variable (`defaultNameserverPort`) only for the tests inside the production code.
It's a string, non-exposed, so it will not have an impact on library users, but it's a bit sad.

This PR is a type of first step because it's possible to use this mock system inside all the DNS provider tests that call `dns01.FindZoneByFqdn`.
Thinking about that pushes my recurring rabbit hole, DNS calls management inside DNS providers, and lego.
I have the persistent feeling that the current implementation, based on functions, is not right.
But this is another topic.

Anyway, I'm happy to finally be able to test DNS behaviors without having to set up a domain and a DNS server or playing with my `/etc/hosts`.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -195,8 +195,8 @@ linters:
         text: dnsTimeout is a global variable
         linters:
           - gochecknoglobals
-      - path: challenge/dns01/nameserver_test.go
-        text: findXByFqdnTestCases is a global variable
+      - path: challenge/dns01/precheck.go
+        text: defaultNameserverPort is a global variable
         linters:
           - gochecknoglobals
       - path: challenge/http01/domain_matcher.go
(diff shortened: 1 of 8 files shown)


dmke:
Will have a look later (maybe tomorrow). I'm quite busy at the moment.

dmke (approved):

PR #2632, opened 2025-08-28
chore: update documentation theme

ldez:
I also renamed the Make targets and cleaned the configuration.


ldez:
Good catch, I didn't spot it.

ldez:
In fact, it was already like that in the current doc.

dmke (approved):

PR #2655, opened 2025-09-23
otc: add example

bartjanssens92:
I had issues to get the Open Telekom Cloud provider to work, my main problem was the value of the OTC_IDENTITY_ENDPOINT.
As there was no example yet I added it with the correct value of that parameter.


bartjanssens92:
Ah good to know, we faced an issue with the value as our main site is eu-nl causing all kinds of these issues. Thank you for clarifying that!

PR #2669, opened 2025-10-15
nifcloud: change the timeout and polling interval to variables

kakakikikeke-fork:
I fixed nifcloud provider. Various values ​​of waif.For were fixed, so they are now variable.


ldez:
The variables `d.config.PropagationTimeout` and `d.config.PollingInterval` are not related to this section of the code.
They are related to the method `DNSProvider.Timeout()`.

The values `d.config.PropagationTimeout` (60s) and `d.config.PollingInterval` (2s) are not the same as the existing values, respectively 120s and 4s.

This difference is expected.

Unless you can explain, in an issue, what problem you are trying to solve, this PR can only be evaluated as a regression.

Issue #2679, opened 2025-10-21
HTTP error 429 Too Many Requests not handled


rds13:
### Welcome

- [x] Yes, I'm using a binary release within the two latest releases.
- [x] Yes, I've searched for similar issues on GitHub and didn't find any.
- [x] Yes, I've included all information below (version, config, etc).

### What did you expect to see?

I am using lego to renew a certificate with dns_01 protocol with Let's Encrypt.
I had to built a custom version of lego 4.26 to fetch keylogger information and to enforce TLS 1.2 to be able to analyze the traffic between my server and Let's Encrypt.
My understanding is that on reception of error code 429 lego should stop sending requests and display the detail of the error code received.

### What did you see instead?

The workflow observed is:

1. Request:  GET /directory HTTP/1.1
2. Response: HTTP/1.1 200 OK
3. Request: GET /acme/renewal-info/XXXXXXX HTTP/1.1
4. Response: HTTP/1.1 200 OK
5. Request: HEAD /acme/new-nonce HTTP/1.1
6. Response: HTTP/1.1 200 OK
7. Request: POST /acme/new-order HTTP/1.1
8. Response: HTTP/1.1 429 Too Many Requests
Detail : Your account is temporarily prevented from requesting certificates for poly.tique.info and possibly others. Please visit: https://portal.letsencrypt.org/sfe/v1/unpause?jwt=XXXXXXX
10. Request: POST /acme/new-order HTTP/1.1
11. Reponse: HTTP/1.1 400 Bad Request
Detail: Unable to validate JWS :: JWS has an invalid anti-replay nonce: \"9FCfmSBhbP-MUFNbwvsH_UWhNBcz1kZXGcMMNmkP0MKaiSfYw28\"
13. Request: POST /acme/new-order HTTP/1.1
14. Response: HTTP/1.1 429 Too Many RequestsDetail : Your account is temporarily prevented from requesting certificates for poly.tique.info and possibly others. Please visit: https://portal.letsencrypt.org/sfe/v1/unpause?jwt=XXXXXXX
14. Request: POST /acme/new-order HTTP/1.1
15. Reponse: HTTP/1.1 400 Bad Request
Detail: Unable to validate JWS :: JWS has an invalid anti-replay nonce: \"9FCfmSBhbP-MUFNbwvsH_UWhNBcz1kZXGcMMNmkP0MKaiSfYw28\"
....

### How do you use lego?

Binary

### Reproduction steps

[...]

PR #2682, opened 2025-10-22
chore: improve retryable HTTP client error handling

ldez:
Parses body as `acme.ProblemDetails` if possible.

Related to #2679


rds13:
👍

PR #2683, opened 2025-10-22
hetzner: add deprecation logs

ldez:
I forgot to add deprecation logs.


ldez:
I prefer it without a helper; the duplication in this context is not a problem for me.

PR #2692, opened 2025-10-28
chore: add debug transport on DNS API clients

ldez:
I added a transport that allows dumping requests and responses on API clients.

This is managed with `LEGO_DEBUG_DNS_API_HTTP_CLIENT` env var.

To do that, I fixed or updated the way to set an HTTP client on API clients.

One commit, one step.

Note, some DNS providers are not supported (21/165):
- `alidns`
- `baiducloud`
- `designate`
- `edgedns`
- `edgeone`
- `huaweicloud`
- `ibmcloud`
- `iij`
- `iijdpf`
- `infoblox`
- `inwx`
- `lightsail`
- `liquidweb`
- `metaname`
- `route53`
- `tencentcloud`
- `ultradns`
- `vegadns`
- `vkcloud`
- `volcengine`
- `yandexcloud`

### DIFF
--- a/docs/content/usage/cli/Options.md
+++ b/docs/content/usage/cli/Options.md
@@ -145,13 +145,14 @@ LEGO_DEBUG_CLIENT_VERBOSE_ERROR=true
 
 ### LEGO_DEBUG_DNS_API_HTTP_CLIENT
 
-**⚠️ WARNING: This will expose credentials in the log output! ⚠️**
+> **⚠️ WARNING: This will expose credentials in the log output! ⚠️**
+> 
+> Do not run this in production environments, or if you can't be sure that logs aren't accessed by third parties or tools (like log collectors).
+> 
+> You have been warned. Here be dragons.
 
-Do not run this in production environments, or if you can't be sure that logs aren't accessed by third parties or tools (like log collectors).
-
-You have been warned. Here be dragons.
-
-The environment variable `LEGO_DEBUG_DNS_API_HTTP_CLIENT` allows debugging the DNS API interaction. It will dump the full request and response to the log output.
+The environment variable `LEGO_DEBUG_DNS_API_HTTP_CLIENT` allows debugging the DNS API interaction.
+It will dump the full request and response to the log output.
 
 Some DNS providers don't support this option.
 


providers/dns/internal/clientdebug/client.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	return &DumpTransport{rt: rt}
+}
+
+func (d *DumpTransport) RoundTrip(h *http.Request) (*http.Response, error) {
+	dump, _ := httputil.DumpRequestOut(h, true)
+
+	println("[HTTP Request]")
+	println(string(dump))
```
dmke: Avoid `print` and `println`. From the spec:

> These functions are documented for completeness but are not guaranteed to stay in the language.
> 
> https://go.dev/ref/spec#Bootstrapping

```suggestion
	fmt.Println("[HTTP Request]")
	fmt.Println(string(dump))
```

providers/dns/internal/clientdebug/client.go
```diff
@@ -0,0 +1,52 @@
...
+
+	resp, err := d.rt.RoundTrip(h)
+
+	dump, _ = httputil.DumpResponse(resp, true)
+
+	println("[HTTP Response]")
+	println(string(dump))
```
dmke:
```suggestion
	fmt.Println("[HTTP Response]")
	fmt.Println(string(dump))
```

ldez:
These are good questions.

First, this is for debugging only, and I created this behavior mainly for lego CLI debugging, so it should not be used outside this scope.

The time limit doesn't seem to be secure enough (and handling the timezone will be a pain).
And as you said, the TTY will not work.

I've thought about other solutions, but nothing seems secure enough to me.

Redacting the data is not possible because credentials can be inside the path, query, body, or headers, depending on the API. And the credentials can result from a unique and "non-deterministic" transformation

I can see only one really secure solution: using a build tag.
By default, lego will be built without this tag, and the library will also ignore this element.
I think we can keep the env var as an extra security level.

This will limit the debug possibilities, but I don't see other ways to secure the system.

There is another possibility: assuming the insecure behavior, and adding a warning inside the documentation about the impact of using this env var.

ldez:
~~:thinking: maybe I have an idea to redact the information, I need to think about it.~~

EDIT: My idea doesn't work.

But I'm waiting for your feedback about my previous ideas.

dmke:
> Redacting the data is not possible because credentials can be inside the path, query, body, or headers, depending on the API. And the credentials can result from a unique and "non-deterministic" transformation.

Indeed (there are also deterministic transformations, like Base64 encoding, e.g. for the Basic Auth header).

> I can see only one really secure solution: using a build tag.

Hm... The user then may forget to replace the debug-build with the production-build, which only shifts the problem.

> There is another possibility: assuming the insecure behavior, and adding a warning inside the documentation about the impact of using this env var.

I like this. Add a big, bold, strongly worded warning to the documentation.

I've had another idea: Let the env var point to an existing file with 0600 permissions, and write the dump to that file (and refuse to dump if the file does not exists or has the wrong permissions). This won't work on Windows, though (but maybe there are some tricks to achieve something similar).

ldez:
> I've had another idea: Let the env var point to an existing file with 0600 permissions

lego CLI is also provided as a Docker image.
It will be a pain with containers.

I'm unsure whether writing inside a file is better or worse, as this is an independent file, so the user may forget to remove it.

> This won't work on Windows, though (but maybe there are some tricks to achieve something similar).

It will be a pain to create and maintain.

ldez:
I found a solution to redact more than 90% of the credentials.

dmke:
> I found a solution to redact more than 90% of the credentials.
>
> → 7cee538d029f56f2bd4f709bb60359c919372b7b, 516a495b30f4a0e8301cd932cf816e106d6496be

That looks interesting, but honestly, I'm not convinced that this is the right approach either. You still need to update every `Wrap` call and investigate which set of options are missing (`WithEnvKeys`, `WithValues`, `WithHeaders`).

This also introduces a maintenance burden because this to be revisited on nearly any dependency upgrade (I'd wish to say "only on major upgrades", but, you know... \*wildly gestures at environment\*).

Additionally, this creates a false promise of "even if you forget to unset the debug env var in production, your secrets are safe". A block-list just can't keep that promise.

Hence, I'm still in favor of keeping it as-is (or, as-was yesterday), and extend the documentation:

> ### LEGO_DEBUG_DNS_API_HTTP_CLIENT
> 
>> ⚠️ **WARNING: This will expose credentials in the log output!** ⚠️
>>
>> Do not run this in production environments or if you can't be sure that logs aren't accessed by third parties or tools (like log collectors).
>>
>> You have been warned. Here be dragons.
> 
> The environment variable `LEGO_DEBUG_DNS_API_HTTP_CLIENT` allows debugging the DNS API interaction. It will dump the full request and response to the log output.
> 
> Some DNS providers don't support this option.
> 
> Example:
> 
> ```bash
> LEGO_DEBUG_DNS_API_HTTP_CLIENT=true lego ...
> ```

ldez:
The redaction is not a promise of security but only a reduction of the risk.

I updated the documentation with your suggestion.

I want to keep the basic redaction of the headers (`Authorization`, `Token`, ...), as I said, this is only a risk reduction approach, not a secure approach.

I will not use `WithEnvKeys`, `WithValues`, `WithHeaders`, but I want to keep it to help me when I'm debugging a new provider with a user.

dmke:
> I want to keep the basic redaction of the headers (Authorization, Token, ...), as I said, this is only a risk reduction approach, not a secure approach.

Fair enough.

Issue #2628, opened 2025-08-24
Gandi API URL appears to be outdated


shaunblyde:
### Welcome

- [x] Yes, I'm using a binary release within 2 latest releases.
- [x] Yes, I've searched similar issues on GitHub and didn't find any.
- [x] Yes, I've included all information below (version, config, etc).

### What did you expect to see?

I expected to see the ACME DNS challenge work successfully using the URL provided by Gandi Documentation for gandiv5: 
https://api.gandi.net/v5/livedns 

### What did you see instead?

Lego appears to be using an outdated gandiv5 API URL of https://dns.api.gandi.net/api/v5/
This gives an error (in logs below)

### How do you use lego?

Through Traefik

### Reproduction steps

Configure Traefik for provider gandiv5, including providing the PAT environment variable, and using the staging server.
Restart Traefik
Check logs.

### Effective version of lego

```console
This command gives "lego: not found" when executed within the Traefik container. Traefik version is 3.5.0
```

### Logs

<details>

```console
ERR Unable to obtain ACME certificate for domains error="unable to generate a certificate for the domains [example.com .example.com]:
error: one or more domains had a problem:
[.example.com] [.example.com] acme: error presenting token: unable to get TXT records for domain example.com and name _acme-challenge:
unable to communicate with the API server:
error: Get "``https://dns.api.gandi.net/api/v5/domains/``example.com``/records/_acme-challenge/TXT\``":
net/http: invalid header field value for "Authorization"
[example.com] [example.com] acme: error presenting token: unable to get TXT records for domain example.com and name _acme-challenge:
unable to communicate with the API server: error: Get "``https://dns.api.gandi.net/api/v5/domains/``example.com``/records/_acme-challenge/TXT\``":
net/http: invalid header field value for "Authorization"
" 
ACME CA=https://acme-staging-v02.api.letsencrypt.org/directory acmeCA=https://acme-staging-v02.api.letsencrypt.org/directory domains=["example.com",".example.com"] providerName=lets-encr.acme routerName=websecure-test@file rule=Host(test.example.com)`
```

</details>

### Go environment (if applicable)

<details>

```console
$ go version && go env
# paste output here
```

</details>

ldez:
Hello,

The API endpoint https://dns.api.gandi.net/api/v5/ used by lego still works, your problem is not related to that.

The error inside the log is `invalid header field value for "Authorization"`

Your configuration is not valid: your PAT is malformed.

PR #2708, opened 2025-11-08
gandiv5: update base API URL

oliverbr:
This change sets the Gandi LiveDNS API URL to the one [listed in the official API docs](https://api.gandi.net/docs/livedns/).
Changes to the API URL were already mentioned in #2628 (but not causing any issues, yet).

With the old API URL (https://dns.api.gandi.net/api/v5), I am currently getting certificate expiry errors when trying to request new certificates:

```log
2025/11/07 22:30:00 Could not obtain certificates:
[REDACTED] [REDACTED] acme: error presenting token: unable to get TXT records for domain REDACTED and name REDACTED: unable to communicate with the API server: error: Get "https://dns.api.gandi.net/api/v5/domains/REDACTED/records/REDACTED/TXT": tls: failed to verify certificate: x509: certificate has expired or is not yet valid: current time 2025-11-07T22:30:00+01:00 is after 2025-11-05T19:07:24Z
[REDACTED] [REDACTED] acme: error presenting token: unable to get TXT records for domain REDACTED and name REDACTED: unable to communicate with the API server: error: Get "https://dns.api.gandi.net/api/v5/domains/REDACTED/records/REDACTED/TXT": tls: failed to verify certificate: x509: certificate has expired or is not yet valid: current time 2025-11-07T22:30:05+01:00 is after 2025-11-05T19:07:24Z
```

Since the old URL isn't mentioned in the official docs and the certificate used for that domain has now expired, I think switching to the new one is the way to go. I can't find any sources for this, but it feels like a deprecation of the old URL.

I have already built a local copy of `lego` with these changes and can confirm that the change introduced by this PR does indeed request a new certificate successfully and no longer produces expiry errors.

Disclaimer: I skimmed the code to see if this change would have any unintended effects, and couldn't find any. I'm neither an expert with lego or Go, however, and may have missed something.

### DIFF
--- a/providers/dns/gandiv5/internal/client.go
+++ b/providers/dns/gandiv5/internal/client.go
@@ -15,7 +15,7 @@ import (
 )
 
 // defaultBaseURL endpoint is the Gandi API endpoint used by Present and CleanUp.
-const defaultBaseURL = "https://dns.api.gandi.net/api/v5"
+const defaultBaseURL = "https://api.gandi.net/v5/livedns"
 
 // APIKeyHeader API key header.
 const APIKeyHeader = "X-Api-Key"


ldez:
The certificate for `https://dns.api.gandi.net` has not expired.

```
CN: dns.api.gandi.net
SANs: dns.api.gandi.net, www.dns.api.gandi.net
Not Before: Fri, 12 Sep 2025 00:00:00 GMT
Not After: Sun, 11 Oct 2026 23:59:59 GMT 
```

http://www.sslchecker.com/sslchecker?su=e0ab9087f6bad502e6904dc3cfe1b65f

oliverbr:
How strange. I definitely did not make those errors up - maybe that was a load balancing issue on Gandi's side, then.

Thanks for accepting the PR, anyhow!

PR #2713, opened 2025-11-11
Add http proxy support for namecheap provider

marcschwaiger:
I created this PR because Namecheap API needs whitelisted IPv4 (IPv4 only) and on my traefik host i don't have a static IPv4..


ldez:
This is not the right approach, could you open an issue?

PR #2715, opened 2025-11-13
namecheap: add experimental proxy support

### LINKED ISSUE #2714: Add http proxy support for namecheap provider
### Welcome

### How do you use lego?

Through Traefik

### Effective version of lego

whatever latest Traefik uses

### Detailed Description

Namecheap API needs whitelisted IPv4 (IPv4 only) and on my traefik host i don't have a static IPv4.. (also it is not reachable from internet)
So I currently rely on an acme DNS server on another host that does not seem actively developed, so it does not satisfy my needs for security of a public facing app. 
My idea was to be able to specify a http proxy for Namecheap DNS challenge and I created a PR #2713 but it got closed.

ldez:
**This is experimental, and it will not be documented.**

I will wait for issues to be reported about the same topic before documenting it and applying it to the rest of the providers (when it's possible).

This PR adds support for the following environment variables:

- `NAMECHEAP_HTTP_PROXY`
- `NAMECHEAP_HTTPS_PROXY`
- `NAMECHEAP_NO_PROXY`
- `NAMECHEAP_REQUEST_METHOD`

The behavior is the same as the standard library ([`http.ProxyFromEnvironment`](https://pkg.go.dev/net/http#ProxyFromEnvironment)), but the environment variables are prefixed with `NAMECHEAP_`.

It allows using different proxy settings for the DNS API client than the ones used for the ACME server client.

Fixes #2714

### DIFF
--- a/providers/dns/namecheap/transport.go
+++ b/providers/dns/namecheap/transport.go
@@ -10,6 +10,16 @@ import (
 	"golang.org/x/net/http/httpproxy"
 )
 
+const (
+	envHTTPProxy       = "HTTP_PROXY"
+	envHTTPProxyLower  = "http_proxy"
+	envHTTPSProxy      = "HTTPS_PROXY"
+	envHTTPSProxyLower = "https_proxy"
+	envNoProxy         = "NO_PROXY"
+	envNoProxyLower    = "no_proxy"
+	envRequestMethod   = "REQUEST_METHOD"
+)
+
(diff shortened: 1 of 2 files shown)


dmke (approved):

marcschwaiger:
Hello @ldez I wanted to let you know that I am using this feature now and it works great. Thank you very much for the implementation.

PR #2721, opened 2025-11-21
allinkl: detect zone through API

### LINKED ISSUE #2538: all-inkl: adding zone override for multi-level subdomain dns challenge cert creation
### Welcome

### How do you use lego?

Docker image

### Effective version of lego

v4.23.1

#### Problem Description

I am generating certificates for multi-level subdomains (e.g., `foo.bar.example.com`) using DNS challenges with the All-Inkl provider. The domain `example.com` is registered with All-Inkl without custom NS delegation for `bar.example.com` (despite potential local DNS differences).

**Expected behavior**: Zone for `foo.bar.example.com` should be `example.com` for this All-Inkl setup, not `bar.example.com` (expected to have an option to override).

#### Current Issue

The All-Inkl provider incorrectly attempts to update the `bar.example.com` zone via KasApi, but this zone doesn't exist at All-Inkl. The TXT records should be created in the `example.com` zone instead.

#### Requested Feature

Add an environment variable to override the automatically detected zone for the All-Inkl provider, allowing manual specification of the correct zone (e.g., `example.com`) when the auto-detection logic fails.

#### Implementation Status

I have prepared a tested modification that adds a provider-specific environment variable for zone override without changing default behaviour by adding 

EnvZoneOverride 	  = envNamespace + "ZONE_OVERRIDE"

Default / unset env does not change behaviour, tested in binary build as well as docker build.

**Question**: Is this a valid feature to request for the all-inkl provider? Can I submit my implementation as a pull request?

Thanks!

ldez:
Uses API to detect the zone.

Fixes #2538

### DIFF
--- a/providers/dns/allinkl/allinkl.go
+++ b/providers/dns/allinkl/allinkl.go
@@ -136,7 +136,7 @@ func (d *DNSProvider) Present(domain, token, keyAuth string) error {
 	for z := range dns01.DomainsSeq(info.EffectiveFQDN) {
 		_, errG := d.client.GetDNSSettings(ctx, z, "")
 		if errG != nil {
-			log.Infof("allinkl: get DNS settings zone[%s] %v", z, errG)
+			log.Infof("allinkl: get DNS settings zone[%q] %v", z, errG)
 			continue
 		}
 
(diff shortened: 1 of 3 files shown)


ldez:
https://github.com/go-acme/lego/issues/2538#issuecomment-3773841020

PR #2728, opened 2025-11-25
edgeone: add zones mapping

### LINKED ISSUE #2726: edgeone: override zone ID
### Welcome

### How do you use lego?

Other

### Effective version of lego

v4.28.1

### Detailed Description

Hello maintainer,

I would like to propose adding an optional configuration item for the EdgeOne provider: `EDGEONE_ZONE_ID`. This would be useful in scenarios such as:

1. The `ZoneName` does not exactly match the FQDN (similar to #358).
2. The API credentials lack the permission to call `DescribeZones` for listing all zones, but are restricted to accessing a specific zone.

ldez:
Closes #2726

I want to be clear for people who will find this PR: this PR is something very specific because the EdgeOne API, in some edge cases, doesn't allow us to find the zone ID.
This is not something we want to propagate to the other implementations; this is an exception, not a rule or a way to follow.

### DIFF
--- a/providers/dns/edgeone/edgeone.go
+++ b/providers/dns/edgeone/edgeone.go
@@ -41,7 +41,7 @@ type Config struct {
 	Region       string
 	SessionToken string
 
-	zonesMapping map[string]string
+	ZonesMapping map[string]string
 
 	PropagationTimeout time.Duration
 	PollingInterval    time.Duration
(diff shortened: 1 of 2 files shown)


fudiwei:
Fantastic! 👍 I truly appreciate your patience and hard work. I'm eagerly looking forward to implementing this in the upcoming release.

PR #2734, opened 2025-11-27
chore: use common implementations of the providers instead of the API clients

ldez:
When a provider uses the same API client, we can create a common provider implementation instead of duplicating the provider implemenation.


ldez:
`internal` is a mechanism for defining packages that can only be imported by nearby code, not by any other package.

This is the case here: the API clients cannot be imported inside a provider (`providers/dns/xxx`).
Those specific clients must only be imported inside the `providers/dns/internal` and not inside `providers/dns/xxx`.

The initial background was related to exposed/unexposed API, but this is not the only way to used it.

That said, I'm not a huge fan of this pattern too, but, as you said, it follows the existing pattern, and it avoids possible mistakes of not using the common providers.

PR #2757, opened 2025-12-16
Prepare release v4.30.1

ldez:
**IMPORTANT**

The 2 commits must be kept: the merge must be done with a **rebase** (neither merge commit nor squash)

2 commits:
- one for the future tagged version
- one for after the release (`detach`)


ldez:
I will bypass the review as the change is trivial, and because a release lasts 1 hour.

PR #2773, opened 2025-12-29
docs: remove email from examples

ldez:
As email is not a requirement anymore, the examples don't need to suggest to use it by default.

Related to #2736


tomkel:
Not sure if it's intentional, but "Obtain/Renew a Certificate" pages still have email throughout.


https://go-acme.github.io/lego/usage/cli/obtain-a-certificate/index.html
https://go-acme.github.io/lego/usage/cli/renew-a-certificate/index.html

ldez:
The PR was focused on the DNS documentation, but I could have cleaned everything, I will do that inside the doc of the v5.

Issue #2798, opened 2026-01-17
Roadmap v5


ldez:
I want to introduce breaking changes.

I limit the changes because I want to be able to create a "quick" release and not fall inside a tunnel of several months.
So, I will implement nothing else than the existing items in the following list.
The existing items will already took a lot of time.

The elements inside the "optional" sections will probably not be implemented in the v5.

We will continue to accept PRs on the v4 during the development of the v5.
But when the v5 will be ready, we will stop the development of the v4.

Some reminders:
- I'm not an employee of a company.
- My only incomes are donations, currently lego represents about $30/month (~25€) average before tax.
- I live in France, the poverty threshold after tax is 1216€/month (~$1411).
- I don't have gifted domains/accounts from DNS companies.
- Traefik Labs is not a sponsor of this project.
- I'm working on lego for about 10 years.
- The changes will take several weeks full time to be implemented (my hope is 1 or 2 months)

If you enjoy my work, or benefit from this work, please consider donating or asking your company to do so.

https://donate.ldez.dev/

## Library

- [ ] add context support
- [ ] new logger (slog)
- [ ] refactor DNS client usages
- [ ] remove deprecated elements
- [ ] remove deprecated DNS providers (azure, etc.)
- [ ] disable Common Name (CN) by default

## CLI

- [ ] new logger options (JSON/std, level, [source?])
- [ ] new output files/directories structure
- [ ] one command (run/renew)
- [ ] new configuration parsing (flags + file)
- [ ] file configuration
- [ ] rename some flags (ex: domains -> domain)
- [ ] pre/post hooks
- [ ] disable Common Name (CN) by default

### Optional

- [ ] migration command (v4 -> v5) for existing certificates
- [ ] interactive migration page for the CLI flags
- [ ] interactive UI (at least CLI)
- [ ] daemon

## Documentation

- [ ] migration guide

xlionjuan:
> Traefik Labs is not a sponsor of this project.

No way.....

ldez:
The v5 is ready. https://ldez.github.io/blog/2026/05/11/lego-v5/

I will ask you to not comment on this issue if you have questions or problems.

You should either open an [issue](https://github.com/go-acme/lego/issues/new?template=bug_report.yml) if there is a bug, or a [discussion](https://github.com/go-acme/lego/discussions/new?category=q-a) if you have questions.

This issue is still open if you want to share your appreciation, spear some love, this is something important to me.

[...]

PR #2799, opened 2026-01-17
feat: add support for context

ldez:
This PR introduces context support.

This is the first part about this topic, the rest will come with other PRs related to other topics.

The support of context for `Present` and `CleanUp` (and so DNS provider implementation) is not in this PR, but will come later.

Related to #2798

Related to #2221
Related to #970


acme/api/internal/nonces/nonce_manager.go
```diff
@@ -55,11 +56,13 @@ func (n *Manager) Nonce() (string, error) {
 		return nonce, nil
 	}
 
-	return n.getNonce()
+	// TODO(ldez): the Nonce method signature cannot be changed because it must implement jose.NonceSource.
+	// Maybe use a dirty context struct field in this case.
```
dmke: I'd be OK with a context field here (and muting the containedctx linter).

I'd also suggest placing a interface satisfaction check near the `Manager` struct definition (or in the test file):

```go
var _ jose.NonceSource = (*Manager)(nil)
```

dmke (changes requested):

acme/api/internal/nonces/nonce_manager.go
ldez: This check is not needed because the code already doesn't compile when you change the signature of the `Nonce` method.

dmke (approved):

Issue #1553, opened 2021-12-26
JSON output mode for CLI


lenovouser:
### Welcome

- [X] Yes, I've searched similar issues on GitHub and didn't find any.

### How do you use lego?

Binary

### Detailed Description

It'd be great if a JSON output mode for the CLI could be added (the stuff that is usually text based)

This would be extremely useful in the case of invoking the CLI programmatically, for example to know when roughly to retry when errors like this happen:

```
acme: error: 429 :: POST :: https://acme-v02.api.letsencrypt.org/acme/new-order :: urn:ietf:params:acme:error:rateLimited :: Error creating new order :: too many certificates (5) already issued for this exact set of domains in the last 168 hours
```

lenovouser:
The need for this was also mentioned by @ambis in https://github.com/go-acme/lego/issues/216#issuecomment-986303145

Issue #1531, opened 2021-11-17
stdout/stderr to log


Anagastes:
### Welcome

- [X] Yes, I've searched similar issues on GitHub and didn't find any.

### How do you use lego?

Binary

### Detailed Description

I am using the current LEGO port for mips and have now tried with logger and tee to log the bash output for later error analysis in case of problems.

But LEGO doesn't seem to output a classic stdout/stderr and I can't catch the output.

If there is no other way technically, a logging function would be nice :)

dmke:
Lego logs to `stderr`. To capture logs through `tee`, you need to redirect that output:

```console
$ lego ... 2>&1 | tee -a lego.log
```

bpereto:
I would like to see the standard behaviour, that errors go to "stderr" and info and related to stdout.

You can take advantage to cron with filtered stdout but enabled errors to receive an email with the error message when a error is logged on stderr.
Or any other ideas, how "email me when error" could be done with lego (without a huge wrapper script) ?

skrivy:
I support this. Errors are supposed to go to the stderr and stdout is for the rest.

hubertbanas:
Are there any chances to get it implemented?

Issue #1134, opened 2020-05-04
Change Log Level


ghost:
When integrating the go-acme/lego as a library into e.g. another CLI / program, the log level is automatically set to INFO instead of e.g. ERROR.

Expected output:
`lego: Please create the following TXT record in your test.com. zone:`

Actual output:
```
2020/05/04 11:46:14 [INFO] acme: Registering account for foo@bar.com
2020/05/04 11:46:15 [INFO] [test.com] acme: Obtaining bundled SAN certificate
2020/05/04 11:46:15 [INFO] [test.com] AuthURL: https://acme-v02.api.letsencrypt.org/acme/authz-v3/4356253034
2020/05/04 11:46:15 [INFO] [test.com] acme: Could not find solver for: tls-alpn-01
2020/05/04 11:46:15 [INFO] [test.com] acme: Could not find solver for: http-01
2020/05/04 11:46:15 [INFO] [test.com] acme: use dns-01 solver
2020/05/04 11:46:15 [INFO] [test.com] acme: Preparing to solve DNS-01
lego: Please create the following TXT record in your test.com. zone:
```

I have not found anything to set the log level from outside, is there something I am missing?

ldez:
Hello,

the log level is not really configurable, because we use a simple wrapper around the std logger.

https://github.com/go-acme/lego/blob/master/log/logger.go

ivanjaros:
@Idez you should remove the `acme: Could not find solver for..` errors since there is no reason for them to log anything. If I set http challenge i do not care that the other ones are not set. There should be log/error only if there are no challenges set.

ldez:
There are many cases where those messages are useful. (ex: when you set multiple challenge types)
And as I already said, there are no log levels in lego.

ivanjaros:
> when you set multiple challenge types <------- exactly

but when i don't set them, don't bother me with messages.

CoderYellow:
+1

PR #2800, opened 2026-01-17
feat: new logger (slog)

ldez:
This PR introduces `slog` as a logger.

The old implementation was using `log` package, and it was very limited.

Now, there are levels and structured information.

This PR is limited to lego as a library, the CLI part is "just plugged" but not yet rewritten.
The CLI part will be done in a separate PR when the CLI is rewritten.

The implementation is inspired by the std lib `slog` package.

https://github.com/golang/go/blob/7251c9e0f00a6d7d37bb441f3e823c160131e9b5/src/log/slog/logger.go

I extended the logger with a `LazySprintf` function because `slog` is not able to format the message because the arguments are structured values and not placeholders. (the problem is related to variadic: it's not possible to have 2 variadic parameters)
The usage is rare and should be avoided, but there are some legitimate cases.

Related to #2798

Related to #866
Related to #1134
Related to #1531
Related to #1553

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -280,6 +280,9 @@ linters:
         text: 'SA1019: linodego\.(DomainsPagedResponse|DomainRecordsPagedResponse) is deprecated'
         linters:
           - staticcheck
+      - path: log/logger.go
+        linters:
+          - gochecknoinits
 
 issues:
   max-issues-per-linter: 0
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -101,9 +101,9 @@ linters:
             - Print
             - Printf
             - Warn
-            - Warnf
             - Fatal
             - Fatalf
+            - LazySprintf
     misspell:
       locale: US
       ignore-rules:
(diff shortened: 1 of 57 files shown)


dmke (approved):
Looks good, AFAICT.

As a matter of taste, I prefer to use `slog.Attr` values instead of key/value pairs, and move each attrs on a separate line, i.e.

```go
slog.Info("message",
    slog.Int("id", 64),
    slog.String("domain", domain)

// instead of
slog.Info("message", "id", 64, "domain", domain)
```

I found style that much more readable and maintainable. It reduces the diff size when adding/removing log fields, and mishaps when there's an unmatched key without value (IIRC, `slog` panics in those cases).

It does require a little thought when building a convenience layer on top of `slog`. If you're interested, I've started something like this here: https://github.com/digineo/texd/tree/refactor/slog/xlog (that branch isn't yet merged, but I've copied the code into multiple other projects, mostly unchanged).

(Maybe this ought to be a separate refactoring step? I'm leaving it up to you how to proceed here.)

ldez:
Inside the lib part, I don't want to deviate from the standard slog interface.

But we can use slog attributes, let's postpone that for later (with the CLI part) because it will have too much impact on my current branches based on this PR.

PR #1984, opened 2023-08-03
Add -4 and -6 flags (redo)

jsumners:
This PR is a redo of #1802. Since that PR has been idle so long, the branches have diverged quite a bit and it was easier to start anew.

The work in this PR includes the work originally done by @dmke in #1802.

This PR is to resolve #1801.

Issue #1406, opened 2021-05-14
desec: hangs forever at waiting for DNS record propagation


FantasyCookie17:
- [x] Yes, I'm using a binary release within 2 latest releases.
- [x] Yes, I've searched similar issues on GitHub and didn't find any.
- [x] Yes, I've included all information below (version, config, etc).

### What did you expect to see?
I expected that `lego` would obtain a certificate once the DNS record was updated.

### What did you see instead?
`lego` hangs forever at repeatedly "waiting for DNS record propagation" even though `dig -tTXT _acme-challenge.mumble.artemislena.eu` shows me that the record has been updated successfully already.

### Steps to reproduce

1. Install NixOS
2. `git clone https://codeberg.org/FantasyCookie17/nixos-server-configs/ && cd nixos-server-configs && ./apply.sh mumble` Note: The relevant config from there is this:
```nix
security.acme = {
    acceptTerms = true;
    email = "fantasycookie17@artemislena.eu";
    certs.mumble = {
    domain = "mumble.artemislena.eu";
    dnsProvider = "desec";
    credentialsFile = "/etc/acme/desec";
    postRun = "systemctl restart murmur";
    group = "murmur";
    };
};
```
3. Change `security.acme.certs.mumble.domain` to what you have on deSEC
4. Put something like this in `/etc/acme/desec`:
```
DESEC_TOKEN=secret

DESEC_HTTP_TIMEOUT=60
DESEC_PROPAGATION_TIMEOUT=4200
DESEC_POLLING_INTERVAL=120
```
5. `systemctl restart acme-mumble.service`

### Details

<details><summary>Version of lego</summary>

```console
$ lego --version
lego version 4.3.1 linux/arm64
```

</details>

<details><summary>Logs</summary>

```console
May 15 00:57:45 rock64-2g systemd[1]: Starting Renew ACME certificate for mumble...
May 15 00:57:45 rock64-2g acme-mumble-start[985]: + echo ac5ca4447a4f8a2d1d8b
May 15 00:57:45 rock64-2g acme-mumble-start[986]: ++ ls -1 accounts
May 15 00:57:45 rock64-2g acme-mumble-start[985]: + '[' -e certificates/mumble.artemislena.eu.key -a -e certificates/mumble.artemislena.eu.crt -a -n acme-v02.api.letsencrypt.org ']'
[...]

PR #2801, opened 2026-01-19
feat: new HTTP-01 and TLS-ALPN-01 servers constructors

ldez:
This PR introduces options to create new HTTP-01 and TLS-ALPN-01 servers.

For now, this is only available for the library.
The CLI part will be done in a separate PR when the CLI is rewritten.

Related to #2798

Related to #355
Related to #364
Related to #1406
Related to #1777
Related to #1801
Related to #1984


dmke (approved):
LGMT

PR #2806, opened 2026-01-21
feat: migrate to urfave/cli v3

### LINKED ISSUE #526: Replace --domains with --domain
Since the `--domains` argument takes *one* domain, it should not be named in plural but singular, i.e. `--domain` (which can then be specified multiple times).

I realize you don't want to just remove `--domains`, but I suggest adding `--domain` as the primary argument and the one listed in the help, and keep `--domains` hidden, marking in deprecated wherever it's mentioned.

The other arguments are singular, luckily.

### LINKED ISSUE #806: Request: make --domains to accept multiple domains
Hi,

Maybe this would be possible. I did make a Script which does read out all websites running on my web server(s). And will dynamically update or request certificates for all sites. But on some I do have multiple hostnames. In those cases I had to make for each sites exceptions and put a manual request command into it. So I always have to manually update those scripts when doing changes on the Websites. Would be much easier to put multiple domains into the --domains parameter. So I can use only one command.

Thanks

### LINKED ISSUE #1776: Boolean option --always-deactivate-authorizations requires string value.
### Welcome

Does it make any sense to require a _String_ value for the `--always-deactivate-authorizations` boolean option?
Other boolean options like `--no-bundle`, `--must-staple` behave as they should.

https://github.com/go-acme/lego/blob/a7b11e0447290f8a8a6e1b38317fbf7e604ca890/cmd/cmd_run.go#L50-L52
https://github.com/go-acme/lego/blob/a7b11e0447290f8a8a6e1b38317fbf7e604ca890/cmd/cmd_renew.go#L70-L72

### What did you see instead?

I get `flag needs an argument: -always-deactivate-authorizations` error.

### How do you use lego?

Binary

### Reproduction steps

```
lego --email test@test.test -d site.com --dns manual run --always-deactivate-authorizations
```

### Version of lego

```console
lego version v4.9.1 linux/amd64
```

### Logs

<details>

```console
# lego --email test@test.test -d site.com --dns manual run --always-deactivate-authorizations
Incorrect Usage: flag needs an argument: -always-deactivate-authorizations

NAME:
   lego run - Register an account, then create and install a certificate

USAGE:
   lego run [command options] [arguments...]

OPTIONS:
   --always-deactivate-authorizations value  Force the authorizations to be relinquished even if the certificate request was successful.
   --must-staple                             Include the OCSP must staple TLS extension in the CSR and generated certificate. Only works if the CSR is generated by lego. (default: false)
   --no-bundle                               Do not create a certificate bundle by adding the issuers certificate to the new certificate. (default: false)
   --preferred-chain value                   If the CA offers multiple certificate chains, prefer the chain with an issuer matching this Subject Common Name. If no match, the default offered chain will be used.
   --run-hook value                          Define a hook. The hook is executed when the certificates are effectively created.

2022/11/28 20:02:51 flag needs an argument: -always-deactivate-authorizations
```

</details>

_No response_

ldez:
I don't know if I will keep urfave/cli, but for now we will switch to v3 because the context management is better.
This allows having something more homogeneous before the CLI refactoring.

The current version of urfave/cli v3 seems stable enough to migrate to it.

As a side effect, the `--domains` flags supports `,` as separator.

The following syntax are supported:
```
-d example.com -d '*.example.com'
-d 'example.com,*.example.com'
--domains example.com --domains '*.example.com'
--domains 'example.com,*.example.com'
```

Fixes #526
Fixes #806

Fixes #1776

Closes #2501


dmke (approved):
:+1:

PR #2808, opened 2026-01-21
chore: replace rainycape/memcache by bradfitz/gomemcache

ldez:
[`rainycape/memcache`](https://github.com/rainycape/memcache) was a hard fork of [`bradfitz/gomemcache`](https://github.com/bradfitz/gomemcache).

But it seems that `bradfitz/gomemcache` is more actively maintained: `rainycape/memcache` has not been updated since 2015.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -27,6 +27,7 @@ require (
 	github.com/aws/aws-sdk-go-v2/service/sts v1.41.5
 	github.com/aziontech/azionapi-go-sdk v0.144.0
 	github.com/baidubce/bce-sdk-go v0.9.256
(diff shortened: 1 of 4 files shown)


dmke (approved):
I'm more surprised that a memcached client library need maintenance... The protocol is stupidly simple, and there is not much that could go wrong :)

ldez:
In fact, I don't know why a fork was used.
The fork was added by the initial PR about the topic: https://github.com/go-acme/lego/pull/296

But there is no explanation.

I also don't know why there is a hard fork instead of a soft fork.

dmke:
Probably because the fork claims to be a "high performance rewrite" :shrug:

With Lego writing on the order of 1 key per 60d, I don't believe we need to look at the performance here...

PR #2815, opened 2026-01-26
feat: add structured logging with key-value field support

fgouteroux:
Hi,

First, thanks for providing this lib which is very useful and complete, I use it for my project [acme-manager](https://github.com/fgouteroux/acme-manager).

I have only one point about logging format, because when we have multiple certificates ordered or renewed in a same time it is hard to filter logs correctly for one domain.

I try to extend the desired log format in the most transparency as possible with current implementation.

Let me know your thoughs about it.

## Summary

  - Add structured logging support with optional `FieldLogger` interface
  - Introduce `log.Info()` and `log.Warn()` functions with key-value pair support
  - Migrate existing log calls to use structured fields for the `domain` context
  - Maintain backward compatibility with standard loggers

## Motivation

  This change enables users to integrate structured logging libraries (e.g., logrus, zap, slog)
  for better log aggregation and parsing in production environments. The domain information
  is now passed as a structured field rather than being embedded in the log message format.

## Changes

  **log/logger.go:**
  - Added `LevelLogger`, `FieldLogger`, and `FieldEntry` interfaces
  - Added `Info()` and `Warn()` functions supporting key-value structured fields
  - Updated `Infof()` and `Warnf()` to use level-aware logging when available
  - Added helper functions for field formatting fallback

  **Affected packages:**
  - `certificate/` - authorization and certificate operations
  - `challenge/dns01/` - DNS-01 challenge solver
  - `challenge/http01/` - HTTP-01 challenge solver
  - `challenge/tlsalpn01/` - TLS-ALPN-01 challenge solver
  - `challenge/resolver/` - challenge solver management
  - `cmd/` - renewal command

## Example Usage

  ```go
  // With standard logger (backward compatible)
  log.Info("acme: Obtaining certificate", "domain", "example.com")
  // Output: [INFO] acme: Obtaining certificate [domain=example.com]

  // With logrus (structured JSON)
  type LogrusWrapper struct {
      *logrus.Logger
  }

  func (l *LogrusWrapper) WithFields(fields map[string]any) log.FieldEntry {
      return l.Logger.WithFields(logrus.Fields(fields))
  }

  logger := logrus.New()
  logger.SetFormatter(&logrus.JSONFormatter{})
  log.Logger = &LogrusWrapper{Logger: logger}

  log.Info("acme: Obtaining certificate", "domain", "example.com")
  // Output: {"level":"info","msg":"acme: Obtaining certificate","domain":"example.com","time":"..."}

### DIFF
--- a/certificate/authorization.go
+++ b/certificate/authorization.go
@@ -1,6 +1,7 @@
 package certificate
 
 import (
+	"fmt"
 	"time"
 
 	"github.com/go-acme/lego/v4/acme"
@@ -40,7 +41,7 @@ func (c *Certifier) getAuthorizations(order acme.ExtendedOrder) ([]acme.Authoriz
 	}
 
 	for i, auth := range order.Authorizations {
-		log.Infof("[%s] AuthURL: %s", order.Identifiers[i].Value, auth)
+		log.Info(fmt.Sprintf("AuthURL: %s", auth), "domain", order.Identifiers[i].Value)
 	}
 
 	close(resc)
--- a/certificate/certificates.go
+++ b/certificate/certificates.go
@@ -164,9 +164,9 @@ func (c *Certifier) Obtain(request ObtainRequest) (*Resource, error) {
 	domains := sanitizeDomain(request.Domains)
 
 	if request.Bundle {
-		log.Infof("[%s] acme: Obtaining bundled SAN certificate", strings.Join(domains, ", "))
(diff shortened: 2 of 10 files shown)


ldez:
https://github.com/go-acme/lego/blob/2ce04a6586ea27253975e10d3ab7d7bb6214c79d/.github/PULL_REQUEST_TEMPLATE.md?plain=1#L3-L7

PR #2824, opened 2026-01-28
refactor: isolate flag usages

### LINKED ISSUE #838: --filename depreciation
I just saw in the readme that it is intended to depreciate the `--filename` option. I currently use this for two reasons:

 - I generate the exact same certificate twice, one `ecdsa` and one `rsa` version.
 - I generate other certificates which have the same `commonName` / first `SAN` entry

I think `lego` currently bases it's filename on the `commonName` / first `SAN` entry (at least it was done that way around the time the `--filename` option was added)

which is why the certificate I did generate first - `ecdsa` for example - would be overwritten by the one that was generated last (`rsa`) or in the 2nd case by a completely different certificate.

I didn't see an explanation yet as to why is it intended to remove that option, and is there an alternative I could use when it happens 😰?

ldez:
This is a pure refactoring PR: the goal is to isolate flag usages as much as possible without introducing intermediate structs.

There is no functional change outside the remove of the deprecated `--filename` flag.

Some flags were not used by some commands to I removed them from those commands.

One commit, one step.

Related to #838


cmd/flags.go
```diff
@@ -313,25 +342,208 @@ func CreateAccountFlags() []cli.Flag {
...
+		},
+		CreatePathFlag(false),
+	}
+}
+
+func CreateDomainFlag() cli.Flag {
+	return &cli.StringSliceFlag{
+		Name:    flgDomains,
+		Aliases: []string{"d"},
+		Usage:   "Add a domain to the process. Can be specified multiple times or use comma as a separator.",
```
dmke: That reads a little weird. Maybe like this?

```suggestion
		Usage:   "Add a domain. For multiple domains either repeat the option or provide a comma-separated list.",
```

dmke (changes requested):

cmd/flags.go
dmke: Do you know whether `-d a,b -d c,d` would yield `[]string{"a","b","c","d"}`?

dmke (approved):

PR #2826, opened 2026-01-28
refactor(cmd): replace Fatal with error

ldez:
Replace `log.Fatal` with `return` + `error`.

This is a better practice, and it allows some simplifications.

Some `log.Fatal` are still present, but only for edge cases that should never happen.


dmke (approved):
LGTM

PR #2828, opened 2026-01-29
fix: deduplicate authz for DNS01 challenge

### LINKED ISSUE #2821: dns-01: some CAs may return the same challenge token for several authorizations
### Welcome

### How do you use lego?

Library

### Effective version of lego

latest

## Problem

When requesting both a base domain and wildcard domain simultaneously (e.g., `example.com` and `*.example.com`), some CAs (like LiteSSL) may return the same challenge token for both authorizations.

Since both authorizations have the same `Identifier.Value` (`example.com`) and the same `Token`, they would require setting identical TXT records at `_acme-challenge.example.com`.

This causes issues because:
1. Many DNS providers fail when trying to create a duplicate TXT record
2. Or the second `Present()` call overwrites the first one

## Expected Behavior

The solver should detect duplicate challenges (same `Identifier.Value` + `Token`) and skip redundant [PreSolve](cci:1://file:///Users/goll/go/src/lego/challenge/dns01/dns_challenge.go:74:0-101:1) and [CleanUp](cci:1://file:///Users/goll/go/src/lego/challenge/resolver/prober.go:24:1-24:48) calls.

## Proposed Solution

Add deduplication logic in [challenge/resolver/prober.go](cci:7://file:///Users/goll/go/src/lego/challenge/resolver/prober.go:0:0-0:0) to track presented challenges and skip duplicates in both [parallelSolve](cci:1://file:///Users/goll/go/src/lego/challenge/resolver/prober.go:162:0-217:1) and [sequentialSolve](cci:1://file:///Users/goll/go/src/lego/challenge/resolver/prober.go:112:0-160:1) functions.

I have a working implementation ready to submit as a PR once this issue is approved.

ldez:
The RFC is not super clear about the topic.

Technically this is not limited to DNS01 challenge, but the other challenges don't have the same constraints with the token.

This is niche because of the specific behavior of one CA, but also I think it should impact only a few DNS provider that don't support duplicated TXT records.

Fixes #2821

### DIFF
--- a/challenge/resolver/prober.go
+++ b/challenge/resolver/prober.go
@@ -98,11 +98,24 @@ func (p *Prober) Solve(authorizations []acme.Authorization) error {
 }
 
 func sequentialSolve(authSolvers []*selectedAuthSolver, failures obtainError) {
+	// Some CA are using the same token,
+	// this can be a problem with the DNS01 challenge when the DNS provider doesn't support duplicate TXT records.
+	// In the sequential mode, this is not a problem because we can solve the challenges in order.
+	// But it can reduce the number of call the DNS provider APIs.
+	uniq := make(map[string]struct{})
+
 	for i, authSolver := range authSolvers {
 		// Submit the challenge
 		domain := challenge.GetTargetedDomain(authSolver.authz)
 
+		chlg, _ := challenge.FindChallenge(challenge.DNS01, authSolver.authz)
+
 		if solvr, ok := authSolver.solver.(preSolver); ok {
+			if _, ok := uniq[authSolver.authz.Identifier.Value+chlg.Token]; ok && chlg.Token != "" {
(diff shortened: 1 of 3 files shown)


ldez:
Currently, inside LE, the tokens can be the same for several challenges (DN01, HTTP01) but only for one unique authz.

IMHO, the LiteSSL implementation is abusing of the RFC, and my guess is that will lead to security problem.

- one authz, same token for challenges -> OK
- two authz, but will only DNS01 challenges (no HTTP01) -> Maybe
- two authz, with 2 DNS01 and HTTP01 sharing the same token -> smell like a root of problems.

dmke:
I agree, but I also fail to come up with a (realistic) scenario where this could be abused. That being said, I may lack the creativity needed for this. It has been a long day, after all :)

PR #2837, opened 2026-01-30
feat: add server shortcodes

ldez:
To improve the usage, this PR allows to use short-codes instead of URL to define the ACME server.

If an URL is provided, the URL will used.

This PR also adds the support of ZeroSSL.

Closes #1501


internal/generators/ca/cas.json
```diff
@@ -0,0 +1,114 @@
...
+    "name": "Google Trust",
+    "type": "production",
+    "code": "GoogleTrust",
+    "directoryURL": "https://dv.acme-v02.api.pki.goog/directory",
+    "documentationURL": "https://docs.cloud.google.com/certificate-manager/docs/public-ca-tutorial#register-acme"
+  },
+  {
+    "name": "Google Trust staging",
+    "type": "staging",
+    "code": "GoogleTrustStaging",
```
dmke: I'd prefer a hyphen to separate the CA name from the type/environment, i.e.

- `googletrust` + `googletrust-staging`, and
- `letsencrypt` + `letsencrypt-staging`.
-

dmke (commented):
I have some thoughts :)

Let me know which direction you'll lean to, and I'll continue the review (for now, I'll call it a day) :wave:

internal/generators/ca/cas.json
ldez: I'm never confident to use value with hyphen as a value of a CLI flag because of possible shell parsing problem, so I'm trying to avoid it.
But I can change that.

cmd/flags.go
```diff
@@ -654,6 +669,23 @@ func createPathFlag(forceCreation bool) cli.Flag {
...
+	if err != nil {
+		// The flag is already validated during flag parsing.
+		log.Debug("Invalid server code, fallback to Let's Encrypt staging server.",
+			slog.String("code", cmd.String(flgServerCode)))
+
+		return lego.DirectoryURLLetsEncryptStaging
+	}
+
+	return directoryURL
+}
```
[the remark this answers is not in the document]
ldez: I found an alternative approach without parsing: still "magical" but it seems right.

dmke (approved):
LGTM :+1:

Issue #2614, opened 2025-08-06
Make it easier to renew multiple key-type certificates for a domain


mrj:
### Welcome

- [x] Yes, I've searched similar issues on GitHub and didn't find any.

### How do you use lego?

Docker image

### Effective version of lego

4.25.1

### Detailed Description

To renew the certificates for a mail server that has both ec256 and rsa2048 versions, it seems I currently have to have my renewal shell script, looping for each key type, first move that type's files into the `.lego/certificates` directory, run lego, then move the updated files back to their separate directory.

A `cert-dir` option would make this unnecessary. Otherwise an option to add the key-type to the filename.

ldez:
Related to #838

n12c:
You can also have different `--path` for different types, then either accept having multiple accounts or symlink the same account key into both `--path` directories from somewhere else.

But, both of these are lame. acme.sh supports this workflow without contortions.

ldez:
> acme.sh supports this workflow without contortions.

I recommend avoiding this type of sentence: I call it "The Phantom Menace".

You don't have the historical context of lego.
You don't know the technical constraints or other use-cases.

So I want to be friendly, and I always try to find solutions, but sometimes maintaining an OSS feels like slavery.

My only incomes are donations: lego represents ~$20 per month avg.
I think I'm pretty active as a maintainer of lego, at least, more active than the time given to me by people with those $20 (thanks to them).

People come with demands or make passive/aggressive insinuations.

I already know the problem related to this issue; the solution is not trivial inside lego, but at some point, I will improve this part.

mrj:
n12c, it seems that `acme.sh` puts the key algorithm in the cert filename. I'm not sure whether it always does this, or only when an algorithm is set on its command line. I used to do this manually for Lego using its `--filename` option. But that has now stopped working. Using `--path` to have a totally different configuration for each algorithm seems like overkill.

Ludovic, I'll likewise make one not-entirely-on-topic spiel:

Yes, n12c's comment was insensitive.

Although much of my work is only possible because of free software, I rarely donate to the open software projects I use. This is first because, as a developer with very limited resources, I'd feel like a chump because most wealthy users don't donate. I also don't want to encourage and reward the squeaky-wheel developers who nag the most (there's no implication about you here).

[...]

Issue #873, opened 2019-05-01
Store certificates by different key type


hcrgm:
I am using both rsa cert and ecc cert on nginx, but when I request both type of certificates, lego seems to store only one kind of certificate. I have done that work personally, but would you like to introduce this feature?😁

ldez:
Related to #838

hcrgm:
But what about renewal? lego won't renew cert correctly if I do so by specifying a filename.

> What's my idea

Generate domain configuration by key type when request cert the first time, also take a better cert file name into consideration. Add `key-type` opinion to `renew` subcommand, so.I can specify what cert to renew.

Add: I find it amazing that what Acme.sh(another acme client) had done on how to store them. It stores certs into folder naming from domain name, or with `ecc` suffix when issuing an ecdsa cert, which is convenient to seek and manage them. I think it would be better to learn how acme.sh save them.

ldez:
you misunderstood, I do not recommend using filename, I just make a link between the 2 issues because it's the same need.

Issue #346, opened 2017-02-02
Add switch to define file output format or include key type in filename


jauderho:
There does not appear to be a way to invoke lego multiple times using different key types to generate RSA and ECC certs for one hostname short of changing the output directory.

Use case for this would be to use nginx in hybrid config to serve up ECC certs to browsers that accept ECC and RSA to the rest.

PR #2850, opened 2026-02-12
feat: account storage per key type

### LINKED ISSUE #2793: Allow user ID to be chosen on the command line
### Welcome

### How do you use lego?

Binary

### Effective version of lego

v4.31.0

### Detailed Description

This is sort of related to #277, which was implemented in #1378 but that approach was rejected in favour of #2736.

The "user ID" component of the data storage path is chosen somewhat arbitrarily: it is the contact email used when the account was registered, or the hardcoded default `noemail@example.com` if the account was registered without an email address. It would be useful to expose this flexibility to the end user, by allowing them to choose their own user ID (e.g. via a new `--user-id` option). Decoupling it from the registration email would allow for private keys registered using the same contact details to be partitioned into different subdirectories, rather than having them all be installed in the same subdirectory.

If this is something you agree would be useful, I have a small PR tested and ready to go.

ldez:
The goal is to be able to use different keys for the same account/email.

Before the directories `keys` was badly named, because an account is linked to a type of key.
It was not possible to have two different keys for the same account/email.

With this PR, the directory `keys` is removed and the key is stored at the same level as the `account.json`.
But the parent directory is based on the key type, and so it possible to have multiple keys for the same account/email.

Before (only one key type was supported):

```
.lego
└── cahostA
    └── accounts
        └── test@example.com
            ├── account.json
            └── keys
                └── test@example.com.key
```

After (multiple key types are supported):

```
.lego
└── cahostA
    └── accounts
        └── test@example.com
            ├── RSA4096
            │   ├── account.json
            │   └── test@example.com.key
            └── EC386
                ├── account.json
                └── test@example.com.key
```

I added a migration command to migrate the old accounts to the new format.

Related to #873
Related to #838
Related to #346
Related to #2614


ldez:
I don't know if you know this new tips:
- Go to the "Commits" section https://github.com/go-acme/lego/pull/2850/commits
- Click on the first commit
- You will have "Next/Prev" buttons to move from commit to commit easily

It can be useful with my commits.

dmke (approved):

PR #2870, opened 2026-02-23
feat: change CertificateService and GetRenewalInfo signatures

ldez:
- `CertificateService.Get` returns `*acme.RawCertificate` instead of 2 slices of bytes.
- `CertificateService.GetRenewalInfo` returns `*acme.ExtendedRenewalInfo` instead of `*http.Response`.

I also remove dead code.


ldez:
Because the work on the new challenge type, and the next major step on configuration, I will "accelerate" the merge of the "low impact" PRs.

PR #2871, opened 2026-02-24
feat: Add support for DNS-PERSIST-01 challenge type

### LINKED ISSUE #2863: dns-persist-01 challenge support
### Welcome

### How do you use lego?

Binary

### Effective version of lego

4.32.0

### Detailed Description

Would love to eventually be able to use the new challenge validation: https://letsencrypt.org/2026/02/18/dns-persist-01.html

Thanks for all you do! 💙

beautifulentropy:
Implement the draft-ietf-acme-dns-persist-00 as a new ACME challenge type within lego along with all of the CLI flags and plumbing necessary to make it work properly. While the IETF draft itself has not been approved just yet, the The CA/Browser Forum ballot SC-088v3, defining “3.2.2.4.22 DNS TXT Record with Persistent Value”, passed unanimously in October 2025: https://cabforum.org/2025/10/09/ballot-sc-088v3-dns-txt-record-with-persistent-value-dcv-method. The ballot featured language that was extremely consistent with the current draft so I doubt that the draft will see significant changes.

For more information about this new challenge type, please see:

- My blog post: https://letsencrypt.org/2026/02/18/dns-persist-01
- Draft specification: https://datatracker.ietf.org/doc/draft-ietf-acme-dns-persist/
- Relevant CABF BRs section: https://github.com/cabforum/servercert/blob/main/docs/BR.md#322422-dns-txt-record-with-persistent-value

Note: the bulk of the DNS resolver/propagation code has been taken directly from the lego-cli v4 DNS-01 (resolver) client, but stripped down into only what is necessary for this challenge type and refactored to avoid mutated global vars.

There are to optional flags added where special care was taken to provide a better UX:
1. dns-persist.persist-until: This maps directly to an optional parameter on the eventual DNS record, but as UTC Unix seconds. Instead, we accept an RFC 3339 timestamp here.
1. dns-persist.issuer-domain-name: This maps directly to a required parameter that is usually sourced deterministically from the challenge object itself. When provided as an argument it will need to match one of the issuer-domain-names included on the challenge. As such, we do our best to normalize the provided value (e.g. BÜCHER.example instead of xn--bcher-kva.example) to improve the likelihood of a match.

Also, updates Pebble and pebble-challtestsrv to v2.10.0, which include the following fixes:
- https://github.com/letsencrypt/challtestsrv/pull/32
- https://github.com/letsencrypt/challtestsrv/pull/34
- https://github.com/letsencrypt/pebble/pull/535
- https://github.com/letsencrypt/pebble/pull/536

### Caveats

- Unlike previous ACME challenge types, DNS-PERSIST-01 depends on an existing account URI before challenge solvers can be initialized, so calls to setupChallenges() have been moved to after account initialization.

- DNS-PERSIST-01 also does not use a token value, so the token field in the challenge is now optional along with a new field which is exclusive to this new challenge, issuerDomainNames ([]string).

- DNS-PERSIST-01 was drafted with the intention of keeping DNS credentials out of the issuance pipeline. The implementation I’ve put forward is intentionally manual-only. It’s also worth noting, because the authorization is intended to be persistent, there is no cleanup step or operation included as part of its API.

- Given the manual-only nature of DNS-PERSIST-01 I’ve also intentionally de-prioritized it in favor of DNS-01 when both challenge types are provided. Feel free to remove this if you disagree with that ordering.  

Closes #2863


challenge/dnspersist01/dns_persist_challenge_test.go
```diff
@@ -0,0 +1,437 @@
...
+			}
+			actual := chlg.hasMatchingRecord(test.records, test.issuer, test.wildcard)
+			assert.Equal(t, test.expect, actual)
+		})
+	}
+}
+
+func ptrTime(t time.Time) *time.Time {
+	return &t
+}
```
ldez:
```suggestion
// TODO(ldez) factorize
func Pointer[T any](v T) *T { return &v }
```

challenge/dnspersist01/issue_values_test.go
```diff
@@ -0,0 +1,206 @@
...
+			issuer:   "authority.example",
+			wildcard: true,
+			expect:   "authority.example; policy=wildcard",
+		},
+		{
+			desc:       "with persistUntil",
+			issuer:     "authority.example",
+			accountURI: "https://authority.example/acct/123",
+			wildcard:   true,
+			persistUTC: ptrIssueTime(time.Unix(4102444800, 0).UTC()),
```
ldez:
```suggestion
			persistUTC: Pointer(time.Unix(4102444800, 0).UTC()),
```

ldez (changes requested):
Could you explain why the PR is in draft mode?

beautifulentropy:
> Could you explain why the PR is in draft mode?

I am finishing up the description and also fixing lints.

ldez:
Just a note: the design of the DNS calls is based on the v4, not the v5.

I will not ask you to change it, because it will require important refactors and some notable factorizations, but I will do it after the merge of the PR.

beautifulentropy:
> Just a note: the design of the DNS calls is based on the v4, not the v5.
>
> I will not ask you to change it, because it will require important refactors and some notable factorizations, but I will do it after the merge of the PR.

Thanks, I appreciate it.

ldez (approved):
LGTM, thank you

ldez:
For people who are watching this PR: the [v5](https://github.com/go-acme/lego/issues/2798) is still in WIP, so I can and will break things.

Also, I will rewrite the DNS-PERSIST-01 implementation related to DNS calls to fit the v5 design, this will probably impact the preexisting elements (lib and CLI).

PR #2901, opened 2026-03-08
feat: hook manager

### LINKED ISSUE #848: Add hook for script before requesting a new certificate
I'm using [`acme-dns`](https://github.com/joohoi/acme-dns) as provider for DNS challenges and I'd like to start it when it's needed and stop it when it's not needed anymore (and additionally copying over certificates and restarting services).

For this various hooks would be great :) A few days ago there was a [PR](https://github.com/go-acme/lego/pull/845) merged which adds a post hook. Are there any plans to add a pre hook too? (which e.g. would be executed when a certificate will get renewed)

### LINKED ISSUE #894: Feature request: pre/post/deploy hook
Thank you for this great product.

We are working on an implementation of lego, but we miss an essential feature for our implementation. For our use case lego should run a shell script, when a certificate should be renewed.
This feature is similar to the certbot flags --pre-hook and --post-hook.
We also want to push the certificate to Hashicorp Vault when the certificate is actually been renewed. This feature is similar to the certbot flag --deploy.

### LINKED ISSUE #2759: Renewal dry-run
### Welcome

### How do you use lego?

Binary

### Effective version of lego

4.30.1

### Detailed Description

Firstly, thank you for writing this awesome software. I've been using it for a while now.

Now onto my feature request...

Forgive me if I am missing something, but It would be nice if there was a way to test if a renewal is necessary and not actually do it. As of now, as far as I can tell, there is no way to do this. You just run the renewal, and if it's not necessary, it says so in standard out and exits 0.

But if you are writing scripts that need to do a more complex operation or other tasks related to renewal being needed (before running the actual renewal), you'd have to run lego, get the output, and check if the no renewal needed string is in there. and it would just renew if necessary in that situation, which might not work with your business process.

I'll probably just make a little shell function to use openssl to examine the certificate date, but it would be a lot more convenient if lego itself could do this.

If there is an existing way to do it better than that, I'm all ears!

ldez:
Add `pre-hook` and `post-hook` for `run` and `renew` commands.

Fixes #848
Fixes #2759

### DIFF
--- a/cmd/cmd_renew.go
+++ b/cmd/cmd_renew.go
@@ -61,11 +61,6 @@ func renew(ctx context.Context, cmd *cli.Command) error {
 
 	certsStorage := storage.NewCertificatesStorage(cmd.String(flgPath))
 
-	meta := map[string]string{
-		// TODO(ldez) add account ID.
-		hook.EnvAccountEmail: account.Email,
-	}
-
 	lazyClient := sync.OnceValues(func() (*lego.Client, error) {
 		client, err := newClient(cmd, account, keyType)
 		if err != nil {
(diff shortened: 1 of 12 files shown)


dmke (approved):
LGTM, just found a duplicate space.

meineerde:
Three remarks:

* The description of the hooks is grammatically strange. I think there are generally too many commas and a superfluous `is`. Shouldn't this read more like
  * `--deploy-hook string            Define a hook. The hook runs after the renewal, in cases where a certificate is successfully created/renewed. [$LEGO_DEPLOY_HOOK]`
  * `--post-hook string              Define a post-hook. This hook runs after the renewal in cases where a certificate renewed regardless of whether any errors occurred. [$LEGO_POST_HOOK]`
  * `--pre-hook string               Define a pre-hook. This hook is runs before the renewal in cases where a certificate will be effectively renewed. [$LEGO_PRE_HOOK]`
* Will the `--pre-hook` and `--post-hook` also applied during initial creation or just during actual renew? This is clearly mentioned for the `--deploy-hook`, but the documentation for the the others implies that they only run during renewal.
* The `cmd/internal/hook/metdata.go` file (and its test file) should probably be named `metadata.go`?

ldez:
1. The descriptions of the hooks are "grammatically strange" because this is complex to provide a clear and short explanation of the flags. I will not change them.
2. The `post` and `pre` hook are launched with the `run` command or the `renew` command. I can improve this part.
3. typos

meineerde:
> 1. The descriptions of the hooks are "grammatically strange" because this is complex to provide a clear and short explanation of the flags. I will not change them.

Fair enough. "The hook is runs" in the descriptions for `deploy-hook` and `pre-hook` is still wrong though (it's correct in the `post-hook` description).

ldez:
PTAL #2903

PR #2911, opened 2026-03-11
feat: support a configuration file

### LINKED ISSUE #121: Add support for a configuration file
The command line call gets more and more cluttered, esp. with the addition of the dns solver and its wide variety of configuration through environment variables. 

Wouldn't it be nice to have a structured configuration file to specify the options?

### LINKED ISSUE #453: Feature request: Configuration file support
With the multitude of options lego supports it would be greatly appreciated if it supported reading its configuration from a file instead of having to specify every parameter on the command line.

I'm a huge fan of [hcl](https://github.com/hashicorp/hcl), configuring Hashicorp products using it is a blast, but I'd be open for anything at this point :smile: 

Here's a reference implementation: https://github.com/hashicorp/vault/blob/c44f1c9817955d4c7cd5822a19fb492e1c2d0c54/command/config.go

ldez:
The root command now accepts a configuration file.
The default configuration file is `.lego.yml`, but you can specify a different file with the `--config` flag.

I spend a lot of time (maybe too much) on the configuration file, I tried several approaches, and this one seems to be the best for me.

The most basic configuration files are:

<details>
<summary>HTTP challenge</summary>

```yaml
# Use HTTP challenge to obtain a certificate.

certificates:
  foo:
    challenge: http-01
    domains:
      - example.com
```

</details>

<details>
<summary>TLS challenge</summary>

```yaml
# Use TLS challenge to obtain a certificate.

certificates:
  foo:
    challenge: tls-alpn-01
    domains:
      - example.com
```

</details>

<details>
<summary>DNS challenge</summary>

```yaml
# Use DNS challenge to obtain a certificate.

challenges:
  cf:
    dns:
      provider: cloudflare

certificates:
  foo:
    challenge: cf # Optional, if there is only one challenge.
    domains:
      - example.com
      - '*.example.com'
```

</details>

There is a reference configuration file in `cmd/internal/configuration/testdata/reference.yml` (this is not a configuration example).
This file is temporary and will be removed in the future, and proper documentation will be added.

The support of PFX certificate output is not yet implemented because I want to think about the best way to handle it (maybe something related to a deployment pipeline).

Also, I plan to add a way to handle the DNS credentials inside the configuration file, but for now, I don't find a good and secure way to do it.

The hooks are not implemented because there are several possibilities, and I want to think about that in a separate topic.

The certificates and the accounts are automatically archived (zip) in the `.lego/archive` directory when a certificate ID or an account ID is removed from the configuration file.
The archived elements are automatically deleted after 30 days.

Some of the code is a kind of duplication of the CLI part, I don't want to factorize it yet (if it's possible to do it), I want to get feedback before doing so.

I added a JSON schema to validate the configuration file, but for now, it will stay "internal" because it will a part of the documentation. 

Despite the apparent simplicity of the code (I hope it is "relatively simple"), this takes me a lot of time to design.
I think I fully implemented the thing 4 times from scratch before finding the current solution.

I made one commit per change to ease the review process.
I isolated refactoring commits and feature commits.

Fixes #121
Fixes #453


ldez:
The CI fails:
- on Windows because Windows file system is a piece of ****.
- with e2e tests because some port conflicts

I will skip the tests related to FS on Windows.

And I will change the ports inside e2e.

meineerde:
Great work on the config file implementation!

In the `reference.yml`, it appears that the `server` is referenced by its URL for an `accounts` entry. As far as I can understand from the code (although I don't know any Go), this appears to help support the default URL in `lego.DirectoryURLLetsEncrypt`. However, wouldn't it be more consistent to reference servers by their name (i.e. `"myserver"` in the `reference.yml` example) rather than their URL? Please excuse me if I have missed the logic somewhere.

As for using this configuration file, as far as I understand the code, Lego does not write this file at all but only reads it. This i find desireable as it helps creating the configuration via config management systems and not having to deal with it changing from random invocations.

To ease working with config management systems however, I'd like to ask for an option where "partial" configurations could be created in a directory, all of which would be loaded in a defined order to create a deep-merged final configuration. 

That way, it would be easy to create partial configuration files e.g. containing only the data for a single certificate, rather than having to manually the details for all requested certificates and having to write a single combined YAML file. Alternatively, an explicit `include` syntax could be used to include files (or multiple files via globs) in the current section.

When adding multiple additional certificates, this would also make it easier to sequentially define (and issue) additional certificates one after another while avoiding to generate a new single config file multiple times.

Finally, having the option for separate files would also possibly help with securing (DNS-)credentials on disk as the the challenges (and their credentials) could be defined in separate files with more restrictive file system permissions.

ldez:
> In the reference.yml, it appears that the server is referenced by its URL for an accounts entry. 

Nope, the field `server` in the `account` can be a URL (not related to `servers`), an existing CA code, or a reference to the `servers` configuration.

ldez:
> As for using this configuration file, as far as I understand the code, Lego does not write this file at all but only reads it. This i find desireable as it helps creating the configuration via config management systems and not having to deal with it changing from random invocations.

Yes, I don't plan to allow lego to write files.

But this is something I may add later inside the migration command.

The configuration and the `run`/`renew` are not really "compatible" together.

Also, in Go, there is no YAML library parsing that allow keeping comments, so any programmatic edition will delete all the comments (I will not play with regexp :wink:).

ldez:
> Finally, having the option for separate files would also possibly help with securing (DNS-)credentials on disk as the the challenges (and their credentials) could be defined in separate files with more restrictive file system permissions.

I talk about that inside the PR description:

> Also, I plan to add a way to handle the DNS credentials inside the configuration file, but for now, I don't find a good and secure way to do it.

dmke:
At a first glance this looks great :+1:

I'll review this in depth over the course of the next few days, time permitting.

ldez:
> To ease working with config management systems however, I'd like to ask for an option where "partial" configurations could be created in a directory, all of which would be loaded in a defined order to create a deep-merged final configuration.
> ...
>  Alternatively, an explicit include syntax could be used to include files (or multiple files via globs) in the current section.

I will not answer your proposals with a yes or a no.

This PR is a first step, I spend a lot of time (several weeks) on this part to get a good base.
It can feel "simple" or maybe "basic" but this is the goal.

Also I tried to create a relatively thin PR because I know how it can be difficult to review this kind of PR.

So one step after another.

meineerde:
> Nope, the field `server` in the `account` can be a URL (not related to `servers`), an existing CA code, or a reference to the `servers` configuration.

Ah yes, I can now see this logic in the `GetServerConfig` function with the multiple possible returns. Sorry for the noise.

> I talk about that inside the PR description:
>
> > Also, I plan to add a way to handle the DNS credentials inside the configuration file, but for now, I don't find a good and secure way to do it.

Yeah, I saw that. My proposal was (partially) meant to suggest a solution for this challenge on how to store credentials :)

> The configuration and the run/renew are not really "compatible" together.

I think that's quite fine for me. Sorry that I have not been clearer. I believe it is a desirable feature that Lego does not write or update the config file on its own. Because it does not touch the config, it is possible to generate the config entirely from external config management as a single point of truth.

> I will not answer your proposals with a yes or a no.
>
> This PR is a first step, I spend a lot of time (several weeks) on this part to get a good base.

Thank you for your work on this and on Lego in general. I'm perfectly happy with this approach and I certainly don't want to "require" any features. I'd just like to offer my perspective and use-case. It's certainly possible for me to work with a single config file as even this would save me from coming up with a homegrown config schema :)

Still, to be a bit more verbose about my preferred use-case: I'd like to entirely create the config file(s) with configuration management to define accounts, challenges and certificates. When adding a new certificate, I would prefer to just create/update a config file to define the new certificate and then to call a simple CLI command to actually request the initial certificate (i.e. any newly added or updated certificate), followed by regular invocations of the CLI (via cron, systemd-timers, ...) to renew any pending certificates according to their definitions in the YAML config. That way, the system would work kind of like a less clunky acme.sh :)

Additionally, for my use-case it would be helpful to be able to detect when the definition for an existing certificate is changed (e.g. when adding additional alternative domains or changing the profile) to be able to decide if an existing certificate should be re-issued right away or is just updated on its next regular renewal. When the desired certificates are defined in the YAML config as a single-point-of-truth, I can easily parse the existing config and comparing the existing and desired state. If required, I would then either run a CLI command to forcefully issue/renew a single defined certificate (I believe there is such a command?), or just update the config for the next regular renewal.

With that being said, I noticed another possible restriction in the config structure. It appears that the `KeyType` for the certificates is only defined for an account rather than the certificate. As such, it appears that it's not possible to request different key types (e.g. RSA2028 and EC256 keys) from a single account. I see that you added #2850 as a related feature recently. Doesn't the current configuration structure defeat this feature?

ldez:
Maybe you missed something: the command allows to `run` and `renew` at the same time.

This means: for a certificate ID (or name but I prefer to say ID) if you change the domains related to this ID, the difference will be detected and the certificate "renewed".

ldez:
> It appears that the KeyType for the certificates is only defined for an account rather than the certificate. As such, it appears that it's not possible to request different key types (e.g. RSA2028 and EC256 keys) from a single account. 

This is expected:
- For compatibility with the exiting CLI commands (keyType is a unique flag)
- The internal lego design is following the existing CLI approach
- I'm not sure it is good idea to have a difference between account keytype and certificate keytype.
- I think there is no a major problem of using one account per keytype.

ldez:
:thinking: maybe I should detect if the profile or the preferred chain as changed.
But this is outside of the scope of this PR because it will required to change the structure used to store the certificate metadata and the CLI.

meineerde (commented):
On Windows, to deleted files which are still opened process, all file handles must be created with the `FILE_SHARE_DELETE` flag.

It seems Go does not want to enable this flag by default on Windows (in contrast to many other programming languages). See https://github.com/golang/go/issues/34681#issuecomment-574782746

ldez:
Do you have a problem on Windows or this is only a comment about the fact that I skipped tests related to FS on Windows?

I'm aware of the problem between Go and Windows, these are not the first tests I skip inside the lego code base because of that.

meineerde:
> Do you have a problem on Windows or this is only a comment about the fact that I skipped tests related to FS on Windows?

Sorry, I think I messed up with the Github UI... This was meant as a comment to ceabd686b7047d9d14b53d07b983cc381f5ca69f where you disabled the tests on Windows. Just wanted to give a hint as I had a similar issue with Windows support for a Ruby library recently.

ldez:
I fixed the problem with tests on Windows (I forgot to close some writers).

Next PR: key type per certificate :wink: (the branch is ready)

ldez:
@dmke As most of the next changes are related to the same sections of the code, I will not be able to apply review changes to the PR inside the PR.

I tried to reduce the number of PRs to review to only important changes and limited the next changes to be able to apply review feedback, but the current situation creates a significant mental load.
I must keep all my local branches and other future changes always "loaded in my mind" to be ready at any time, and very constrained.
Also, my current financial situation is not good at all, so I don't know how many times I will be able to continue in this way.

All of this represents a significant amount of pressure, and I'm having some difficulty managing it.

I need to move forward: I will merge the PR as is and let you do the review on the closed PR.
I will take your feedback and apply it afterward.

Be sure that I fully understand that your time is limited, and your reviews are more important to me than you can imagine.

I don't like bypassing the review process, but I don't have a better solution, I'm sorry, and I hope you understand.

dmke:
Completely understandable.

If it helps, you could just merge your PRs directly (apart from some typos here or there, I usually don't find many issues :+1:).

When the v5 release approaches I could (try to) make block some dedicated time for reviews.

Issue #1016, opened 2019-11-28
.lego directory location


Kuchenm0nster:
Hello, I noticed the lego directory containing accounts and certificates get saved at the current working directory.
This might introduce inconsistency depending on the location you run it.
Saving the directory under $HOME like many other programs might be favorable.
Regards

ldez:
hello,

I understand your point, it's a breaking change so we will not introduce it quickly.

dmke:
Could we change the mechanism for determining the path location to prefer `$PWD/.lego`, and if that doesn't exist fallback to `$HOME/.lego` (or actually `$XDG_CONFIG_HOME/lego` on Linux)? This way existing installations would continue to work. Of course, `--path` still takes precedence.

In full, the code would look like this:

```ruby
def default_path
  # $PWD/.lego
  if (local = Pathname.new(__dir__).join(".lego")).exist?
    return local
  end

  # Linux: ${XDG_CONFIG_HOME:-$HOME/config}/lego
  home = Pathname.new ENV.fetch("HOME")
  if Etc.uname[:sysname] == "Linux"
    xdg = ENV.fetch("XDG_CONFIG_HOME", home.join(".config")))
    return Pathname.new(xdg).join("lego")
  end

  # Windows: %AppData%\Roaming\lego
  if Etc.uname[:sysname] == "Windows"
    return Pathname.new(ENV.fetch("APPDATA")).join("Roaming/lego")
  end

  # other: $HOME/.lego
  home.join(".lego")
end
```

(This is pseudo-code in Ruby, the actual implementation could make use of appropriate build tags and e.g. implement the Linux-specific code in `config_dir_linux.go` to get rid of runtime conditionals.)

ldez:
Currenlty the `./.lego` folder is created if this folder don't exist, so a fallback it's not possible without breaking the current behavior.

dmke:
Why is that?

- Processes invoking `lego` on existing installations would continue to use `./.lego`.
- Processes on new installations would create `~/.config/lego`.
- Automated processes should be configured to use `--path`.

Am I missing a use case?

PR #2974, opened 2026-04-11
feat: add experimental support for dotenv files

ldez:
There are two cases:
1. when using `lego run`
2. when using `lego` (the root command)

When using `lego run`, the environment variables are loaded from the `.env` file.

When using `lego`, the environment variables are loaded from the `.env` file, and from the file `.env.XXX` where `XXX` is the name defined in the configuration file as the challenge name (not the provider name).
The environment variables defined in the file `.env.XXX` overrides the environment variables defined in file `.env` ("merge").

This is experimental for several reasons: I don't know if it's a good idea, maybe it requires some adjustments.

In the end, this PR allows storing environment variables in files (easier to use), the configuration file doesn't store sensitive information, and the approach is relatively standard.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -253,6 +253,9 @@ linters:
       - path: cmd/cmd_run_renew.go
         text: cyclomatic complexity \d+ of func `(renewForDomains|renewForCSR)` is high
         linters:
+      - path: cmd/cmd_run.go
+        text: cyclomatic complexity \d+ of func `run` is high
+        linters:
           - gocyclo
       - path: cmd/internal/root/process_renew.go
         text: cyclomatic complexity \d+ of func `(renewForDomains|renewForCSR)` is high
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -254,11 +254,6 @@ linters:
         text: cyclomatic complexity \d+ of func `(renewForDomains|renewForCSR)` is high
         linters:
           - gocyclo
-      - path: cmd/internal/root/process.go
-        text: (cognitive|cyclomatic) complexity \d+ of func `process` is high
-        linters:
-          - gocognit
-          - gocyclo
       - path: cmd/internal/root/process_renew.go
         text: cyclomatic complexity \d+ of func `(renewForDomains|renewForCSR)` is high
         linters:
--- a/cmd/cmd_run.go
+++ b/cmd/cmd_run.go
@@ -10,6 +10,7 @@ import (
 	"github.com/go-acme/lego/v5/certcrypto"
 	"github.com/go-acme/lego/v5/cmd/internal/flags"
 	"github.com/go-acme/lego/v5/cmd/internal/storage"
+	"github.com/go-acme/lego/v5/internal/dotenv"
 	"github.com/go-acme/lego/v5/lego"
 	"github.com/go-acme/lego/v5/log"
 	"github.com/urfave/cli/v3"
@@ -36,6 +37,11 @@ func run(ctx context.Context, cmd *cli.Command) error {
 	account, err := store.Account.Get(cmd.String(flags.FlgServer), keyType, cmd.String(flags.FlgEmail), cmd.String(flags.FlgAccountID))
 	if err != nil {
 		return fmt.Errorf("set up account: %w", err)
+	}
+
+	_, err = dotenv.Load(dotenv.BaseFilePrefix)
+	if err != nil {
+		return fmt.Errorf("set up environment: %w", err)
 	}
 
 	lazyClient := sync.OnceValues(func() (*lego.Client, error) {
--- a/cmd/internal/root/process.go
+++ b/cmd/internal/root/process.go
@@ -75,39 +75,48 @@ func process(ctx context.Context, cfg *configuration.Configuration) error {
 			// each certificate is different, so the metadata is different, except for the account information.
 			hookManager := hm.Clone()
 
-			lazySetup := sync.OnceValues(func() (*lego.Client, error) {
-				client, errC := lazyClient()
-				if errC != nil {
-					return nil, fmt.Errorf("set up client: %w", errC)
-				}
-
-				client.Challenge.RemoveAll()
-
-				errC = setupChallenges(client, chlgNode.Challenge, networkStack)
-				if errC != nil {
-					return nil, fmt.Errorf("setup challenges: %w", errC)
-				}
-
-				return client, nil
-			})
-
-			for _, cert := range chlgNode.Certificates {
-				// Renew
-				if store.Certificate.ExistsFile(cert.ID, storage.ExtResource) {
-					err = renew(ctx, lazySetup, cert.ID, cert, store.Certificate, hookManager)
-					if err != nil {
-						return err
-					}
-
-					continue
-				}
-
-				// Run
-				err := obtain(ctx, lazySetup, cert.ID, cert, store.Certificate, hookManager)
-				if err != nil {
-					return err
-				}
+			err := processChallenges(ctx, lazyClient, chlgNode, store, hookManager, networkStack)
+			if err != nil {
+				return err
 			}
+		}
+	}
+
+	return nil
+}
+
+func processChallenges(ctx context.Context, lazyClient lzSetUp, chlgNode *configuration.ChallengeNode, store *storage.Storage, hookManager *hook.Manager, networkStack challenge.NetworkStack) error {
+	lazySetup := sync.OnceValues(func() (*lego.Client, error) {
+		client, errC := lazyClient()
+		if errC != nil {
+			return nil, fmt.Errorf("set up client: %w", errC)
+		}
+
+		client.Challenge.RemoveAll()
+
+		errC = setupChallenges(client, chlgNode.Challenge, networkStack)
+		if errC != nil {
+			return nil, fmt.Errorf("setup challenges: %w", errC)
+		}
+
+		return client, nil
+	})
+
+	for _, cert := range chlgNode.Certificates {
+		// Renew
+		if store.Certificate.ExistsFile(cert.ID, storage.ExtResource) {
+			err := renew(ctx, lazySetup, cert.ID, cert, store.Certificate, hookManager)
+			if err != nil {
+				return err
+			}
+
+			continue
+		}
+
+		// Run
+		err := obtain(ctx, lazySetup, cert.ID, cert, store.Certificate, hookManager)
+		if err != nil {
+			return err
 		}
 	}
 
--- a/cmd/internal/root/process.go
+++ b/cmd/internal/root/process.go
@@ -15,6 +15,7 @@ import (
 	"github.com/go-acme/lego/v5/cmd/internal/configuration"
 	"github.com/go-acme/lego/v5/cmd/internal/hook"
 	"github.com/go-acme/lego/v5/cmd/internal/storage"
+	"github.com/go-acme/lego/v5/internal/dotenv"
 	"github.com/go-acme/lego/v5/lego"
 	"github.com/go-acme/lego/v5/registration"
 )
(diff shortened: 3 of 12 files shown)


meineerde:
Hmmm, I'm not sure if it's a good idea to indiscriminately load `.env` files from the current directory. 

Depending on the execution context of lego, this may result in security issues if the working directory of a lego invocation is not strictly controlled and if an attacker is able to add a `.env` file with malicious environment variables to the working directory. This can often happen by accident if the environment is not carefully controlled.

It also feels generally surprising that the behavior of a tool is influenced by "random" files in the current working directory.

Maybe the loading and inclusion of the `.env` files should be enabled with a separate flag (which defaults to `false`)?

Alternatively, as it seems that you mainly intend this to be used to provide environment variables for the various DNS provider challenges, it may be possible to provide an explicit option for the inclusion of a specifically named `.env` file in the configuration YAML for each challenge? That way, it's up to the user to provide a path to the env file deliberately.

Even more, it could be possible to set / override the base directory (or more general: the `BaseFilePrefix`) for loaded `.env` files globally in the configuration YAML to ensure that only env files which are guaranteed to be safe are ever loaded from a specific path or prefix. From my understanding of the code, it appears that you have build something like that already for the tests with the `12b79c45_2153_4e99_9518_67b3350d878b` environment variable. For the setting in n the config file, you could even use a default of `.env` to have the same behavior as today, while providing the option to to override and/or disable this inclusion (e.g. by setting the config value to `null` or `false`).

ldez:
When I opened the PR I was sure to get feedback from you, I was surprise this doesn't happen sooner :smile: 

The `.env` is something "standard" to handle env vars in a file, so I based my implementation on this.

The main reason, I didn't add an option to add a file per challenge is the "path relativity".
From my experience with other software, I know that nobody expect the same thing when defining a relative path.
One will want it relative to the config file, one will want it relative to launch location.
I don't want to deal with that.

And forcing an absolute path is not a viable approach.

About the naming of the files:
1. `.env` name is the standard.
2. `.env` file doesn't have prefix but suffix (this is the convention)

> Depending on the execution context of lego, this may result in security issues if the working directory of a lego invocation is not strictly controlled and if an attacker is able to add a .env file with malicious environment variables to the working directory. 

IMHO, if an attacker has access to your machine, you have a bigger problem than the fact it can edit a `.env` file.

The dotenv are standard and used everywhere I'm not sure to understand why do you think this is a specific problem with lego.

> Maybe the loading and inclusion of the .env files should be enabled with a separate flag (which defaults to false)?

Maybe :thinking:  I need to think about it, but adding a flag to allow loading `.env` seems weird because the goal is to have something "transparent"/"effortless".

> Even more, it could be possible to set / override the base directory 
> ...
> it appears that you have build something like that already for the tests with the 12b79c45_2153_4e99_9518_67b3350d878b environment variable. 

This was the idea behind the hash env var.
Obviously, I don't want the user to define an env var for that, but more an option inside the file.

meineerde:
> When I opened the PR I was sure to get feedback from you, I was surprise this doesn't happen sooner 😄

😄

> The `.env` is something "standard" to handle env vars in a file, so I based my implementation on this.

That is true. However, most of the time, these files are only used during the local development of a program (such as a web application) to provide default values for required environment variables during development and testing.

Generally, these `.env` files are not used in production however. Here, other mechanisms of the respective hosting platform are used to set environment variables securely.

In the end, the main advantage of environment variables is that the (sensitive) values do not actually have to be present in a file readable to the executing user but can be provided in the environment by the, well, outside environment in a more secure way (such as the environment with Docker or by a systemd service). With that, the values can not be read and extracted from the filesystem, which helps protect against many vulnerabilities.

> The main reason, I didn't add an option to add a file per challenge is the "path relativity". From my experience with other software, I know that nobody expect the same thing when defining a relative path. One will want it relative to the config file, one will want it relative to launch location. I don't want to deal with that.

I think the accepted "standard" for `.env` files is to load files from to the current directory where the respective tool is run from, i.e. whatever is in `$PWD`. Your implementation follows this.

> About the naming of the files:
> 
>     1. `.env` name is the standard.
>     2. `.env` file doesn't have prefix but suffix (this is the convention)

Right. When I wrote about a "prefix" above, I meant either the directory of the file (which now is implied to be `./`) or the directory and an optional filename prefix (which is now `./.env` but could possibly be defined as e.g. `$HOME/.lego/env` so that lego would read this file instead of `./.env` and would read `$HOME/.lego/env.one` as an environment file for the `one` challenge.

> > Depending on the execution context of lego, this may result in security issues if the working directory of a lego invocation is not strictly controlled and if an attacker is able to add a .env file with malicious environment variables to the working directory.
> 
> IMHO, if an attacker has access to your machine, you have a bigger problem than the fact it can edit a `.env` file.

With (shared) hosting, the server operator generally does not entirely control all files on the server but allows users to upload custom files.

Depending on the setup, lego may be run for these users, e.g. to get and install certificates for custom hostnames. If it is possible that lego runs in a `$PWD` directory which is writable by the user, the system may have a new vulnerablity where users may override things they could not before.

This is not generally an issue with web applications since they often rely much more on an implicit project root and where it is common to protect this directory.

> The dotenv are standard and used everywhere I'm not sure to understand why do you think this is a specific problem with lego.

I'm not very knowledgeable in the best practices for Go application tbh., but from the top of my head, I do not know any (standalone) command line application which loads environment variables from a `.env` by default.

---

With that being said, I now realize that the default storage directory for lego is in fact `./.lego` which also relies on `$PWD` today. I forgot this here, sorry for that. As this already relies on the current directory, reading a `.env` file from there would probably not make things much "worse" in this regard.

I'm not sure how much users are actually discouraged from editing stuff within the `.lego` directory, but maybe a better default would be to read the the env files from within the storage directory by default? At least then, everything would be stored in one place, although this would create chicken-and-egg issue with the `LEGO_PATH` variable...

There's already #1016 which requested to update the default storage directory to something in `XDG_CONFIG_HOME` / `$HOME`. From a standardization point of view, this would also be a good place for any configuration, including environment variable files. Maybe the breaking v5 release would be a good opportunity to change the (default) storage location for `.lego` and the env files and to remove the use of the `PWD` by default everywhere?

Sorry that I'm jumping around here a bit :/ I hope that wasn't too confusing...

ldez:
> maybe a better default would be to read the env files from within the storage directory by default? At least then, everything would be stored in one place, 

This is exactly what I want to avoid: `.lego` directory is like a "DB", the users should not edit its content because lego relies on it to work.
This directory must be free of users data: only "programmatic files" are expected here.

>  to update the default storage directory to something in XDG_CONFIG_HOME / $HOME

I am reluctant to do so for several reasons:
1. This will increase the complexity of the migration for me and for user.
2. I don't think this is a good idea to place the different keys and certificates inside the user's home.
3. The location of this directory can already be changed with `--path`.

> Generally, these .env files are not used in production however.

Are you sure? Because there are clear references to `.env.production` inside the ecosystem.

- https://github.com/bkeepers/dotenv?tab=readme-ov-file#customizing-rails
- https://bun.com/docs/runtime/environment-variables#setting-environment-variables
- https://vite.dev/guide/env-and-mode#modes

meineerde:
> `.lego` directory is like a "DB", the users should not edit its content because lego relies on it to work.

Understood.

> I am reluctant to do so for several reasons:
> 
> 1. This will increase the complexity of the migration for me and for user.

I see that issue. However, for users, it could just be a simple directory move. If no separate path was given, lego could use the db in `$XDG_CONFIG_HOME/lego` by default but still check for `./.lego` and issue a loud warning if it exists (for a few releases). If desired, users could then set `--path .lego` and continue with their existing location.

As v5 changes a lot of command invocations however (changed command names, moved flags), about all users *have* to update their invocations anyway. It would be just one more step.

> 2. I don't think this is a good idea to place the different keys and certificates inside the user's home.

Sure, it's debatable, but I believe the `$HOME` is a "less unexpected" place than the respective `$PWD` of the command invocation, given that `$HOME` or `$XDG_CONFIG_HOME` is generally stable while the `$PWD` usually changes a lot if lego is used by humans.

> 3. The location of this directory can already be changed with `--path`.

That's true. I would argue (without having any actual evidence of course) that most people set this for production setups. We are just thinking about defaults here :)

Circling back to the `.env` files here, this would be an important point though. Unless I've missed something, with the current implementation, users can not affect the location where lego reads the `.env` files from (other than changing `$PWD`).

> Are you sure? Because there are clear references to `.env.production` inside the ecosystem.

Most web application frameworks (Rails, bun, vite, ...) generally load a `.env` and `.env.RAILS_ENV` file (among others). Most frameworks seem to follow what the dotenv gem resp. Rails did first. It's often not spelled out in such detail, but for production environments of web apps, the variables set in these files are generally not sensitive but more like general configuration (e.g. feature flags, logging and monitoring URLs, ...). 

Sensitive values are generally not loaded from there but are set through other means to avoid leaking them in the app's got repo. For the dotenv gem, they have a recommendation here: https://github.com/bkeepers/dotenv?tab=readme-ov-file#should-i-commit-my-env-file

Things are heavily dependent on the actual apps though. For the (web-) apps I'm responsible for, we often store secret environment settings in a file only readable by root. When starting the app (as root), the file is read and loaded into the environment. Then, we drop our privileges to the app user (while retaining the loaded environment) and actually start the app. That way, for the app the secrets are only ever available in the environment but are never readable on the filesystem by the app user.

In addition to these secrets, we also have `.env.production` files where non-secret things are configured and which may be committed into git and are loaded by the app itself using `dotenv` or equivalent[^1].

With that being said, I've been thinking a bit more about this general feature for `.env` files with lego on the train ride home:

The general filename of the loaded `.env` file may be confusing in the context of lego as its CLI is not called from the directory of a larger project (such as the root directory of a web application) but as a standalone command, often from the `$PATH`. As such, there is no "natural" context directory for the location of these environment files, apart from the `$PWD` which I'm arguing against.

If lego is used in the context of such a web app, the actual `.env` filename (with a possible suffix) may even clash with the environment files used for the web app. I believe this to be another reason not to load them from `$PWD` but either from a central location such as `$XDG_CONFIG_HOME/lego` / `$HOME/.config/lego`, or just from a different filename, or even from a deliberate path or filename explicitly set by the user when invoking the CLI (which would be my preferred option).

Now, thinking a bit more about how these `.env` files may be used in production given that there is a YAML config file now, I think their main advantage would be to extract secrets from the YAML config and store them very deliberately in a "more secure" location rather than a hidden file in the `$PWD`. If users extract secrets into such a file, it may be advantageous for everyone to explicitly state its location somewhere (e.g. in the YAML config) rather than to infer it to avoid "magic-at-a-distance" and "littering" with secrets.

Interestingly, users can "manually" approximate this feature (minus the separate env files per challenge) today with a simple wrapper script like this to invoke lego: [^2]

```
#!/bin/sh
. .env
exec /usr/local/bin/lego "$@"
```

However, with plain `.env` files, there isn't even _that_ much additional security to be gained at all as the files would still have to be readable by the user running lego and would thus inherit the same access level as the YAML config. Even though, if the storage location for the secrets would not matter, users could just use the YAML config file in the first place and not bother with any separate environment files at all. [^3]

I believe the main advantage of this `.env` loading feature would be to provide this additional possible security with a little less friction than the manual solution.

Sorry, this went on longer and longer... Let me close by saying that I'm incredibly excited for the changes in v5. This will be extremely awesome! THANK YOU SO MUCH FOR WORKING ON THIS PROJECT. <3

[^1]: As for Rails, they moved to a separate system of [encrypted YAML files](https://guides.rubyonrails.org/security.html#environmental-security) as a best-practice recently to allow secrets to be committed to git in an encrypted form. In production, the environment then just provides an encryption key. I'm not much of a fan though...

[^2]: Actually, I think the main reason why large web frameworks integrate reading of the various `.env` files into the framework itself is to allow multiple command related to the framework to use the same environment details so that users do not have to manually load the environment variables. Also: cargo-culting :) But this integration helps most during development. In production, this feature is much less important as there are more secure ways to set the variables there.

[^3]: That is: until there would be a way to include other (partial) config files into the YAML config as briefly discussed in #2911.

ldez:
:thinking: I think I will change a bit the behavior.

For `lego run`, I will add a flag `--env-file=.env`. The `.env` will not be loaded by default.

For `lego`, I will remove the support of `.env` file, but keep the `.env.xxx`.
Maybe I will change the pattern (ex: `.env.lego.XXX`) to be easier to gitignore those files.
I need to check how to change the directory that contains dotenv files with the configuration files.

PR #2981, opened 2026-04-15
feat: change experimental support for dotenv files

ldez:
For `lego run`, I added a flag `--env-file=.env`.
The `.env` is not loaded by default.

For `lego`, I removed the support of implicit dotenv files. The environment variables are loaded from the dotenv file defined by `envFile` in the configuration file for the DNS provider.

~~For `lego`, I removed the support of `.env` file, and changed the pattern to `.env.lego.XXX`.~~
~~I added `envFilesDirectory` inside the configuration file to change the directory that contain the dotenv files.~~

Related to #2974


ldez:
I'm not happy with `envFilesDirectory` option + dotenv patterns.

I need to think more about the topic.

ldez:
My initial goal was to have something "transparent" by using a convention.
But the convention seems more complex than an explicit path.

Also the `envFilesDirectory` option has the problem of "path relativity" I explained [here](https://github.com/go-acme/lego/pull/2974#issuecomment-4246130053).

So, in this context, I think adding an `envFile` option on the DNS challenge configuration is a better solution: no conventional pattern, and the same problem of "path relativity" than `envFilesDirectory`.

Less magic and with some know limitations, but easier to use.

PR #2984, opened 2026-04-16
feat: require EAB parameters only when needed

ldez:
Currently, lego requires the EAB parameters to create a new Client if the CA requires them, but the EAB is only required for `newAccount` requests and specifically for new account.

This means any other requests not related to `newAccount` don't require the EAB parameters.
And this also means that `newAccount` requests related to existing accounts (keyrollover, recovery) don't require the EAB parameters.

> externalAccountRequired (optional, boolean): If this field is
> present and set to "true", then the CA requires that all
> newAccount requests include an "externalAccountBinding" field
> associating the new account with an external account.
> https://www.rfc-editor.org/rfc/rfc8555.html#section-7.1.1

IMHO, the RFC is not obvious on that, this could have been expressed more clearly, but I checked with the pebble implementation and added tests.

I move the check only when `newAccount` is called to create a new account (register).

Note: this wrong check was here since the v1.0.0 (2018), it was added during the support of ACME v2 inside the PR #516.

### DIFF
--- a/cmd/cmd_accounts_register.go
+++ b/cmd/cmd_accounts_register.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log/slog"
 	"sync"
@@ -109,6 +110,10 @@ func registerAccount(ctx context.Context, cmd *cli.Command, client *lego.Client)
 	accepted := handleTOS(cmd, client)
 	if !accepted {
 		log.Fatal("You did not accept the TOS. Unable to proceed.")
+	}
+
+	if client.GetServerMetadata().ExternalAccountRequired && !cmd.IsSet(flags.FlgEAB) {
+		return nil, errors.New("server requires External Account Binding (EAB)")
 	}
 
 	if cmd.Bool(flags.FlgEAB) {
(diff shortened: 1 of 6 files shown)


ldez:
:thinking: when reading https://www.rfc-editor.org/rfc/rfc8555.html#section-7.3.4

> If such a CA requires that newAccount requests contain an
> "externalAccountBinding" field, then it MUST provide the value "true"
> in the "externalAccountRequired" subfield of the "meta" field in the
> directory object.  If the CA receives a newAccount request without an
> "externalAccountBinding" field, then it SHOULD reply with an error of
> type "externalAccountRequired".

We can say that the EAB check must be done by the CA server and not the ACME client, but there is a "SHOULD" and not a "MUST" :thinking:

ldez:
> SHOULD   This word, or the adjective "RECOMMENDED", mean that there
> may exist valid reasons in particular circumstances to ignore a
> particular item, but the full implications must be understood and
> carefully weighed before choosing a different course.
> https://datatracker.ietf.org/doc/html/rfc2119

We can do it because the server could not handle it as this is a recommendation and not a requirement.

dmke (approved):

PR #3024, opened 2026-04-25
docs: add sponsors section

ldez:
Add a sponsors section to the homepage.


meineerde:
Both sponsors link to the same URL: https://www.fairssl.dk/. I'd assume that Canonical would use a different link?

ldez:
I through I was fixed that, maybe during my different tries I reverted the fix.

PR #3034, opened 2026-05-02
docs: Add Installation Instructions for Fedora

PuercoPop:
Fedora package [lego]. Adding it to the Installation page, at the end of the official packages. As the list seemed to be ordered by official/unofficial and then lexicographically so I followed that convention when adding the entry

[lego]: https://packages.fedoraproject.org/pkgs/golang-github-acme-lego/golang-github-acme-lego/

### DIFF
--- a/docs/content/installation/_index.md
+++ b/docs/content/installation/_index.md
@@ -30,6 +30,12 @@ docker run goacme/lego -h
   yay -S lego-bin
   ```
 
+- [Fedora](https://packages.fedoraproject.org/pkgs/golang-github-acme-lego/golang-github-acme-lego/) (official):
+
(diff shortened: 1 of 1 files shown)


ldez:
We flag packages as official when we maintain them.
We don't maintain the Fedora package.

ldez:
Also this package is using a very old version of lego:
[v4.23.1](https://github.com/go-acme/lego/releases/tag/v4.23.1) (2025-04-16).

I think we cannot evaluate this package as maintained.

ldez:
I can't add this package to the documentation: as it is unmaintained it contains CVEs, so we cannot recommend it.

Thank you anyway.

PR #3100, opened 2026-05-17
-

### LINKED ISSUE #842: Randomize challenge selection OR retry failed challenge with another one
First, thanks for the fantastic, consistent work on this library. You have no idea how much I appreciate it. :pray:

I've been getting some reports lately of certificates not renewing, and it turns out it's because the TLS-ALPN challenge always fails when Cloudflare (or some other CDN) terminates TLS. Of course this is to be expected.

However, in our case, the TLS-ALPN and HTTP challenges are both enabled, yet only one gets used, because the selection of a challenge is deterministic and always returns TLS-ALPN first: https://github.com/go-acme/lego/blame/55572c26060b91518381fe99910dfacf46035544/challenge/resolver/solver_manager.go#L64

It used to be that renewals would eventually succeed because challenges were randomly ordered, so sometimes it would try the HTTP challenge, which often succeeds in the case of being fronted by a CDN.

So my feature request is that this behavior of random selection be restored, OR lego should retry a failed challenge with a different available challenge right away.

As it currently is, having both the TLS-ALPN and HTTP challenges enabled results in the HTTP challenge never being used, thus resulting in less reliability when there are network issues or -- heaven forbid -- a challenge type has to be disabled for security reasons, or the CA is doing maintenance, etc.

I would really love for both enabled challenges to be able to be used. It will greatly enhance the reliability of this package.

What do you think?

Abzaek:
## Summary

When multiple challenge types are enabled (e.g., HTTP-01 and TLS-ALPN-01), the challenge solver was deterministically picking only the first matching solver. If that solver failed (e.g., TLS-ALPN-01 fails behind a CDN that terminates TLS), no other challenge type was ever tried, causing certificate renewal failures to be permanent until the first-picked solver eventually succeeds by chance.

## Changes

- **New method `chooseSolvers()`**: Returns ALL matching solvers for an authorization instead of just the first one
- **Fallback logic in `Prober.Solve()`**: When the primary solver fails, subsequent matching solvers are tried automatically
- **Backward compatibility**: `chooseSolver()` is preserved and delegates to `chooseSolvers()`
- **Proper cleanup**: All solvers are properly cleaned up via deferred cleanup handlers

## Example scenario

Before this fix: User configures both HTTP-01 and TLS-ALPN-01 providers. The challenge list is sorted deterministically (tls-alpn-01 > http-01), so TLS-ALPN-01 is always picked. If the domain is behind Cloudflare (which terminates TLS), TLS-ALPN-01 always fails and HTTP-01 is never tried.

After this fix: When TLS-ALPN-01 fails, HTTP-01 is automatically attempted as a fallback, giving the certificate renewal a much higher chance of success.

Fixes #842

## Testing
- All existing tests pass
- Verified with `go test ./...` across the entire project

### DIFF
--- a/challenge/resolver/prober.go
+++ b/challenge/resolver/prober.go
@@ -3,6 +3,7 @@ package resolver
 import (
 	"context"
 	"errors"
+	"fmt"
 	"log/slog"
 	"time"
 
@@ -34,8 +35,8 @@ type sequential interface {
 
 // an authz with the solver we have chosen and the index of the challenge associated with it.
 type selectedAuthSolver struct {
-	authz  acme.Authorization
-	solver solver
+	authz   acme.Authorization
+	solvers []solver
 }
 
 type Prober struct {
@@ -50,6 +51,9 @@ func NewProber(solverManager *SolverManager) *Prober {
 
 // Solve Looks through the challenge combinations to find a solvable match.
 // Then solves the challenges in series and returns.
+// If the first solver fails, subsequent matching solvers are tried as fallback.
+// For example, when both HTTP-01 and TLS-ALPN-01 are enabled and TLS-ALPN-01 fails
+// due to CDN TLS termination, HTTP-01 is automatically tried next.
 func (p *Prober) Solve(ctx context.Context, authorizations []acme.Authorization) error {
 	failures := errutils.NewDomainsError("resolver")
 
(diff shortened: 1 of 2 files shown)


ldez:
https://github.com/go-acme/lego/blob/f9f9645cf7be7d399c025ec596484263eb9f963a/.github/PULL_REQUEST_TEMPLATE.md?plain=1#L5-L7

Issue #3140, opened 2026-05-28
namecheap: custom source IP


pinghe:
### Welcome

- [x] Yes, I'm using a binary release or the library within the [two latest releases](https://github.com/go-acme/lego/releases).
- [x] Yes, I've searched for similar issues on GitHub and didn't find any.

### Detailed Description

Namecheap requires an IP whitelist. When running in a NAT environment using DNS-01, it can automatically retrieve the external IP address from https://ifconfig.co/ip and supports dual-stack.

### Effective version of lego

v5.1.0

### How do you use lego?

Binary

### In which context are you using lego?

Personal use

### 🌱 Supporting the Maintainer

- [x] Yes, I find this project useful, and it has saved me time in my workflow/business.
- [x] Yes, I have [starred](https://github.com/go-acme/lego) this repository to help others find it.
- [ ] Yes, I have [sponsored](https://donate.ldez.dev) this project to keep the project stable and maintained.

### Confirmation

- [x] Yes, I've included all the information above (version, usage, etc.).

ldez:
The IP is already obtain dynamically.

ldez:
After reading again your PR, I think I understand why you are trying to ask.

Your PR is not valid, I will work on the topic.

ldez:
| Name     | Type   | MaxLength | Required? | Description                                                                                   |
|----------|--------|-----------|-----------|-----------------------------------------------------------------------------------------------|
| ClientIp | String | 15        | Yes       | An IP address of the server from which our system receives API calls (only IPv4 can be used). |

https://www.namecheap.com/support/api/global-parameters/


The documentation is clear "only IPv4 can be used".

And the client IP is already obtained dynamically.

PR #3141, opened 2026-05-28
feat(namecheap): add IP source logging, network stack awareness, and custom source IP

pinghe:
#3140

The namecheap DNS provider's IP auto-detection had two issues:
1. GetClientIP used dual-stack HTTP dialing, ignoring --ipv4only/--ipv6only flags
2. No way to determine from logs whether ClientIP came from env or auto-detection

Changes:

- Add NAMECHEAP_SOURCEIP env var to explicitly set the source IP for the Namecheap API
- Restrict GetClientIP HTTP dialer to IPv4 or IPv6 based on --ipv4only/--ipv6only
- Sync network stack CLI flags to env vars (syncNetworkStackEnv) before provider initialization, so providers can respect the network stack setting
- Change IP detection endpoint to https://ifconfig.co/ip
- Add detailed logging: IP source (env var vs auto-detect), network stack used, and detection result
- Add tests for env var priority and config passthrough

### DIFF
--- a/cmd/internal/root/process_challenges.go
+++ b/cmd/internal/root/process_challenges.go
@@ -6,6 +6,8 @@ import (
 	"strings"
 	"time"
 
+	"os"
+
 	"github.com/go-acme/lego/v5/challenge"
 	"github.com/go-acme/lego/v5/challenge/dns01"
 	"github.com/go-acme/lego/v5/challenge/dnspersist01"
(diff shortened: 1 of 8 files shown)


ldez:
https://github.com/go-acme/lego/blob/0ec9bf07300034abeaf4a7466bc72119159820d4/.github/PULL_REQUEST_TEMPLATE.md?plain=1#L3-L7

PR #3162, opened 2026-06-04
chore: update to a maintained yaml library

mikelolasagasti:
`gopkg.in/yaml` is not maintained anymore and easiest alternative is `go.yaml.in/yaml/`.

Also migrated from `v2` to `v3` for designate.


ldez:
I didn't do it because the lib is also used by other dependencies.
I don't want to have multiple versions of the lib.

Also:

https://github.com/go-acme/lego/blob/d0f86af280d5879c4888fa6d7ad26b6982e83be0/.github/PULL_REQUEST_TEMPLATE.md?plain=1#L3-L7

PR #3215, opened 2026-07-28
fix infomaniak numeric DNS record IDs

paddy73-ch:
## Summary

Fix Infomaniak DNS record creation when its API returns the new record ID as a JSON number.

## Root cause

`POST /1/domain/{id}/dns/record` returns a successful response such as `{"result":"success","data":36352129}`. The provider decoded `data` as a string, causing JSON unmarshalling to fail after the TXT record had already been created.

## Change

Decode the record ID as `uint64` and convert it to a string for the existing cleanup URL path. Update the provider test to exercise the numeric API response.

## Validation

`go test ./providers/dns/infomaniak/...`

### DIFF
--- a/providers/dns/infomaniak/internal/client.go
+++ b/providers/dns/infomaniak/internal/client.go
@@ -49,14 +49,17 @@ func (c *Client) CreateDNSRecord(ctx context.Context, domain *DNSDomain, record 
 		return "", fmt.Errorf("failed to create request: %w", err)
 	}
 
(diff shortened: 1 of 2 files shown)


ldez:
https://github.com/go-acme/lego/blob/a0e15c2daf87956943f176fd54cc6f291f221e2d/.github/PULL_REQUEST_TEMPLATE.md?plain=1#L3-L7

